PR #3, opened 2017-06-08
add breaking test, fix edge case for replacing browser "main" field

mvayngrib:
**Summary**

fixes edge case for replacing browser "main" field, when "main" field value is a file name rather than a file path

**Test plan**

added breaking test in 1st commit, fix in 2nd commit

### DIFF
--- a/packages/metro-bundler/src/node-haste/Package.js
+++ b/packages/metro-bundler/src/node-haste/Package.js
@@ -47,12 +47,25 @@ class Package {
     let main = json.main || 'index';
 
     if (replacements && typeof replacements === 'object') {
-      main =
-        replacements[main] ||
-        replacements[main + '.js'] ||
-        replacements[main + '.json'] ||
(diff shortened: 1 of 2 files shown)


jeanlauliac:
That went under the radar, let me look at this tomorrow.

mvayngrib:
@jeanlauliac great, thanks

jeanlauliac:
Hey @mvayngrib, I'm good with this since it adds a new test and doesn't change the existing behavior for the rest; landing now. However I'm not too familiar with how these fields are supposed to work. Do you have a link to some kind of specification about the `browser` field? If so, we could add a link to the spec in an inline comment. If there's none it sounds like we shall write one and add that to the repo documentation 📝 😄

mvayngrib:
@jeanlauliac this is the spec that browserify and webpack use: https://github.com/defunctzombie/package-browser-field-spec

jeanlauliac:
Thank you for sending that pull request! Lemme know if anything looks off.

PR #13, opened 2017-06-17
Add the blacklist creation function to the list of exports

ide:
We use a custom rn-cli.config.js file that specifies a custom blacklist. In addition to blacklisting some of our own directories, we still want to keep RN's default rules for what to blacklist so we also call into the packager's blacklist.

Test Plan: Added a small unit test


davidaurelio:
sure, why not

PR #14, opened 2017-06-19
Add font file extensions (otf and ttf) to default supported extensions

ide:
This makes it a little more convenient to dynamically load font files. We currently do with help from the "assetExts" config flag and find that we specify it a lot. Would you be interested in taking this PR?

Test Plan: Have tested that with custom native code, we can load otf and ttf files.

### DIFF
--- a/packages/metro-bundler/src/defaults.js
+++ b/packages/metro-bundler/src/defaults.js
@@ -15,6 +15,7 @@ exports.assetExts = [
   'm4v', 'mov', 'mp4', 'mpeg', 'mpg', 'webm', // Video formats
   'aac', 'aiff', 'caf', 'm4a', 'mp3', 'wav', // Audio formats
   'html', 'pdf', // Document formats
+  'otf', 'ttf', // Font formats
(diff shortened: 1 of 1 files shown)


jeanlauliac:
As for me I'd like to avoid adding more types to this array because there are many more types of assets out there and we cannot draw the line. Rather than an ever growing list of type, I'd either:

# Consider any file that is not a source file type to be an asset by default (no more extensions to specify manually). But, that could cause a breaking change, and conflicts, and then we'd need a blacklist, etc.
# Extract lists of assets/defaults into packages, so that people can use whichever set of assets/defaults they want and reuse that across projects that share the same defaults. That's by far my preferred solution :)

Anyhow, meanwhile, I'm okay to merge that and do (2) eventually. @davidaurelio  do you have any opinion about it?

davidaurelio:
@jeanlauliac I like both ideas. For (1) we could simply use every file that is not in the configured source extensions. I’m ok with shipping this for the time being.

PR #16, opened 2017-06-26
fix support for node v4

jeanlauliac:
Node v4 support has been broken for some time, with CircleCI failing tests. That's because it needs an additional transform for the spread operator in call position. Ex `foo(...smth)` expanding to `foo.apply(null, smth)`. Also, the build script was using syntax not supported by Node v4.

This changeset fixes the build script, and adds the missing transform. This will be used both for jest test and for the produced production output. This is needed for prod output since we want it to be able to be ran on Node v4 out-of-the-box.

CircleCI will confirm these changes work properly on all supported versions of Node.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -17,6 +17,8 @@
  *  node ./scripts/build.js /user/c/metro-bundler/packages/metro-abc/src/abc.js
  */
 
+'use strict';
+
 const fs = require('fs');
 const path = require('path');
 const glob = require('glob');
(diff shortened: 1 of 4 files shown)


scripts/build.js
```diff
@@ -17,6 +17,8 @@
  *  node ./scripts/build.js /user/c/metro-bundler/packages/metro-abc/src/abc.js
  */
 
+'use strict';
```
jeanlauliac: Needed to use `let` and `const` on Node v4.

davidaurelio:
lgtm, thank you for the fix!

PR #21, opened 2017-07-06
metro-bundler: tweak babel-register callsite

jeanlauliac:
This fixes https://github.com/facebook/react-native/issues/14530 on my local repro. The reason the original problem appears is because when requiring the preset itself, what I think is a bug in babel-register kicks in and it starts transforming the presets themselves. That fails because these preset don't actually have the correct transforms plugins installed/specified in their `package.json` (they do it at prepublish time).


davidaurelio:
Thank you for the fix. I agree that it looks like a problem with babel-register itself

PR #22, opened 2017-07-07
[babel] Add missing babel-plugin dependencies

ide:
`babelRegisterOnly.js` requires several Babel plugins that are listed as devDependencies in this repo's root package.json but aren't listed as dependencies in the metro-bundler package's dependencies. If you install metro-bundler in another project and require `babelRegisterOnly`, the Babel plugins are missing at runtime. This adds those missing Babel plugins to package.json.

They are pulled straight from the root package.json (same semver strings) and the list matches those listed in https://github.com/facebook/metro-bundler/blob/master/packages/metro-bundler/src/babelRegisterOnly.js#L23


cpojer:
I think @jeanlauliac and I discussed that we don't need this file any more and that we can move it to RN now.

ide:
Oh okay - can you guys coordinate that from your side since the change spans two repos and it's easier if you can do it from the monorepo side?

PR #31, opened 2017-07-21
Add an option to enable/disable .babelrc lookup

ide:
By default, when Babel transforms a file, it looks for .babelrc in the file's directory and all parent directories until it finds a .babelrc file. This means that when Metro tries to transform `<PROJECT_ROOT>/node_modules/dep/mod.js`, it searches for `<PROJECT_ROOT>/node_modules/dep/.babelrc`. If that .babelrc actually exists, Babel will try to apply it.

In practice, this often causes problems because packages that include .babelrc often do so unintentionally -- they don't intend for the package consumer to look at that .babelrc file. One thing we've done a lot is `rm` all .babelrc files under node_modules -- this commit effectively achieves the same in a less destructive way by telling Babel not to look up .babelrc files. (To clarify, Metro will still apply the .babelrc file in the project root.)

Since the current behavior is to look at .babelrc files, this commit keeps that behavior for now. We'll consider overriding the default behavior (that is, making Babel not lookup .babelrc files) in the default configuration of Expo/RN projects. If this goes well and we empirically find that people are having a better time, we may want to consider flipping this option's default in Metro, so that Metro tells Babel not to look up .babelrc files by default.

Test Plan: Ran `yarn test`.

Made the corresponding changes to react-native to add this option to rn-cli.config.js. Then made a dummy .babelrc file in react-native's node_modules, where the bundler would eventually look:

```
$ cat node_modules/react/.babelrc
{
  "plugins": ["dummy"]
}
```

Loaded the RNTester project with `enableBabelRCLookup` set to true and got a redbox saying `Unknown plugin "dummy"`.

screenshot 2017-07-21 02 04 28

Changed the config on the RN side to set `enableBabelRCLookup` to false, cleared the packager cache, and reloaded the bundle successfully.

### DIFF
--- a/packages/metro-bundler/src/Bundler/__tests__/Bundler-test.js
+++ b/packages/metro-bundler/src/Bundler/__tests__/Bundler-test.js
@@ -41,6 +41,7 @@ var commonOptions = {
   allowBundleUpdates: false,
   assetExts: defaults.assetExts,
   cacheVersion: 'smth',
+  enableBabelRCLookup: true,
   extraNodeModules: {},
   platforms: defaults.platforms,
   resetCache: false,
@@ -176,6 +177,7 @@ describe('Bundler', function() {
             minify: false,
             platform: undefined,
             transform: {
+              enableBabelRCLookup: true,
               dev: true,
               generateSourceMaps: false,
               hot: false,
--- a/packages/metro-bundler/src/Bundler/index.js
+++ b/packages/metro-bundler/src/Bundler/index.js
@@ -124,6 +124,7 @@ type Options = {|
   +assetServer: AssetServer,
   +blacklistRE?: RegExp,
   +cacheVersion: string,
+  +enableBabelRCLookup: boolean,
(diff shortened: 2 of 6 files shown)


jeanlauliac:
Looks good, thanks

hramos:
@jeanlauliac let me know if https://github.com/facebook/react-native/pull/15136 should be imported once this PR lands.

jeanlauliac:
We can import the other one but we'll have to publish a new major version of Metro bundler anyway

ide:
I believe there aren't any breaking changes in this commit (keeps existing behavior, adds new option) so publishing Metro with this commit as 0.10.1 should be fine. RN already requires `"metro-bundler": "^0.10.0"` so it would pick up 0.10.1 automatically (though probably more correct to bump the version to 0.10.1).

jeanlauliac:
Ah yes sorry! We don't need to upgrade, just to publish a minor ^_^

PR #39, opened 2017-08-08
Sometimes you should remove the haste map in the temp. area in order …

silyevsk:
…to fix the issue

**Summary**
Sometimes the packager exits with an error with 3 suggestions on what can be done. Sometimes however, the suggested methods are not sufficient, and the haste map in the temp area should be removed to fix the issue. This pull request adds this to the error message

**Test plan**

I'm not sure how to reproduce the error (it was fixed after I removed the files), but if there's any test that checks the error message then it should be updated.


cpojer:
@jeanlauliac mind taking this over?

cpojer:
It seems like we never got around to merging this, and now it is really outdated so I'm gonna close it.

Issue #35, opened 2017-07-31
Option for stable module Ids?


lelandrichardson:
**Do you want to request a *feature* or report a *bug*?**
Feature

i’m working on OTA updates for RN at airbnb, and one of the things that i’m looking to do is to ensure that we can push just the code that’s changed and not the whole bundle to the clients. since the RN packager uses unstable module ids after bundling code (ie, `require(123)` might need to turn into `require(124)` after a relatively trivial change in module loading order).

I was thinking about submitting a PR to metro-bundler that would add in an option to use "stable ids" for the require algorithm. This would probably end up using the `localPath` of the module, as opposed to the auto-incrementing module id.

I'm still considering strategies, but I'm thinking that in the situation where the user is using `unbundle`, a map of `localPath => moduleId` could be added to a global variable or something, which the `require.js` polyfill could utilize.

If I made such a PR, would FB be willing to accept it? It would be done in such a way that it would be completely opt-in.

cc @jeanlauliac @cpojer

mc-zone:
Also need this feature. There is also an issue #6 I opened earlier to discuss about it.

rafeca:
Hey @lelandrichardson! this was added in f347e4ff470ca858cdb8af5a80d409cc5061e9ee and is available on 0.24.1

lelandrichardson:
@rafeca awesome, thanks!

Issue #6, opened 2017-06-09
[Discuss]Problems about unstable numeric module ID.


mc-zone:
**Do you want to request a *feature* or report a *bug*?**
Feature, or discuss first.

**What is the current behavior?**
### 1. Unfriendly to diff.
Supposing there is an entry which some deps like following:
```
import a from "./lib/a";
import b from "./lib/b";
import c from "./lib/c";
import d from "./lib/d";
```
Bundle it first. Then remove one of deps, for example, `./lib/a`:

```diff
-import a from "./lib/a";
import b from "./lib/b";
import c from "./lib/c";
import d from "./lib/d";
```

Bundle it again. **Check bundle's diff**:

![image](https://user-images.githubusercontent.com/4403937/26960562-dcf75432-4d0a-11e7-88fd-dc7c32cd8b35.png)

As you see, all define/require code (module ID,  exactly) about deps after `./lib/a` were changed. 

This may give rise to lots of unecessary mutated in some projects's bundle which has huge dependency graph. It's not friendly to bundle diff/update. (Some Apps and 3rd. tools will update bundle by diff algorithm to reduce the consumption from un-splitting bundle size.)

### 2. Can' t run muti-bundle(RN App) at one context.
In our team, there are many native Apps which have been integrated React Native paritially, and each native App may include mutiple RN Apps (with mutiple entries). 

We are finding a way that can run mutiple RN Apps in a common JS context so that we don't need to recreate bridge/context, run bundle and define method/modules over and over, even can share data and modules with each other.

Think about the style of Node.js server application, we don't need to redefine modules like react, react-dom, react-router blah-blah for each route when it called by requests.

![image](https://user-images.githubusercontent.com/4403937/26965347-e90249e0-4d27-11e7-92d0-c6d031e48bd0.png)

But at least now, It can't be implemented because of numeric module ID. Bundles will override other's modules which always start with 0, 1, 2 but not an unique ID.

### 3. Unfriendly to bundle-splitting (module reference).
We are also trying to split bundle for decrease update download size. You can see an earlier incomplete PR facebook/react-native#10804 I've sent.

We splited most of library modules, like react, react-native and our common utils into a base bundle, only pack business code into business bundle, then maintaining a `manifest.json` file to handle module call from business bundle to base bundle (Like webpack's DLL way). 

![image](https://user-images.githubusercontent.com/4403937/26966101-4fcad1b2-4d2b-11e7-8be4-767ed6c45cb3.png)

[...]

PR #44, opened 2017-08-21
Add stable id option to bundler

lelandrichardson:
**Summary**

This PR adds an option to the bundler for a `makeStableId` function of type `(m: Module, id: number) => string | number`. The purpose of this option is to allow people to specify module ids that are stable and are based on the identity of the module, rather than it's index in the topological sorting of the dependency graph (which is how it works now).

The numeric module ids are important and necessary for the RAMBundle format, however one can create a mapping from `stable id => module id` that can be used in tandem with this and the RAMBundle format.

At Airbnb we need this in order to provide diffs from one build to another. Without a stable id, a single line change can result in every file in the code base requiring a diff (ie, inserting a module will create a +1 offset on all of the modules.

The `makeStableId` function could be provided in the `rn-cli.config.js`. The default will simply be: `(m, id) => id` which will preserve the current functionality and make this a non-breaking and backwards compatible change.

This change has been requested and discussed in the following issues: #6 and #35 

I urge the metro bundler team to consider such a change. Let me know if there is anything I can do or need to prove in order to make this accepted.

**Test plan**

I've been using this code locally and it works. I'll make sure that tests and stuff pass but if there's anything you want me to show or verify, let me know.

### DIFF
--- a/packages/metro-bundler/src/Bundler/__tests__/Bundler-test.js
+++ b/packages/metro-bundler/src/Bundler/__tests__/Bundler-test.js
@@ -50,6 +50,7 @@ var commonOptions = {
   sourceExts: defaults.sourceExts,
   transformModulePath: '/path/to/transformer.js',
   watch: false,
+  makeStableId: (m, id) => id,
 };
 
 describe('Bundler', function() {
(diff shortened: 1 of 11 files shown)


jeanlauliac:
Hello @lelandrichardson, I'd like to figure out a way to adapt this to the current version of Metro (I know it's been a long time, and I'm now trying to catch up). In particular I'd like to measure the perf impact that may have on RAM bundles. Do you confirm you're using RAM bundles in production, and so that you need support of stable IDs for RAM bundles?

AdamLiechty:
Would https://github.com/facebook/metro/pull/100 be a suitable solution for you @lelandrichardson ?

rafeca:
Closing this PR, since we already merged https://github.com/facebook/metro/pull/100 (and is available on v0.24.1)

PR #45, opened 2017-08-23
Fixes Platform.OS and process.env.NODE_ENV inlining resulting in an invalid AST

jevakallio:
Resolves https://github.com/facebook/metro-bundler/issues/27

**Summary**

There is an edge case where tools like react-primitives need to assign to `Platform.OS`,
but the inline transformation results in an invalid AST.

When Platform.OS is unconditionally inlined, the following scenario: 
````
Platform.OS = 'ios';
````
Is transformed to: 
```
"ios"="ios"
````

And results in error `Property left of AssignmentExpression expected node to be of a type ["LVal"] but instead got "StringLiteral"**`.

See issue in react-primitives: https://github.com/lelandrichardson/react-primitives/issues/79

This patch checks whether the current node is on the left hand side of an AssignmentExpression
and skips the inlining when this is the case.

It also does the same for `process.env.NODE_ENV`, which suffers from the same problem. See related closed issue https://github.com/facebook/react-native/issues/7607#issuecomment-221425153 which was resolved by using bracket notation `process.env["NODE_ENV"]` instead.

**Test plan**

Unit tests included.

**Notes**

This is my first contribution to Metro, and haven't worked with Babel AST a lot, so constructive feedback on the implementation is welcome.

### DIFF
--- a/packages/metro-bundler/src/JSTransformer/worker/__tests__/inline-test.js
+++ b/packages/metro-bundler/src/JSTransformer/worker/__tests__/inline-test.js
@@ -294,6 +294,90 @@ describe('inline constants', () => {
     );
   });
 
+  it(`doesn't replace Platform.OS in the code if Platform is the left hand side of an assignment expression`, () => {
+    const code = `function a() {
+      Platform.OS = "test"
+    }`;
+    const {ast} = inline('arbitrary.js', {code}, {platform: 'ios'});
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jevakallio** [comment]
The AppVeyor build has failed for unrelated reasons - looks like the tests related to this change ran successfully:
```
PASS  packages\metro-bundler\src\JSTransformer\worker\__tests__\inline-test.js
```

PR #49, opened 2017-09-06
upgrade Flow to v0.54.0

jeanlauliac:
Recent upgrade of Flow on the internal monorepo introduced changes in JS that make it fail in OSS. Upgrade Flow to fix (we'll need to upgrade in react-native as well, most likely).

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "eslint-plugin-flowtype": "^2.28.2",
     "eslint-plugin-prettier": "2.1.1",
     "eslint-plugin-react": "^7.2.1",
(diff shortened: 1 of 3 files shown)


hramos:
The internal diff was abandoned. Are you still looking into this? React Native's OSS tests are also broken, due to its dependency on metro-bundler.

jeanlauliac:
Yes, there is another changeset for it that will land today.

kassens:
Also bump the version in the `.flowconfig` once #53 lands.

cpojer:
Done in https://phabricator.intern.facebook.com/D6361276 which will land soon.

PR #53, opened 2017-09-14
Add version to .flowconfig

kassens:
**Summary**

Without this editors or a globally installed flow would use the latest version.
This ensures that flow uses the right version when checking the code.

**Test plan**
```
$ flow
```

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -13,3 +13,6 @@ suppress_comment=\\(.\\|\n\\)*\\$FlowExpectedError
 
 unsafe.enable_getters_and_setters=true
 munge_underscores=true
+
+[version]
+^0.48.0


cpojer:
Thanks, please feel free to land this, but internally we use the latest version of flow for this repo, so I'm not sure if this is the right fix. What problems are exposed with Flow on the open source repo that we cannot fix easily?

kassens:
Nuclide just complained and I thought development would happen on GH.

Seems like a good idea to pin the version to what is used in the package.json and therefor running on CI?

cpojer:
Nope, development is fb first and this repo is somewhat of a mess unfortunately. @byk is planning to fix it soon.

kassens:
Upgrade to flow 54 seems to be a bit involved and @jeanlauliac is trying that in #49.

cpojer:
Let's wait for that other diff then. You mind making your change agains the internal version? :)

kassens:
Sure

PR #63, opened 2017-09-21
Chore: update yarn.lock for Yarn 1.0

BYK:
**Summary**

Cloned the repo, run `yarn install` and it updated the `yarn.lock` file. Assuming this is due to Yarn 1.0 upgrade, done recently.

**Test plan**

Run `yarn install`, expect no change in `yarn.lock` file.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,9 +1,9 @@
 machine:
   environment:
-    YARN_VERSION: 0.24.3
+    YARN_VERSION: 1.0.2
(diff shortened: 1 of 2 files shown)


BYK:
@jeanlauliac I think they do so this patch needs to be updated :(

cpojer:
Done with https://phabricator.intern.facebook.com/D6361276 which will land soon.

PR #64, opened 2017-09-21
Chore: resolve lint errors

BYK:
**Summary**

Ran `yarn lint-fix` and got these updates. Also fixed 5 ESLint warnings.
Need these in place for builds to pass.

**Test plan**

Run `yarn lint` and expect it to pass without errors.

### DIFF
--- a/packages/metro-bundler/src/Resolver/__tests__/Resolver-test.js
+++ b/packages/metro-bundler/src/Resolver/__tests__/Resolver-test.js
@@ -16,7 +16,6 @@ jest.useRealTimers();
 
 jest.mock('path');
 
-const {join: pathJoin} = require.requireActual('path');
 const DependencyGraph = jest.fn();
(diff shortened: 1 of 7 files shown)


packages/metro-bundler/src/Resolver/__tests__/Resolver-test.js
```diff
@@ -219,14 +218,14 @@ describe('Resolver', function() {
       const module = createModule('test module', ['x', 'y']);
       const resolutionResponse = new ResolutionResponseMock({
         dependencies: [module],
-        mainModuleId: 'test module',
+        mainModuleId: 'test module'
```
jeanlauliac: In the internal lint/prettier config, we force adding the commas, so it wouldn't be compatible. We should change prettier configuration instead, it seems.

cpojer:
Superseded by https://phabricator.intern.facebook.com/D6362298 which will land shortly.

PR #67, opened 2017-10-05
Exempt Resolver/polyfills/require.js from extract-dependencies

### LINKED ISSUE #65: Getting "require() must have a single string literal argument" when bundling moment.js

Bug

When attempting to package a project which makes use of the https://github.com/moment/moment.js library, the packager crashes with the error `require() must have a single string literal argument`

The library line in question is: https://github.com/moment/moment/blob/develop/moment.js#L1830 which appears to be perfectly javascript code.

This behavior does not seem to happen when I build the bundle with RN 0.48.4, however that particular build has issues when running on an phone running Oreo.

I can work around this issue by disabling the error in question by adding a `return` statement before this line: https://github.com/facebook/metro-bundler/blob/master/packages/metro-bundler/src/JSTransformer/worker/extract-dependencies.js#L39

I am working on a repo to reproduce this behavior,  however my time is limited due to pressing issues that need to be addressed within the next week so I have not been able to get a proper minimalist reproduction of this behavior.

I will update this issue with a link once I can get a good example.

There should not be any errors when packaging the JS bundle with the above library

**Please provide your exact metro-bundler configuration and mention your metro-bundler, node, yarn/npm version and operating system.**

RN: 0.49.0-rc.5
metro-bundler: 0.13.0
OS: Arch Linux
Mobile OS: Android 8.0.0

jrylan:
**Summary**

`metro-bundler/src/JSTransformer/worker/extract-dependencies.js` parses every file to gather the usage of `require(...)` and extract the module name for building the dependency array.

The problem is that `exract-depdendencies.js` additionally analyzes `metro-bundler/src/Resolver/polyfills/require.js` which includes `require(moduleId)`. So it's failing every single time because `moduleId` is an `Identifier` (variable) instead of a `StringLiteral` as required in `extract-dependencies.js`.

Fixes #65

**Test plan**

Generate a successful build as normal.

### DIFF
--- a/packages/metro-bundler/src/JSTransformer/worker/index.js
+++ b/packages/metro-bundler/src/JSTransformer/worker/index.js
@@ -18,6 +18,7 @@ const extractDependencies = require('./extract-dependencies');
 const inline = require('./inline');
 const invariant = require('fbjs/lib/invariant');
 const minify = require('./minify');
+const path = require('path')
 
 import type {LogEntry} from '../../Logger/Types';
 import type {MappingsMap} from '../../lib/SourceMap';
@@ -85,6 +86,11 @@ type TransformCode = (
   Callback<Data>,
 ) => void;
 
+const pathToRequirePolyfill = path.resolve(
+  __dirname,
+  '../../Resolver/polyfills/require.js'
+)
+
 const transformCode: TransformCode = asyncify(
   (
     transformer: Transformer<*>,
@@ -141,8 +147,8 @@ const transformCode: TransformCode = asyncify(
       code = code.replace(/^#!.*/, '');
     }
 
-    const depsResult = isJson
-      ? {dependencies: [], dependencyOffsets: []}
+    const depsResult = isJson || filename === pathToRequirePolyfill
+      ? { dependencies: [], dependencyOffsets: [] }
       : extractDependencies(code);
 
     const timeDelta = process.hrtime(


jrylan:
The CI check is failing on code unrelated to my PR.

cpojer:
This is weird, and hasn't been a problem for us. I would prefer not to hardcode the specific module path here. cc @davidaurelio what do you think?

jrylan:
@cpojer I don't know if perhaps I'm hitting the issue because I use a custom rn-cli.config.js just to use a custom transform for TypeScript, but I hit the issue literally every single time. My build gets to 99%+ and then I get the red screen with "require() must have a single string literal argument". I edited `extract-dependencies.js` to inspect the value of `nodeArgs` in the `pushDependency` function and it was always an `Identifier` (instead of `StringLiteral` with the name `moduleId`.

So then I hoped over `src/JSTransformer/worker/index.js` and wrapped a try/catch around the use of `extractDependencies` to catch and log the filename of the problematic file... and it prints out `src/Resolver/polyfills/require.js` every time.

To work around the issue, I implemented a node module resolver hook to intercept the import of `src/JSTransformer/worker/index.js` and replace it with the code as seen here in my PR. Works perfectly fine and exactly as expected.

cpojer:
How about instead ignoring that file from having deps extracted? It's almost certainly the config you have, and I apologize that the current config system allows you to shoot yourself in the foot.

jrylan:
@cpojer you mean add it to the blacklistRE in rn-cli.config.js? Trying it now.

jrylan:
@cpojer if I add the following in my rn-cli.config.js I still get `bundling failed: Error: require() must have a single string literal argument`.

```js
  getBlacklistRE(platform) {
    return blacklist([
      /.*\/polyfills\/require\.js.*/
    ])
  },
```

Is this what you meant for me to try or is there something else?

cpojer:
Is that on the same file? That would be weird.

jrylan:
@cpojer yup, same file as before -- always erroring on the require polyfill. Here's the contents of my rn-cli.config.js:

```js
const blacklist= require('metro-bundler/src/blacklist')

export default {
  getBlacklistRE(platform) {
    return blacklist([
      /.*\/polyfills\/require\.js.*/
    ])
  },

  getTransformModulePath() {
    return require.resolve('./transformer.js')
  }
}
```

The transformer is very simple, just a basic TypeScript transform.

cpojer:
Alright, this has devolved into a question thread rather than a PR. I don't think there is anything wrong with Metro, and I'd like to keep this code outside of the extraction code. For TypeScript compilation with RN, you could reach out to @orta. Thanks for understanding.

jrylan:
@cpojer This isn't a question thread. I'm providing you with the details of the scope of my setup so that there is less need for back-and-forth questions to resolve the problem.

The issue is completely unrelated to TypeScript, none of the problem is happening in the transform at all. Somehow even with adding polyfills/require.js (part of the Metro source code) to the blacklist, Metro still tries to extract dependencies from a file where it shouldn't be... A file completely outside of my code and never referenced in my code.

I am having no problem at all with TypeScript and in fact the previous version of Metro and React Native have worked perfectly fine in my setup for a long, long time. This is a regression issue in the latest version of Metro.

I understand that you're swamped and don't have much time to invest in issues, but I think it was premature to close this.

cpojer:
Mind looking at the default config to figure out what's different with your config?

jrylan:
@cpojer - my sincere apologies, I was wrong and it is related to the custom transform in my rn-cli.config.js. Others may run into this problem, so if you think it's worthy of me documenting and updating docs, please let me know....

Basically, in my custom transform, I have a custom method that checks the file path to see if my transformer should actually do anything or just return the source code that was provided to it unchanged. Essentially, if a filename contains `node_modules` and then isn't followed after that by something like `react|react-*|native-*` then my transformer skips it and returns 

```js
return {
  filename,
  ast: null,
  code: src, // This is the src that is passed into the custom transform defined in rn-cli.config.js
  map: null
}
```

What is going on is that in previous versions of Metro and React Native, it was fine for me to just return the source of any Metro Bundler file unchanged -- that is now no longer the case. Metro Bundler must be added to my transform whitelist in order to accommodate the following two files that are added to your app bundle:

1. metro-bundler/src/Resolver/polyfills/prelude_dev.js
2. metro-bundler/src/Resolver/polyfills/require.js

My custom transform handles TypeScript when the file extension matches .ts or .tsx, otherwise it passes it off to Babel using the `babel-preset-react-native/configs/main` or `babel-preset-react-native/configs/hmr`.... So basically, in this case my transformer was never allowing either of those Metro files to hit Babel, and it looks like they need to now.

cpojer:
Hey, thanks for following up @jrylan. It's ok to be wrong sometimes, I am wrong often too :) I appreciate that you documented this issue so thoroughly here. I'm wondering if we should just whitelist these two files and transform them with a default worker in Metro to make this easier? That way your transformer doesn't have to concern itself with Metro's own code.

PR #69, opened 2017-10-07
Prevent error on safely-handled requires using non-static arguments

richardgirges:
**Summary**
This is the workaround that @cpojer suggested in #65 and should resolve the biggest pain point that developers are experiencing in that issue: using moment.js in React Native. Hopefully this at least serves as a starting point to reaching a decent solution for libraries using non-static `require()` statements.

**Test plan**
##### 1. Two unit tests added:
- Expected error when non-static-argument-based `require()` is nested more than one level deep within a `try` statement
- Expected no error when non-static-argument-based `require()` is nested directly within a `try` statement

##### 2. Ran the following command and got a clean bundle:
```bash
ip-192-168-1-10:trueflipapp richard$ react-native bundle --entry-file index.js --platform ios --bundle-output /tmp/test.js --reset-cache
Scanning folders for symlinks in /repo/trueflip/sandbox/code/trueflipapp/node_modules (8ms)
Scanning folders for symlinks in /repo/trueflip/sandbox/code/trueflipapp/node_modules (8ms)
Loading dependency graph, done.
warning: the transform cache was reset.
bundle: start
bundle: finish
bundle: Writing bundle output to: /tmp/test.js
bundle: Done writing bundle output
```

### DIFF
--- a/packages/metro-bundler/src/JSTransformer/worker/__tests__/extract-dependencies-test.js
+++ b/packages/metro-bundler/src/JSTransformer/worker/__tests__/extract-dependencies-test.js
@@ -102,6 +102,22 @@ describe('Dependency extraction:', () => {
     );
   });
 
+  it('throws on calls to require with non-static arguments, nested deeper than one level inside of a try block', () => {
+    const code = "try { if (1) { require('foo/' + bar) } } catch (e) { }";
+
+    expect(() => extractDependencies(code)).toThrowError(
+      'require() must have a single string literal argument',
+    );
+  });
+
+  it('does not throw on calls to require with non-static arguments, nested directly inside of a try block', () => {
+    const code = "try { require('foo/' + bar) } catch (e) { }";
+
+    const {dependencies, dependencyOffsets} = extractDependencies(code);
+    expect(dependencies).toEqual([]);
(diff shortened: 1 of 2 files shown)


richardgirges:
@cpojer I tried to give issue #65 a go in this PR. Please let me know if this looks like a good starting point to getting a fix in.

cc: @TikiTDO who I know has had some pretty significant opinions on this issue. I apologize for doing the same work you were looking into - I couldn't wait much longer due to a large number of dependencies we're using that have moment as a sub-dependency.

richardgirges:
The five errors thrown in the CI were present in the master branch. Didn't want to digress from the issue at hand by fixing them.

TikiTDO:
@richardgirges In all honestly, I just want it to work. The sooner the better. If you can get this resolved yesterday my life would be much easier.

cpojer:
Oh nice, thanks for sending this PR. This is definitely the right direction. I'll talk to the Metro team next week about whether this is sufficient or whether we'll need to do other work to make the bundling work properly.

aethiss:
@cpojer we can have some news asap about that issue so ?
Tnx on advance for this job ;)

PR #70, opened 2017-10-07
Whitelist node opt --max-old-space-size for JSTransformer

zk:
Whitelist node opt `--max-old-space-size` for JSTransformer.

Bundling can choke on the default node memory settings for those of us generating large index.js files. This change allows supplied node options to make it through to the node processes spawned by the JSTransformer worker farm.


zk:
CLA signed

PR #71, opened 2017-10-09
Update README with some useful information

mjesun:
Add useful information in the README, as an initial first version. Not all of the fields are explained, and we still miss some useful bullet points (e.g. how to release), but that should be good to give some initial guidance as a standalone project.


cpojer:
Why do you send a PR? :O

PR #83, opened 2017-11-07
Fix circular dependencies support when module.exports is replaced

cgupta-logmein:
… was part of that and it was suggested to merge with this project.

Link to original request
https://github.com/facebook/react-native/pull/13980

…to their reliance on some CommonJS/node module behavior not implemented by the React Native require polyfill from the packager.

**This PR adds some of these behaviors:**

Circular dependencies when exporting by assignment to the module.exports object. Should fix #3298 for good.
Exporting by assignment to this object. Fixes #2862
The module.id represents a unique identifier that can be used in require() calls.
There were no tests for the require polyfill so I added one.

**Motivation (required)**

What existing problem does the pull request solve?
Circular dependencies when exporting by assignment to the module.exports object. Should fix #3298 for good.
Exporting by assignment to this object. Fixes #2862

**Test Plan (required)**

There is no change to UI or API. However, tests have been added for require.js polyfill.

Next Steps

Individual and company CLA have been signed already.

**Summary**

**Test plan**

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
     "timers": "fake",
     "setupFiles": [
       "<rootDir>/scripts/setupJest.js"
-    ],
-    "testEnvironment": "node"
+    ]
   }
 }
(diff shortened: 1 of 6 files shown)


cgupta-logmein:
Can we verify how big is a performance hit? We are running a very large project and it is very unconvinient to make these changes everytime.

cpojer:
Closing as we aren't going to merge this PR.

PR #88, opened 2017-11-12
Setup Continuous Delivery

Kureev:
**Summary**
As we discussed on Discord channel on Nov 6th, this PR adds appropriate hooks to CircleCI in order to deploy successful builds (on master branch) to `npm` using tag that has been assigned to the latest commit.

This is how deployment process should look like:
Once you want to deploy, push a new tag to master. That will trigger a deployment script on CI. [According to CircleCI documentation](https://circleci.com/docs/1.0/npm-continuous-deployment/), the best way is to use
```
npm version x.y.z
```
in order to update package version and create git tag.  Then, you can push your code using 
```
git push --follow-tags
```
After successful build, a new version will be released.

**Test plan**
Not sure if there is a way to properly test it before the merge. I think the only way is to merge it in, setup NPM_TOKEN env variable and try to create a new version by running build on CI. Please, let me know if you see alternative scenarios.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -12,6 +12,7 @@ dependencies:
       fi
     - nvm install 4
     - nvm install 6
+    - echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> ~/.npmrc
 
   cache_directories:
     - ~/.yarn
@@ -26,3 +27,8 @@ test:
     - yarn run test-ci
     - nvm alias default 8
     - yarn run test-ci
+deployment:
+  npm:
+    tag: /v[0-9]+(\.[0-9]+)*/
+      commands:
+        - npm publish
--- a/circle.yml
+++ b/circle.yml
@@ -10,16 +10,6 @@ aliases:
       - node_modules
     key: dependencies-{{ .Branch }}-{{ checksum "package.json" }}
 
-  - &yarn-install
-    run: |
-      sudo npm i -g yarn@^1.2.1
-      yarn --no-progress
-
-  - &yarn-install-no-sudo
-    run: |
-      npm i -g yarn@^1.2.1
-      yarn --no-progress
-
   - &deploy
     command: |
       npm run prepare-release
@@ -56,7 +46,8 @@ jobs:
     steps:
       - checkout
       - restore-cache: *restore-cache
-      - *yarn-install
+      - run: sudo npm i -g yarn@^1.2.1
+      - run: yarn --no-progress
       - save-cache: *save-cache
       - run: yarn run test-ci
 
--- a/circle.yml
+++ b/circle.yml
@@ -27,6 +27,17 @@ aliases:
 
 version: 2
 jobs:
+  test-node-9:
+    working_directory: ~/metro-bundler/packager/metro-bundler
+    docker:
+      - image: circleci/node:9
+    steps:
+      - checkout
+      - restore-cache: *restore-cache
+      - run: yarn --no-progress
+      - save-cache: *save-cache
+      - run: yarn run test-ci
+
   test-node-8:
     working_directory: ~/metro-bundler/packager/metro-bundler
     docker:
@@ -67,6 +78,7 @@ workflows:
   version: 2
   build-and-deploy:
     jobs:
+      - test-node-9
       - test-node-8
       - test-node-6
       - deploy:
--- a/circle.yml
+++ b/circle.yml
@@ -11,9 +11,11 @@ aliases:
     key: dependencies-{{ .Branch }}-{{ checksum "package.json" }}
 
   - &deploy
-    command: |
-      npm run prepare-release
-      npm deploy
+    working_directory: packages/metro-bundler
+    steps:
+      - run: npm run prepare-release
+      - run: npm publish
+      - run: npm run cleanup-release
 
 version: 2
 jobs:
--- a/circle.yml
+++ b/circle.yml
@@ -74,7 +74,5 @@ workflows:
       - test-node-6
       - deploy:
           filters:
-            branches:
-              only: master
             tags:
               only: /v[0-9]+(\.[0-9]+)*/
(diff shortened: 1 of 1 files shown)


rafeca:
That's awesome! Thanks a lot for working on this, this will allow us move the npm publish step from our internal infra to GH-first, something we've wanted to do for a long time.

Unfortunately, we can't `npm publish` directly the content of the repo, since we need to run a build step before to remove the flow annotations from the code.

The (super-hacky)™ way that I've been using lately to publish to npm from the github repo is the following:

1. Run `yarn` to install the needed dev-dependencies.
2. Run `yarn run build`, to create the publish-ready files (they're stored in `packages/metro-bundler/build`).
3. `rm -fr packages/metro-bundler/src; mv packages/metro-bundler/build packages/metro-bundler/src` (to replace the `src` content with the built code, since the `"main"` field in our package.json [points to `src/`](https://github.com/facebook/metro-bundler/blob/master/packages/metro-bundler/package.json#L5) ).
4. `cd packages/metro-bundler; npm publish` (we need to publish from that sub-folder).
5. `git checkout .; git clean -dxf` (to clean-up the repo).

As you can see, this process is terrible, so if you have any suggestion to improve it will be greatly appreciated :) In the meantime we could have an npm script that does this process which can be run using `yarn run publish` from circleCI.

Kureev:
Hi @rafeca!

Thanks for the feedback! I have a question regarding №3: any particular reason why you want to point to `src` instead of `build`? Is it done so to make development process easier or something?

In the meantime, I'll make a script that reflects your algorithm above so we can start from something :)

rafeca:
>Thanks for the feedback! I have a question regarding №3: any particular reason why you want to point to src instead of build? Is it done so to make development process easier or something?

This is because internally at FB people don't use the npm version of metro. Instead, its master version is required directly from our monorepo (with require hooks to remove flow annotations)

Kureev:
I've updated PR according to your feedback and discussion that took place in Discord.
Looking forward for your feedback! 🙌

Kureev:
Can you take another glance, @rafeca?

circle.yml
```diff
@@ -1,28 +1,80 @@
...
-    - yarn run test-ci
+  - &save-cache
+    paths:
+      - node_modules
+    key: dependencies-{{ .Branch }}-{{ checksum "package.json" }}
+
+  - &deploy
+    command: |
+      npm run prepare-release
+      npm deploy
```
rafeca: shouldn't this be `npm publish`?

circle.yml
```diff
@@ -1,28 +1,80 @@
...
+  version: 2
+  build-and-deploy:
+    jobs:
+      - test-node-9
+      - test-node-8
+      - test-node-6
+      - deploy:
+          filters:
+            branches:
+              only: master
```
rafeca: We usually do minor releases from branches, so I think we can just remove the branches filter (only filtering by the tag name is enough)

circle.yml
rafeca: Also, let's use yarn for running the `prepare-release` script and also run the `cleanup-release` script afterwards:

```sh
yarn run prepare-release
npm publish
yarn run cleanup-release
```

circle.yml
```diff
@@ -1,28 +1,80 @@
...
+
+  deploy:
+    working_directory: ~/metro-bundler/packager/metro-bundler
+    docker:
+      - image: circleci/node:8
+    steps:
+      - checkout
+      - restore-cache: *restore-cache
+      - run: |
+          cd packages/metro-bundler
```
rafeca: Instead of executing `cd` you can use `working_directory`, this way the step does not have "side effects":

```yaml
- run:
    working_directory: packages/metro-bundler
    command: yarn --no-progress
```

(you'll also need to update the deploy alias, adding the `working_directory`)

circle.yml
Kureev: > shouldn't this be npm publish?

🤦‍♂️ 

> Also, let's use yarn for running the prepare-release script and also run the cleanup-release script afterwards:

Do we really need a cleanup here?

rafeca: > Do we really need a cleanup here?

In this current scenario is not really needed, but it's still good to leave the repo cleaned up (less confusion for the future engineer that may want to add another step after deploy)

Kureev:
Should be good now 😉

rafeca:
Niceeeee! I've just imported the PR to phabricator, to land it from there

BYK:
YISSS, we got the CircleCI builds. Thanks a lot for doing this

BYK:
Also, check out our config in Yarn to see how to enable macOS builds too: https://github.com/yarnpkg/yarn/blob/master/.circleci/config.yml

circle.yml
```diff
@@ -1,28 +1,80 @@
...
+# Workflows enables us to run multiple jobs in parallel
+workflows:
+  version: 2
+  build-and-deploy:
+    jobs:
+      - test-node-9
+      - test-node-8
+      - test-node-6
+      - deploy:
+          filters:
```
BYK: You need
```
branches:
   ignore: /.*/
```

here. Otherwise, this will deploy every single commit in any branch. See https://circleci.com/docs/2.0/workflows/#git-tag-job-execution

>CircleCI treats tag and branch filters differently when deciding whether a job should run.
>
>    For a branch push unaffected by any filters, CircleCI runs the job.
>    For a tag push unaffected by any filters, CircleCI skips the job.

BYK:
Also looks like you need a rebase :)

circle.yml
Kureev: @BYK maybe, we need
```
branches:
  only: master
```
instead? I think it will be a little bit more straight forward

Kureev:
Thanks for your feedback, @BYK!

Should be adjusted now.

Kureev:
Any updates on this, @rafeca?

rafeca:
@Kureev I'm importing the PR to phabricator right now to land it :)

Next step is getting the CI tests passing, @cpojer offered to fix them 😄

Kureev:
"offered" 😆

PR #89, opened 2017-11-13
Normalize module name to have platform specific folder separators

bweggersen:
**Summary**

It is currently not possible to resolve specific module imports (such as `react-native/Libraries/Image/AssetRegistry` using forward slashes as folder separators) using a custom mapping defined in `extraNodeModules` on Windows. This PR solves this issue by normalizing module names to use platform-specific folder separators before splitting the module name using `path.sep` as separators.

**Test plan**

We use `extraNodeModules` to create a mapping between `react-native` and a forked and scoped version of react-native (i.e. `@scope/react-native`). If we import a specific file from react-native (such as [`react-native/Libraries/Image/AssetRegistry`](https://github.com/facebook/react-native/blob/master/local-cli/core/Constants.js), which gets referenced when importing image assets), `ModuleResolution.js` is not able to extract the first folder name on Windows. This first folder name is the name of the module (`react-native` in the previous example) and is used to query `extraNodeModules` for possible matches. It is not able to find this folder name because the module name is split using `path.sep` as a separator, which is `'\\'` on Windows. Most module names use forward slashes as folder separators. The solution is to normalize `toModuleName` before we split on `path.sep`.

### DIFF
--- a/packages/metro-bundler/src/node-haste/DependencyGraph/ModuleResolution.js
+++ b/packages/metro-bundler/src/node-haste/DependencyGraph/ModuleResolution.js
@@ -266,7 +266,7 @@ class ModuleResolver<TModule: Moduleish, TPackage: Packageish> {
     const extraSearchQueue = [];
     if (this._options.extraNodeModules) {
       const {extraNodeModules} = this._options;
-      const bits = toModuleName.split(path.sep);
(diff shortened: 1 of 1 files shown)


bweggersen:
Thanks @jeanlauliac! Let me know if you need anything else from regarding this issue.

PR #100, opened 2017-12-02
Expose createModuleIdFactory as bundler option

AdamLiechty:
**Summary**

`createModuleIdFactory` is already used in `metro` internally, but is currently always a fixed function.
This enables `metro.runBuild()` to be run with a custom module ID factory.

One use case: building a base bundle, on top of which other application-specific bundles could reference modules in the base.  The application-specific IDs need to not conflict with those in the base bundle, and all references to modules in the base must resolve to the correct ID in the base bundle.  A custom ID factory can make all this possible.

**Test plan**

Using `metro.runBuild(...)` with these changes, I was able to substitute in a custom ID factory

```javascript
const fs = require('fs')
const metro = require('metro')

const baseManifestFileContents = JSON.parse(fs.readFileSync('./baseManifest.json'))
// baseManifest looks like:
// { "modules": { ...
//    "react/index.js": { "id": 12 },
//    "react/cjs/react.production.min.js": { "id": 13 }, ...  } }

const opts = {
  dev: false,
  entry: 'index.js',
  out: './out.bundle',
  platform: 'ios',
  projectRoots: ['.', 'node_modules'],
  config: {
    createModuleIdFactory: createModuleIdFactory(baseManifestFileContents)
  }
}

metro.runBuild(opts)

// Creates a sample custom ID factory
function createModuleIdFactory(manifestFileContents) {
  return function createModuleIdFactory() {
    const fileToIdMap = new Map()
    let nextId = manifestFileContents ? getNextIdAfterBaseManifest(manifestFileContents) : 0

    return path => {
      const sourcePath = path
        .replace(process.cwd() + '/node_modules/', '')
        .replace(process.cwd(), '.')

      // If module is in the base manifest, return its ID
      if (manifestFileContents && manifestFileContents.modules[sourcePath]) {
        return manifestFileContents.modules[sourcePath].id
      }

      // Otherwise, get it from the map or create a new ID
      if (!fileToIdMap.has(path)) {
        fileToIdMap.set(path, nextId)
        nextId += 1
      }
      return fileToIdMap.get(path)
    }
  }

  function getNextIdAfterBaseManifest(manifestFileContents) {
    return Object.keys(manifestFileContents.modules).reduce((id, key) => {
      if (manifestFileContents.modules[key].id > id) {
        return manifestFileContents.modules[key].id
      }
      return id
    }, 0) + 1
  }
}
```

With the sample module ID factory above, the output looks like the following, where defined module IDs start at a higher number to avoid the base module IDs, but may depend on modules in the base bundle (lower numbers).

```javascript
...
__d(function(r,o,t,i,n){t.exports=r.ErrorUtils},551,[]);
__d(function(n,t,o,r,u){'use strict';var e,c=t(u[0]);e=c.now?function(){return c.now()}:function(){return Date.now()},o.exports=e},552,[553]);
...
__d(function(e,t,r,s,l){'use strict'; ...},564,[18,565,27]);
...
```

### DIFF
--- a/packages/metro/src/Bundler/index.js
+++ b/packages/metro/src/Bundler/index.js
@@ -97,7 +97,6 @@ type Options = {|
   +assetServer: AssetServer,
   +blacklistRE?: RegExp,
   +cacheVersion: string,
-  +createModuleIdFactory?: () => (path: string) => number,
   +enableBabelRCLookup: boolean,
   +extraNodeModules: {},
   +getPolyfills: ({platform: ?string}) => $ReadOnlyArray<string>,
@@ -123,7 +122,7 @@ type Options = {|
 
 class Bundler {
   _opts: Options;
-  _getModuleId: (path: string) => number;
+  _getModuleId: ({path: string}) => number;
   _transformer: Transformer;
   _resolverPromise: Promise<Resolver>;
   _projectRoots: $ReadOnlyArray<string>;
@@ -156,7 +155,7 @@ class Bundler {
       transformModuleHash,
     ];
 
-    this._getModuleId = (opts.createModuleIdFactory || createModuleIdFactory)();
+    this._getModuleId = createModuleIdFactory();
 
     let getCacheKey = (options: mixed) => '';
     if (opts.transformModulePath) {
--- a/packages/metro/src/Bundler/index.js
+++ b/packages/metro/src/Bundler/index.js
@@ -97,6 +97,7 @@ type Options = {|
   +assetServer: AssetServer,
   +blacklistRE?: RegExp,
   +cacheVersion: string,
+  +createModuleIdFactory?: () => (path: string) => number,
   +enableBabelRCLookup: boolean,
   +extraNodeModules: {},
   +getPolyfills: ({platform: ?string}) => $ReadOnlyArray<string>,
@@ -122,7 +123,7 @@ type Options = {|
 
 class Bundler {
   _opts: Options;
-  _getModuleId: ({path: string}) => number;
+  _getModuleId: (path: string) => number;
   _transformer: Transformer;
   _resolverPromise: Promise<Resolver>;
   _projectRoots: $ReadOnlyArray<string>;
@@ -155,7 +156,7 @@ class Bundler {
       transformModuleHash,
     ];
 
-    this._getModuleId = createModuleIdFactory();
+    this._getModuleId = (opts.createModuleIdFactory || createModuleIdFactory)();
 
     let getCacheKey = (options: mixed) => '';
     if (opts.transformModulePath) {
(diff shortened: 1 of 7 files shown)


jeanlauliac:
@rafeca does the DeltaBundler have any requirement on the IDs, like being contiguous? Otherwise this look reasonable to me. I'd prefer to have a good solution built-in for stable ID, but I'm okay allowing custom IDs in the meantime.

Before landing this, let's add some unit test for DeltaBundler related to this feature.

packages/metro/src/Bundler/index.js
```diff
@@ -155,7 +156,7 @@ class Bundler {
       transformModuleHash,
     ];
 
-    this._getModuleId = createModuleIdFactory();
+    this._getModuleId = (opts.createModuleIdFactory || createModuleIdFactory)();
```
rafeca: You can safely delete these 3 occurrences here, I forgot to remove them when deprecating the old bundler 😄

packages/metro/src/DeltaBundler/DeltaTransformer.js
```diff
@@ -108,7 +108,7 @@ class DeltaTransformer extends EventEmitter {
     // different bundles to share the same ids (on HMR, where we need to patch
     // the correct module).
     this._getModuleId = this._bundleOptions.isolateModuleIDs
-      ? createModuleIdFactory()
+      ? (bundleOptions.createModuleIdFactory || createModuleIdFactory)()
```
rafeca: Note that with this, you're going to use the custom `createModuleIdFactory()` only when building prod bundles (for development, `isolateModuleIDs` is usually set to `false`.

That is good, since for development we need to ensure that the same `getModuleId` function instance is used for each `DeltaTransformer` instance (to keep HMR working correctly)

rafeca:
The only requirement of the Delta Bundler is to have integer moduleIds (due to RAM bundles), which this PR satisfies, so it looks good from my point of view.

@AdamLiechty is your plan to commit the `baseManifest.json` file in source control and keep it updated by CI?

As @jeanlauliac mentions, it would be good to have some tests here (the DeltaTransformer has no unit tests yet, so if you could create one test for this feature that would be awesome 😃 ).

jeanlauliac:
> The only requirement of the Delta Bundler is to have integer moduleIds (due to RAM bundles), which this PR satisfies, so it looks good from my point of view.

Right, though sparse IDs will not work with RAM bundle however. But it's all fine as long as the custom function generates contiguous ID, or if RAM bundles are not used 😄

AdamLiechty:
Thanks @rafeca @jeanlauliac . I've added a simple test for this setting to `DeltaTransformer`.  After spending more time, I wasn't able to get additional mocks set up correctly to test other DeltaTransformer functionality.

rafeca:
Thanks for adding the test @AdamLiechty! This looks good to me 👍

packages/metro/src/DeltaBundler/__tests__/DeltaTransformer-test.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
```
rafeca: Can you add the license section at the top of the file (Plus the @format and @emails annotations), like for example https://github.com/facebook/metro/blob/master/packages/metro/src/__tests__/index-test.js#L1-L12 ?

AdamLiechty: Done; thanks, @rafeca

PR #107, opened 2017-12-07
Fix 404 on front page 'Learn more' link

pelmers:
Currently index.js points to snapshot-testing which doesn't exist. Perhaps the intent was to link to the API page so that is what I changed.

front page 'learn more' link should work


cpojer:
Thanks for fixing it. As you can see, our docs are pretty sparse and you might have discovered that the website scaffold was copied from Jest :D

PR #118, opened 2018-01-11
Add back support for the assetPlugin option

ide:
**Summary**
Metro used to have support for "asset plugins", which allowed developers to specify arbitrary JS modules that could export a function for adding more fields to asset data objects. Some of this functionality was removed in the delta bundler work -- this PR adds it back.

**Test plan**
Made existing unit tests pass and added unit tests to test asset plugin behavior. Also tested E2E in a React Native project by adding `assetPlugin=/path/to/pluginModule` to a JS bundle URL and ensuring that the plugin ran.

### DIFF
--- a/packages/metro/src/Assets/__tests__/Assets-test.js
+++ b/packages/metro/src/Assets/__tests__/Assets-test.js
@@ -19,9 +19,12 @@ const {getAssetData, getAsset} = require('../');
 const crypto = require('crypto');
 const fs = require('fs');
 
+const mockImageWidth = 300;
+const mockImageHeight = 200;
+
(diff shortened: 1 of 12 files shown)


ide:
create-react-native-app for RN 0.52 depends on this change, which is part of the motivation for this PR.

ide:
Thanks for the review and updating the Jest config in RN too! Would it be possible to publish 0.24.4 to npm so that installing RN 0.52 (which depends on ^0.24.1) will get this update?

rafeca:
@ide I've just published v0.24.4 to npm, let me know if there is any issue

PR #124, opened 2018-01-15
Fix RN redbox messages for syntax errors by including error messages in payload

ide:
**Summary**
With RN 0.52, when there was a redbox due to a syntax error in a source file (with regular, non-delta bundler), the redbox would say just "No message provided". The JSON that Metro sent to RN did not include a "message" field because `JSON.stringify(error)` does not include `message`.

**Test Plan**
Add a syntax error to one of the files in RNTester's JS and load the RNTester app (from RN master). See that the redbox now says there was a transform error with the syntax error's location.

Also tested adding a syntax error with HMR enabled and saw that the error `message` field was set in the payload as expected.

Also added a Jest test to Server-test.js.

### DIFF
--- a/packages/metro/package.json
+++ b/packages/metro/package.json
@@ -47,6 +47,7 @@
     "mkdirp": "^0.5.1",
     "request": "^2.79.0",
     "rimraf": "^2.5.4",
+    "serialize-error": "^2.1.0",
     "source-map": "^0.5.6",
     "temp": "0.8.3",
     "throat": "^4.1.0",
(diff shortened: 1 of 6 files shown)


ide:
This code (JSON.stringifying an error) goes back more than three years so I don't think the original author has the same context. My guess is that `error.description` used to be set and used before, or perhaps something in RN changed recently that now expects Metro to include a `message` field in the JSON body.

@rafeca Do you think you could take a look at this (you most recently worked on error formatting a few months ago)?

ide:
@rafeca Could I ask you to publish 0.24.5 after this lands? I think (hoping...) this is the last commit we need to show TransformError messages in create-react-native-app for RN 0.52.

rafeca:
@ide we've just published 0.24.5 (we wanted to add a few more things to this release, this is why it took a few days). Let me know if this fixes all the issues in create-react-native-app

ide:
Sorry for the late reply - 0.24.5 successfully fixed a specific error message when there was a syntax error with HMR.

PR #125, opened 2018-01-15
Use `error.message` to set the `description` property of a nested error payload

ide:
Depends on #124.

---

**Summary**
Metro reports errors using a JSON payload that has an `errors` array. Each item in this array has a `description` field. For transform errors, this field was set using the value in `error.description` -- however, JS Error objects only have a `message` field. (Grepping the Metro code, no errors (except in one test) ever get a `description` field.) This commit uses `error.message` instead of `error.description` when creating JSON payloads.

```
$ git grep description -- 'packages/**/*.js'
packages/metro/src/JSTransformer/__tests__/Transformer-test.js:        babelError.description = message;
packages/metro/src/lib/formatBundlingError.js:    description: string,
packages/metro/src/lib/formatBundlingError.js:): {type: string, message: string, errors: Array<{description: string}>} {
packages/metro/src/lib/formatBundlingError.js:      errors: [{description: message}],
packages/metro/src/lib/formatBundlingError.js:        description: error.message,
packages/metro/src/node-haste/__tests__/Module-test.js:  description: "A require('foo') story",
```

**Test Plan**
Added a unit test to check that the description field is set for transform errors (with the delta bundler).

Also in a test RN app, inspected the error payload that is received by RN when there's a syntax error with HMR turned on and verified that `data.body.errors[0].description` was set.

### DIFF
--- a/packages/metro/package.json
+++ b/packages/metro/package.json
@@ -47,6 +47,7 @@
     "mkdirp": "^0.5.1",
     "request": "^2.79.0",
     "rimraf": "^2.5.4",
+    "serialize-error": "^2.1.0",
     "source-map": "^0.5.6",
     "temp": "0.8.3",
     "throat": "^4.1.0",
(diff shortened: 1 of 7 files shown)


ide:
@mjesun This is one place (the last place, I believe) where `error.description` was left over.

PR #136, opened 2018-02-07
Pass numWorkers instead of maxWorkers to jest-worker

misogl:
**Summary**

Jest Worker accepts parameter `numWorkers`, not `maxWorkers`, see:
https://github.com/facebook/jest/tree/master/packages/jest-worker#numworkers-number-optional

**Test plan**


mjesun:
Good catch! Could you change the parameter name in the function so we keep the shorthand version when passing the object? 😉

Don't worry about the CI failures, we'll take care of that when importing it internally. Thanks for the contribution! 🙂

mjesun:
Argh, our internal importer is broken. I've got to import this manually... 😒

PR #138, opened 2018-02-13
Replaces glob with fast-glob

jsdario:
Simple and direct drop-in replacement for a library that claims to have 1.5x-10x better performance than glob.

Repo https://github.com/mrmlnc/fast-glob
Benchmarks https://gist.github.com/mrmlnc/f06246b197f53c356895fa35355a367c


cpojer:
I don't think it's necessary for this build script to adopt a different glob implementation without a clear reason. Theoretical faster performance is great, but this script is only invoked during releases and not generally slow. Thanks for submitting a PR!

PR #140, opened 2018-02-16
Add `safari10: true` to Uglify options

jamesreggio:
**Summary**

JSCore has known issues with block-scoped variables which were not addressed until iOS 11.

These problems are generally masked by an over-zealous Babel configuration that removes all block scoping. If you modernize your Babel configuration to retain some ES6 features, you'll quickly run into
'illegal duplicate declaration' issues on Android and iOS 10.

This is a quick-and-simple fix that ensures the code emitted by Uglify will work properly on these platforms.

**Test Plan**

This fixes issues I was encountering in the wild. Let me know if you feel this needs automated testing.

### DIFF
--- a/packages/metro-minify-uglify/src/minifier.js
+++ b/packages/metro-minify-uglify/src/minifier.js
@@ -36,7 +36,12 @@ function withSourceMap(
 
 function minify(inputCode: string, inputMap: ?BabelSourceMap) {
   const result = uglify.minify(inputCode, {
-    mangle: {toplevel: true},
+    mangle: {
+      toplevel: true,
+      // JSC has issues with block-scoped variables.
+      // https://github.com/mishoo/UglifyJS2/issues/1753#issuecomment-324814782
+      safari10: true,
+    },
     output: {
       ascii_only: true,
       quote_style: 3,


rafeca:
Thanks @jamesreggio !

I'm going to import the PR to do some thorough internal testing, since changes in the minification process can lead to substantial perf differences and we want to be sure that they don't cause regressions.

BTW, @qfox just landed [a change](https://github.com/facebook/metro/commit/5aa19c7626cf8cb3bac58e17d8276f06f451081f) to allow fully configuration of the minification step, so this kind of tweaks will be possible on user land soon :)

qfox:
We currently require our code to be transformed to ES5 semantics (one important reason is the one you pointed out except more generically) and I don't think that's going to change any time soon. As such ES6 compat efforts have to be significant for us to consider. And minification settings are especially delicate.

As @rafeca pointed out, I've introduced a way for minifiers to be plug-and-playable recently. If you need/want specific config settings for minification then that is probably the best way to go about it. You should get zero loss (you can start from the `metro-minify-uglify` package at https://www.npmjs.com/package/metro-minify-uglify) and have full control beyond that.

I'll leave it to you or @rafeca to close this out. Let us know if you feel this isn't sufficient for your use case.

PR #141, opened 2018-02-16
Minor fixes to aid in the adoption of Babel 7

jamesreggio:
**Summary**

With great effort, I was able to get Babel 7 to work with Metro 0.26.0. I should probably write about it, but before I can do that, I'd like for you to consider these two changes that I made locally to get Metro to play nicely with the new Babel.

The first change (for the external helpers plugin) corrects what appears to be a minor oversight.

The second change no-ops a code-path that I cannot imagine has ever worked. The maintainer of [`babel-plugin-react-transform`](https://github.com/gaearon/babel-plugin-react-transform#this-project-is-deprecated) insists its deprecated, so there's no point in making it work with Babel 7. I'm not sure what this means for HMR, but HMR is something I can do without if it means I can access the new transforms in Babel 7.

**Test plan**

Something that seemed totally borked is now less broken 😇

### DIFF
--- a/packages/metro/src/babel-bridge.js
+++ b/packages/metro/src/babel-bridge.js
@@ -107,51 +107,11 @@ module.exports = {
   getPreset: IS_BABEL7 ? getPreset7 : getPreset6,
 };
 
+// Plugin not functional in Babel 7, is deprecated, and no replacement exists.
 function makeMakeHMRConfig7() {
-  // from: babel-preset-react-native/configs/hmr
-  /**
-   * Copyright (c) 2015-present, Facebook, Inc.
-   * All rights reserved.
-   *
-   * This source code is licensed under the BSD-style license found in the
-   * LICENSE file in the root directory of this source tree. An additional grant
-   * of patent rights can be found in the PATENTS file in the same directory.
-   */
-  'use strict';
-
-  var path = require('path');
-  var hmrTransform = 'react-transform-hmr/lib/index.js';
-  var transformPath = require.resolve(hmrTransform);
-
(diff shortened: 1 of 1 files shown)


rafeca:
Thanks for contributing!

/cc @qfox who's driving the migration to Babel 7

simonbuchan:
Well there is https://github.com/gaearon/react-hot-loader which has `react-hot-loader/babel`, but given it requires specific code changes anyway it seems like the user might as well add it themselves.

jamesreggio:
@simonbuchan, `react-hot-loader` claims to have problems with [React Native](https://github.com/gaearon/react-hot-loader#react-native).

There has likely been little incentive to resolve them since "React Native supports hot reloading natively" — however, the 'native' support is actually dependent upon the older `babel-plugin-react-transform`. Hence the problem.

simonbuchan:
It's possible it was just an issue with both transforms happening, AFAICT RHL "only" proxies the `react` component type, so it shouldn't matter who's rendering it (`react-dom` vs `react-native`).

But I agree with pulling it from the default config anyway, since it's so easy to add yourself now, and has had weird bugs in the past. (maybe later re-add with a config option? e.g. `presets: [['react-native', { hmr: true }]]`)

qfox:
Since Babel 7 landed I think this can be closed (cc @rafeca). 

I do want to thank you for your effort. It took quite a bit more to get things working I hope you like the result.

rafeca:
Closing this PR, since metro is currently based on Babel 7. Thanks a lot @jamesreggio for your contribution!

Issue #142, opened 2018-02-18
metro 0.25.1 can not be used from npm


vovkasm:
**What is the current behavior?**
Can't install metro@0.25.1 or have it as dependency.

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**
```sh
mkdir fuck-metro && cd fuck-metro
npm init # all answers default
npm i -P metro@0.25.1 
npm ERR! code EUNSUPPORTEDPROTOCOL
npm ERR! Unsupported URL Type "npm:": npm:babylon@^7.0.0-beta
```

**What is the expected behavior?**
Package `metro` should successfully be installed into `node_modules` and recorded in `package.json`

**Please provide your exact Metro configuration and mention your Metro, node, yarn/npm version and operating system.**
OS: macOS Sierra 10.12.6 (16G1212)
```
npm --versions
{ 'fuck-metro': '1.0.0',
  npm: '5.6.0',
  ares: '1.13.0',
  cldr: '32.0.1',
  http_parser: '2.7.0',
  icu: '60.2',
  modules: '59',
  napi: '2',
  nghttp2: '1.29.0',
  node: '9.5.0',
  openssl: '1.0.2n',
  tz: '2017c',
  unicode: '10.0',
  uv: '1.19.1',
  v8: '6.2.414.46-node.18',
  zlib: '1.2.11' }
```

Error exactly here: [packages/metro/package.json](https://github.com/facebook/metro/blob/v0.25.1/packages/metro/package.json#L56) and it still in master.

And in case one can't find it, I provide link to documentation of `package.json` format: https://docs.npmjs.com/files/package.json#urls-as-dependencies

koloff:
+1

vovkasm:
@davidaurelio As a reviewer of breaking change, can you comment on this? Thanks!

vovkasm:
@qfox does commit https://github.com/facebook/metro/commit/9ac44657bcae49a6fb82d603a0120338653a7558 is your?

jeanlauliac:
A way to fix this, I think, is to create a published package ex. `metro-babylon7` that itself depends on `babylon@7` and re-export it verbatim, then depend on `babylon@6` and `metro-babylon7` from `metro`.

vovkasm:
@jeanlauliac Looks like solution, should I create pull request?

vovkasm:
[...]

PR #143, opened 2018-02-24
Fix npm deps

vovkasm:
**Summary**

This should fix install of the metro with npm (issue #142).
We can't have both babylon@6 and babylon@7 as deps for metro package, so as @jeanlauliac suggested in https://github.com/facebook/metro/issues/142#issuecomment-368217279 lets create wrapper-package `metro-babylon7` and reexport babylon@7 from it.

**Test plan**

I do manually `yarn && yarn test` - all checks passed.

### DIFF
--- a/packages/metro-babylon7/src/index.js
+++ b/packages/metro-babylon7/src/index.js
@@ -1,0 +1,3 @@
+const babylon = require('babylon');
+
+module.exports = babylon;
--- a/packages/metro/package.json
+++ b/packages/metro/package.json
@@ -53,7 +53,6 @@
     "babel-preset-react-native": "^4.0.0",
     "babel-register": "^6.24.1",
     "babylon": "^6.18.0",
-    "babylon7": "npm:babylon@^7.0.0-beta",
     "chalk": "^1.1.1",
     "concat-stream": "^1.6.0",
     "connect": "^3.6.5",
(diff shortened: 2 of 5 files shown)


packages/metro-babylon7/src/index.js
```diff
@@ -0,0 +1,3 @@
+const babylon = require('babylon');
+
+module.exports = babylon;
```
rafeca: Can you please add the license header to the file?

```
/**
 * Copyright (c) 2015-present, Facebook, Inc.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 *
 * @flow
 * @format
 */
```

packages/metro/package.json
```diff
@@ -53,7 +53,6 @@
     "babel-preset-react-native": "^4.0.0",
     "babel-register": "^6.24.1",
     "babylon": "^6.18.0",
-    "babylon7": "npm:babylon@^7.0.0-beta",
```
rafeca: You should still add a dependency from `metro` to `metro-babylon7`:

```
"metro-babylon7": "0.27.0",
```

vovkasm: Opps, thanks! Fixed!

packages/metro-babylon7/src/index.js
vovkasm: Fixed!

PR #149, opened 2018-03-09
ignore build artifacts

jeanlauliac:
When running `yarn` it compiles all the packages files into `build` folders. We should probably ignore this to avoid them ending up checked-in on the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /coverage
 lerna-debug.log
 npm-debug.log*
+packages/*/build/


mjesun:
Can you make sure that when releasing with NPM, these files will be put into the package? AFAIK NPM uses `.gitignore` if no `.npm<something>` is present.

jeanlauliac:
> Can you make sure that when releasing with NPM, these files will be put into the package? AFAIK NPM uses .gitignore if no .npm<something> is present.

I think it's doesn't matter, because we don't actually release the `build` folders, we rename them to `src` just for publishing, for compatibility with our monorepo.

PR #153, opened 2018-03-13
Fix runtime error

vikrampunchh:
**Summary**

**Test plan**

### DIFF
--- a/packages/metro/src/Bundler/index.js
+++ b/packages/metro/src/Bundler/index.js
@@ -122,7 +122,7 @@ class Bundler {
     opts.projectRoots.forEach(verifyRootExists);
 
     this._opts = opts;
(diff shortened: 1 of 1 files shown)


mjesun:
Based on [this line](https://github.com/facebook/metro/blob/master/packages/metro/src/Bundler/index.js#L83), `cacheStores` can never be `null` or `undefined`, so we don't need to verify for its validity. In addition, `cacheStores` is passed from `Server`, [which _also_ has it as a required field](https://github.com/facebook/metro/blob/master/packages/metro/src/Server/index.js#L73). What is probably happening is that you are not passing that option; providing an empty array (`[]` should be enough).

There is also a lack of summary and test plan within the commit, so we can't go ahead with that.

PR #162, opened 2018-04-19
fix config in babel 7

gengjiawen:
**Summary**

* Fix ast is null in babel 7, see https://github.com/facebook/react-native/commit/f8d6b97140cffe8d18b2558f94570c8d1b410d5c#r28647044.

* change sourceType to "unambiguous", see https://github.com/facebook/react-native/commit/f8d6b97140cffe8d18b2558f94570c8d1b410d5c#r28569395.

With those change, react native can fix all jest test.
```
Test Suites: 2 skipped, 100 passed, 100 of 102 total
Tests:       4 skipped, 553 passed, 557 total
Snapshots:   37 passed, 37 total
Time:        78.455s
Ran all test suites.
Done in 79.46s.
```

**Test plan**

pass react native test-ci task.

### DIFF
--- a/packages/metro/src/transformer.js
+++ b/packages/metro/src/transformer.js
@@ -111,6 +111,10 @@ function buildBabelConfig(filename, options, plugins?: BabelPlugins = []) {
     config = Object.assign({}, config, hmrConfig);
   }
 
+  if (process.env.BABEL_VERSION === '7') {
+    config = Object.assign({}, config, {ast: true});
+  }
+
   return Object.assign({}, babelRC, config);
 }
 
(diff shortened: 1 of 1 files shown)


gengjiawen:
@qfox Can you take a look at this pr, set options on preprocessor won't work.

qfox:
@gengjiawen this problem should be resolved once the lockdown lands. For now your workaround is to manually and explicitly install babel at 7.0.0-beta.40

These api changes were added later (which is why they slipped under our radar). The lockdown fixes will hopefully be published in a new version today.

Note that if you still require the BABEL_VERSION flag to be set at all then you're using an older version of Metro. Regardless, to fix the option problem you'll need to wait a few hours. Working hard on it.

rafeca:
@qfox is this PR still relevant with your latest changees?

PR #163, opened 2018-04-20
add missing dependency

gengjiawen:
**Summary**

The dependency are missing for react native.

**Test plan**


rafeca:
Hey @gengjiawen ! these dependencies are already in `babel-preset-react-native` package, which is the one that's actually using them: https://github.com/facebook/react-native/blob/master/babel-preset/package.json#L39-L40

We're going to release a new version of `babel-preset-react-native` probably today with these dependencies

gengjiawen:
But jest preprocessor use the packages too, how to fix that ?

qfox:
These packages are part of the babel-preset-react-native .. preset (https://github.com/facebook/react-native/blob/master/babel-preset/package.json#L39). The only problem is that Metro doesn't depend on the correct version yet. We're ironing things out and will hopefully publish a fix for these discrepancies today. 

This can be closed.

gengjiawen:
@rafeca @qfox I recheck the code again, still not get it. I know `babel-preset-react-native` has the packages in creating this pr. But react native [package.json](https://github.com/facebook/react-native/blob/master/package.json) not depend on babel-preset-react-native. So the jest preprocessor will still fails for lack of the packages.

gengjiawen:
I see, it's in here https://github.com/facebook/metro/blob/master/packages/metro/package.json. But in this way it looks the babel packages are declared twice.

qfox:
> it looks the babel 7 packages are declared twice.

Please explain. If you're talking about the Babel 6 packages, that's a matter of cleanup. This upgrade was complex and there was no need to drop these packages for it. A smaller diff is better. However, this is explicitly version 6 so I'm not sure what duplication you're referring to.

gengjiawen:
I mean babel 7. You can see https://github.com/facebook/metro/blob/master/packages/metro/package.json and https://github.com/facebook/metro/blob/master/packages/metro/package.json.

qfox:
(same link)

gengjiawen:
@qfox fixed.

qfox:
Sure. They are different packages. Metro can be used without ReactNative. I'm not familiar enough with the preset to give better insights on that, I just bumped it to their version 7 counterparts.

PR #173, opened 2018-05-22
Handle scoped modules from extraNodeModules config

sfrdmn:
**Summary**

Metro allows configuring `extraNodeModules` for additional module resolution mappings. It chokes on scoped module names (e.g. `@org/module`), because it doesn't compute the path correctly

**Test plan**

```
// In metro.config.js
extraNodeModules: {
  '@org/module': path.resolve(__dirname, "wherever")
}

// In consuming code
import module from '@org/module'
```

Assuming the module is neither in `node_modules` nor comes from the Haste map, the former code will fail without this patch

### DIFF
--- a/packages/metro-resolver/src/resolve.js
+++ b/packages/metro-resolver/src/resolve.js
@@ -106,7 +106,7 @@ function resolve(
     let bits = path.normalize(moduleName).split(path.sep);
     let packageName;
     // Normalize packageName and bits for scoped modules
-    if (bits[0] && bits[0][0] === '@') {
+    if (bits.length >= 2 && bits[0].startsWith('@')) {
       packageName = bits.slice(0, 2).join('/');
       bits = bits.slice(1);
     } else {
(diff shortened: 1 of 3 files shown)


packages/metro-resolver/src/resolve.js
```diff
@@ -103,8 +103,15 @@ function resolve(
   const extraPaths = [];
   const {extraNodeModules} = context;
   if (extraNodeModules) {
-    const bits = path.normalize(moduleName).split(path.sep);
-    const packageName = bits[0];
+    let bits = path.normalize(moduleName).split(path.sep);
+    let packageName;
+    // Normalize packageName and bits for scoped modules
+    if (bits[0] && bits[0][0] === '@') {
```
rafeca: Can you do `if (bits.length >= 2 && bits[0].startsWith('@')) {` here?

rafeca:
This is awesome! Thanks for the fix!

Can you please add a simple test for this? We currently only have integration tests for the resolution logic, but it should be easy to add a new test using the following one as a base: https://github.com/facebook/metro/blob/master/packages/metro/src/DeltaBundler/__tests__/traverseDependencies-integration-test.js#L1506 ?

sfrdmn:
Updated, but tests failing on master it seems?

rafeca:
Thanks! Looks good!

Yup, we have an unrelated issue with the tests on node v6

PR #186, opened 2018-06-24
Update to Docusaurus 1.3.2

endiliey:
## Summary
Docusaurus 1.3.2 brings about some [backwards-incompatible CSS](https://github.com/facebook/Docusaurus/blob/master/CHANGELOG.md#breaking-changes) changes. So the team is helping the major users update 😄

Other
- Activated `cleanUrl` / pretty url. All url is now accessible with & without .html extension

cc @JoelMarcey @yangshun 

## Test Plan
Ran the site locally and visited every page.


yangshun:
The node 6 tests are failing but I don't think they're related to this PR. It's already broken on master.

rafeca:
Yay! thanks for the contribution!

PR #193, opened 2018-07-09
Migrate away from deprecated new Buffer

thymikee:
**Summary**

Gets rid of Buffer warnings on Node 10.
`new Buffer()` constructor was deprecated in Node 10. Since Metro is supporting Node 8+ now, we can migrate to `Buffer.from` and `Buffer.alloc` (added in v5.10). 

**Test plan**

Existing tests + Flow should pass.

### DIFF
--- a/packages/metro-memory-fs/src/index.js
+++ b/packages/metro-memory-fs/src/index.js
@@ -382,14 +382,14 @@ class MemoryFs {
     const fd = this._open(pathStr(filePath), flag || 'r');
     const chunks = [];
     try {
-      const buffer = new Buffer(1024);
+      const buffer = Buffer.alloc(1024);
       let bytesRead;
       do {
         bytesRead = this.readSync(fd, buffer, 0, buffer.length, null);
         if (bytesRead === 0) {
           continue;
         }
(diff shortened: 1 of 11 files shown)


packages/metro-memory-fs/src/index.js
```diff
@@ -638,12 +638,12 @@ class MemoryFs {
...
-          encoding?: ?Encoding,
+          autoClose?: boolean,
+          encoding?: Encoding,
           fd?: ?number,
-          flags?: ?string,
-          mode?: ?number,
-          start?: ?number,
+          flags?: string,
+          mode?: number,
+          start?: number,
```
thymikee: Adjusted based on https://nodejs.org/api/fs.html#fs_fs_createwritestream_path_options. Replacing `new Buffer` with `Buffer.from` resulted in type error for `opts.start` in https://github.com/facebook/metro/pull/193/files#diff-edc370c71b61b9fe317068a63d935be9R1164

rafeca:
Thanks for fixing this! We used to support node v6, but since RN does not support it anymore we can get also drop its support in metro 😁

PR #194, opened 2018-07-10
Set license to MIT in package.json files

mtcmorris:
**Summary**

This patch sets the license in package.json to "MIT" (in line with [LICENSE](https://github.com/facebook/metro/blob/master/LICENSE)) so it no longer comes up as "UNKNOWN" for the metro packages when running `yarn licenses list`


rafeca:
Thanks for the contribution!

PR #198, opened 2018-07-12
Use babel runtime instead of relying on global babelHelpers and regenerator

janicduplessis:
**Summary**

The RN transformer currently relies on the enviroment providing babelHelpers and regeneratorRuntime as globals by using 'babel-external-helpers'. This wasn't really a problem before since helpers were stable and we could maintain our copy easily but it seems like there are more now with babel 7 and it makes sense to include only those used by the app.

This is exactly what @babel/transform-runtime does. It will alias all helpers and calls to regeneratorRuntime to files in the @babel/runtime package.

This will solve issues like this https://github.com/facebook/react-native/issues/20150 caused by missing babelHelpers. This solution also avoids bloating babelHelpers to fix OSS issues like the one linked before.

**Test plan**

- Updated tests so they all pass.
- Tested that it actually works by applying the changes locally in an RN app.
- Added a test for async functions, to make sure regenerator is aliased properly and doesn't depend on the global.
- Made sure require-test.js still fails if the require implementation contains babel helpers (by adding an empty class in the file).

### DIFF
--- a/packages/metro/package.json
+++ b/packages/metro/package.json
@@ -17,7 +17,6 @@
     "@babel/generator": "^7.0.0",
     "@babel/helper-remap-async-to-generator": "^7.0.0",
     "@babel/parser": "^7.0.0",
-    "@babel/plugin-external-helpers": "^7.0.0",
     "@babel/plugin-proposal-class-properties": "^7.0.0",
     "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
     "@babel/plugin-syntax-dynamic-import": "^7.0.0",
--- a/packages/metro/src/reactNativeTransformer.js
+++ b/packages/metro/src/reactNativeTransformer.js
@@ -12,7 +12,6 @@
 'use strict';
 
 const crypto = require('crypto');
-const externalHelpersPlugin = require('@babel/plugin-external-helpers');
 const fs = require('fs');
 const inlineRequiresPlugin = require('babel-preset-fbjs/plugins/inline-requires');
 const json5 = require('json5');
@@ -28,7 +27,6 @@ type ModuleES6 = {__esModule?: boolean, default?: {}};
 
 const cacheKeyParts = [
   fs.readFileSync(__filename),
-  require('@babel/plugin-external-helpers/package.json').version,
   require('babel-preset-fbjs/package.json').version,
 ];
 
@@ -127,7 +125,7 @@ function buildBabelConfig(filename, options, plugins?: BabelPlugins = []) {
   let config = Object.assign({}, babelRC, extraConfig);
 
   // Add extra plugins
-  const extraPlugins = [externalHelpersPlugin];
+  const extraPlugins = [];
 
   if (options.inlineRequires) {
     extraPlugins.push(inlineRequiresPlugin);
--- a/packages/metro-react-native-babel-preset/package.json
+++ b/packages/metro-react-native-babel-preset/package.json
@@ -40,6 +40,7 @@
     "@babel/plugin-transform-react-jsx": "^7.0.0",
     "@babel/plugin-transform-react-jsx-source": "^7.0.0",
     "@babel/plugin-transform-regenerator": "^7.0.0",
+    "@babel/plugin-transform-runtime": "^7.0.0",
     "@babel/plugin-transform-shorthand-properties": "^7.0.0",
     "@babel/plugin-transform-spread": "^7.0.0",
     "@babel/plugin-transform-sticky-regex": "^7.0.0",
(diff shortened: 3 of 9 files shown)


janicduplessis:
Not sure how we should go about including the @babel/runtime dependency, it could be added here or in RN, I think RN makes more sense since that is where the babelHelpers / regeneratorRuntime are defined atm.

packages/metro/package.json
```diff
@@ -15,7 +15,6 @@
   "dependencies": {
     "@babel/core": "7.0.0-beta.48",
     "@babel/generator": "7.0.0-beta.48",
-    "@babel/helper-remap-async-to-generator": "7.0.0-beta.48",
```
janicduplessis: I was checking deps and those were not used in the metro package.

packages/metro/src/reactNativeTransformer.js
```diff
@@ -28,7 +27,6 @@ type ModuleES6 = {__esModule?: boolean, default?: {}};
 
 const cacheKeyParts = [
   fs.readFileSync(__filename),
-  require('babel-plugin-external-helpers/package.json').version,
```
janicduplessis: Should we still keep the babel version as part of that cache key?

rafeca:
Thanks for doing this! This looks like the right approach, much more robust 👍 I think it's better to have this in metro tbh.

I guess that you're going to remove https://github.com/facebook/react-native/blob/master/rn-get-polyfills.js#L21 once this change gets propagated to RN?

I'm going to import this PR to our internal systems to check that it does not break anything or it does not cause any build size regression

janicduplessis:
@rafeca Yes we’ll be able to remove that as well as the global regenerator we add in InitializeCore.

newyankeecodeshop:
This looks great! Will this make it possible for RN apps which target a modern JavaScriptCore (e.g. iOS 10.3+) to use a custom babel config that does not even need the regenerator runtime? (because generator functions are natively supported?)

newyankeecodeshop:
Any movement on this? It looks like a great addition to decouple babel dependencies from React Native itself.

ItsNoHax:
Will this fix the whole regeneratorRuntime missing issue?

loganfsmyth:
@janicduplessis Babel maintainer here. This would be great overall, since I think `external-helpers` isn't that useful now that `@babel/runtime` is trimmed down. 

I'll also say that the 7.0.0 `external-helpers` transform also accepts the same whitelist option that `babel.buildExternalHelpers` does, https://github.com/babel/babel/blob/master/packages/babel-plugin-external-helpers/src/index.js#L10, so if there are issues with getting this landed in the short term, a quick fix would be to pass the list of helpers that RN builds and pass it to the transform. That way it will fall back to injecting the helpers into each individual file, instead of inserting references to helpers that don't exist.

I'd expect this to become more and more of a problem since we'll likely add new helpers over 7.x's lifetime and we assume that, if you inject a reference, you're saying that you actually have that helper available.

wangshangming:
Any movement of this?

rafeca:
Hey folks, when I tried to import this PR it caused a few errors in our internal builds, I'm going to try again next week (probably tweaking it a bit) and see if I can squeeze it in 😃

janicduplessis:
@rafeca Cool, I just rebased on master to fix conflicts.

wangshangming:
@rafeca Will this cherry-picked in next release?

janicduplessis:
@wangshangming This has not been merged to master yet, when it does we'll know when it will make it in a release.

rafeca:
I'm still dealing with some internal issues in order to be able to merge this PR, but making slow but steady progress.

To give some context, these are the two main issues found:

* The current `[babelHelpers](https://github.com/facebook/react-native/blob/master/Libraries/polyfills/babelHelpers.js)` that we use are from `babel-7.0.0-beta.47`. Between that beta and latest stable differ slightly in behaviour, making some of our tests fail.
* Because `@babel/runtime` adds a bunch of requires instead of calling the static methods, tests that use `jest.autoMock()` stop working since they also mock the runtime. Unfortunately we still have lots of old tests with that so I'm dealing with them.

Lastly, once I get everything working I'll do some performance measurement to check that the startup times of RN apps does not regress (adding new `require` calls adds a small overhead, and this together with inline requires could cause perf regressions).

I'll keep you informed

LinusU:
> [...] adding new `require` calls [...]

This might not be relevant at all, but just so you know, you can pass `useESModules: true` to `@babel/plugin-transform-runtime` in order to use `import` instead of `require`. That could potentially be optimized better...

PR #203, opened 2018-07-21
babel-preset: Add "modules" option to allow passing through of ES6/ha…

hedgepigdaniel:
## Motivation

Since the majority of react-native code is written using ES6 modules, it is possible to do static analysis to achieve some new things:
 * Eliminate dead code, since if an ES6 export is never imported from the application, it cannot run
 * Identify programmer errors faster, since it is possible to detect and warn when the programmer imports a value that was not exported from a module.

Using [haul](https://github.com/callstack/haul) (which uses babel-preset-react-native by default) is it possible to compile a react-native project with [webpack](https://webpack.js.org/). Webpack natively supports ES6 modules, and will [generate a warning or error](https://github.com/webpack/webpack/pull/4348) if user code imports a value that was not exported. This only works for ES6 imports/exports however, so it does not work if babel transpiles all the imports/exports to CommonJS before webpack bundles them.

**Test plan**

I tested a similar change on `babel-preset-react-native` 4.0.0 for this PR: https://github.com/facebook/react-native/pull/17693

Unfortunately I don't have a project set up that works with babel 7 + react-native, so I haven't directly tested this commit. The change is very simple though - if the `modules` options is not used, there is no change, and if it is and is set to `false`, the commonJs transform is not included in the preset.

### DIFF
--- a/packages/metro-react-native-babel-preset/README.md
+++ b/packages/metro-react-native-babel-preset/README.md
@@ -24,6 +24,21 @@ Then, create a file called `.babelrc` in your project's root directory. The exis
 }
 ```
 
+## Options ##
+
+### `modules`
+
+`"commonjs" | false`, defaults to `"commonjs"`
+
+Enable transformation of ES6 module syntax to CommonJS. E.g:
+```
+{
+  "presets": [
+    ["metro-react-native-babel-preset", { "modules": false }]
+  ]
+}
+```
+
 You can further customize your Babel configuration by specifying plugins and other options. See [Babel's `.babelrc` documentation](https://babeljs.io/docs/usage/babelrc/) to learn more.
 
 ## Help and Support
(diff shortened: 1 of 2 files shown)


rafeca:
Thanks for the contribution! apologies for not taking a look earlier.

We recently added a similar feature, via a `disableImportExportTransform` option in the preset (https://github.com/facebook/metro/blob/17e98bc86272df2f915e12b464492f560283800b/packages/metro-react-native-babel-preset/src/configs/main.js#L103-L105).

Also, we're starting to experiment with native support for ES6 `import`/`export` in Metro: currently it can be activated via `experimentalImportSupport` and this optimizes the imports to not go through the `babelInteropRequireDefault` helper (plus Metro is able to inline them as it does with requires).

Next step is going to be minification of export names + tree shaking 😄

PR #205, opened 2018-07-26
feat: validate config with jest-validate

thymikee:
**Summary**

This PR introduces validation for the new `metro-config`. For now I'm keen on only validating new config, because I can see there's yet unused `convertConfig` which, I guess, defeats the purpose. Let me know otherwise.

This is still work in progress, as I need to fix some stuff upstream (e.g. recursive object validation and pretty-printing). Just letting you know I'm working on it :)

**Test plan**

Added a test.

cc @CompuIves @cpojer


CompuIves:
This is so cool @thymikee, thanks for building this!

> For now I'm keen on only validating new config, because I can see there's yet unused `convertConfig` which, I guess, defeats the purpose. Let me know otherwise.

I agree, we only have `convertConfig` available right now to make the transition from the old config easier, but we're planning to remove this anyways in a near-future version.

thymikee:
Update: https://github.com/facebook/jest/pull/6802

thymikee:
facebook/jest#6802 was merged, so now we wait for new Jest version to be released by @mjesun :)

cpojer:
@thymikee can you rebase this PR with the new Jest version? I'll land it then :)

thymikee:
Looks like I'll need to sort out printing functions in `jest-validate`, `toString()` is not the best idea :D

cpojer:
Let me know once that part is ready so we can move ahead on this PR :)

cpojer:
ping @thymikee let's make it happen? :)

thymikee:
Changed the test to avoid printing functions as it's Jest's responsibility to test the output (which also may slightly change at any time, so the test would be prone to errors). Regarding `jest-validate` I think it may still be nice to see function signature in validation message as it usually doesn't go through istanbul. Of course it may be altered by Babel, but at least the number of arguments will be shown.

The CI failures seem unrelated.

cpojer:
Seems to fail at the import step, no idea why. Retrying.

thymikee:
Can you share the error?

cpojer:
Nothing from your side, I've been fixing this up for the last hour.

PR #212, opened 2018-08-01
Fix GettingStarted rendering

passy:
**Summary**

By having the comment inline, the list breaks.

See https://facebook.github.io/metro/docs/en/getting-started.html

screenshot 2018-08-01 15 19 10

**Test plan**

eyes


rafeca:
Thanks! 👍

Issue #210, opened 2018-08-01
postProcessModules not working now


cliberal:
**Do you want to request a *feature* or report a *bug*?**
probably
**What is the current behavior?**
``` js
// metro.config.js
module.exports = {
  resolver: {
    /* resolver options */
  },
  transformer: {
    /* transformer options */
  },
  serializer: {
    /* serializer options */
    postProcessModules: function(modules, entryFiles) {
      console.log(modules);
      console.log(entryFiles);
    },
  },
  server: {
    /* server options */
  },
};

```
I try to add breakpoint but it doesn't execute the postProcessModules function.

rafeca:
Hi @cliberal ! Thanks for reporting the issue! `postProcessModules` is currently being unused at the moment and we're thinking about deprecating this option (I'm not sure that the API that it exposes is the right one).

Out of curiosity, what are you trying to do with it?

cliberal:
thanks for explanation，I want to skip some common dependencies before generating the final code. 
e.g.
```
postProcessModules: function(modules, entryFiles) {
    return modules.filter(module => !module.path.match(MY_REGEX))
}
```
[https://github.com/facebook/react-native/pull/10804](https://github.com/facebook/react-native/pull/10804)

suppose there‘re several RN applications need to integrate with existing native app. we want to split the common libarary like `react-native`、`polyfills` from the full bundle.  Maybe there's an alternative way I missed it ...

rafeca:
Makes sense! 😀 We're currently working on a new way to customize the output of the bundle. With the new API you are going to be able to write your own serializer which takes the whole dependency graph as an input and returns the bundle as an output (we have all the internals already ready for this, but we are working on a nice API to expose this. As an example, currently [this is the default Metro JS serializer](https://github.com/facebook/metro/blob/master/packages/metro/src/DeltaBundler/Serializers/plainJSBundle.js)).

I'm estimating that we can have this API ready in ~1-2 months, and this is going to allow implementing code splitting quite easily.

In the mean time, code splitting can be done by asking Metro to build the dependency graph and serializing it manually (using the Metro internals). I have [an example of this](https://github.com/rafeca/metro-sample-app/commit/a167bff2b2b6a6fb67efe9b3ed642999a88108db) in my sample app (**disclaimer**: use this on your own, since accessing Metro internals means very probable breakages in the future).

[...]

PR #215, opened 2018-08-05
Add processModuleFilter and remove postProcessModules

cliberal:
**Summary**
remove postProcessModules hook since it is not working now. Add a processModuleFilter hook to skip some modules before code generating. see #210 

**Test plan**
pass


docs/Configuration.md
```diff
@@ -104,4 +104,4 @@ These options are only useful with React Native projects.
...
 | `postProcessBundleSourcemap`    | `PostProcessBundleSourcemap`                                           | An optional function that can modify the code and source map of the bundle before it is written. Applied once for the entire bundle.        |
 | `getModulesRunBeforeMainModule` | `(entryFilePath: string) => Array<string>`                             | An array of modules to be required before the entry point. It should contain the absolute path of each module.                              |
-| `postProcessModules`            | `(modules: Array<Module>, entryFiles: Array<string>) => Array<Module>` | A function that can change the resulting modules output.                                                                                    |
+| `processModuleFilter`            | `(module: Array<Module>) => boolean` | A function that can change the resulting modules output.                                                                                    |
```
rafeca: Can you change the description to: `A filter function to discard specific modules from the output`?

packages/metro/src/DeltaBundler/Serializers/plainJSBundle.js
```diff
@@ -42,6 +43,7 @@ function plainJSBundle(
     ...getAppendScripts(entryPoint, graph, options),
   ]
     .filter(isJsModule)
+    .filter(options.processModuleFilter)
```
rafeca: This needs to be done to [each of the serializers](https://github.com/facebook/metro/tree/master/packages/metro/src/DeltaBundler/Serializers) to have consistent results no matter the type of output that the user is generating (it's specially important to update the sourcemap serializers to not end up with incorrect sourcemaps).

cliberal:
Thanks for review
1. I did not change `getAllFiles.js` since it does't seem to be used.
2. There some redundant code but I'm not sure whether can I remove it.

```js
// sourceMapObject.js

function fullSourceMapObject(
  pre: $ReadOnlyArray<Module<>>,
  graph: Graph<>,
  options: {|
    +excludeSource: boolean,
    +processModuleFilter: (module: Module<>) => boolean,
  |},
): BabelSourceMap {
  const modules = [...pre, ...graph.dependencies.values()]
    .filter(isJsModule)
    .filter(options.processModuleFilter) // this line can be removed if we don't use it in `_sourceMapForURL` function below
    .map(module => {
      return {
        ...getJsOutput(module).data,
        path: module.path,
        source: options.excludeSource ? '' : module.getSource(),
      };
    });

  return fromRawMappings(modules).toMap(undefined, {
    excludeSource: options.excludeSource,
  });
}

// Server.js
async _sourceMapForURL(reqUrl: string): Promise<MetroSourceMap> {
    const options: DeltaOptions = this._getOptionsFromUrl(reqUrl);

    const {graph, prepend} = await this._getGraphInfo(options, {
      rebuild: false,
    });

    return sourceMapObject(prepend, graph, {
      excludeSource: options.excludeSource,
      processModuleFilter: this._config.serializer.processModuleFilter,
    });
 }
```

rafeca:
> 1. I did not change getAllFiles.js since it does't seem to be used.

`getAllFiles()` sadly is being used by [`Server.getOrderedDependencyPaths()`](https://github.com/facebook/metro/blob/master/packages/metro/src/Server.js#L238), which sadly is being used by the `react-native` cli (there's [a cli command](https://github.com/facebook/react-native/blob/master/local-cli/dependencies/dependencies.js#L52) to get all the files in the bundle).

> 2. There some redundant code but I'm not sure whether can I remove it.

The `sourceMapObject()` method is being used by the Server [symbolication entrypoint](https://github.com/facebook/metro/blob/master/packages/metro/src/Server.js#L857). This is an http entrypoint `http://localhost:8081/symbolicate` that allows simulators to display the original file/line number of errors/warnings (you may have noticed that RN development error boxes initially display the raw line number and after a small delay it shows the original file/line/column: this is because the app is making a request to the symbolication entrypoint).

PR #216, opened 2018-08-08
Upgrade Babel to 7.0.0-beta.56

LinusU:
**Summary**

This pull request bumps Babel to 7.0.0-beta.56 to incorporate the latest bugfixes from Babel.

**Test plan**

One way to test is to add a simple TypeScript file with a type as the default export:

```typescript
export const test = 1
export default interface A {}
```

and then import that file from any other file:

```js
import { test } from './typescript-test'
```

Before this would throw an error, but with the latest Babel this is now fixed.


rafeca:
Cool! Thanks for the contribution!

Can you add the `import { test } from './typescript-test'` statement to the `TypeScript.ts` test file that you created in your previous PR? I've checked and this makes the test fail in current master 👍 

Also, I'm going to let @pvdz take a look at this one, since he's the Babel guy in Metro 😄

LinusU:
> Can you add the ...

Good thinking 👍 

I added a default interface export to the file, and verified that it indeed doesn't work on current master.

rafeca:
Thanks! I've just talked to @pvdz and upgrading babel for us internally can potentially be complex (depending on whether there are some subtle breaking changes between beta54 and beta56 that forces us to update some of our internal babel plugins...

I'm going to import this PR and check if it causes any internal issues and report back

LinusU:
@rafeca any news? ☺️ 

If there were problems, anything I can do to help? I think that the largest changes are around the runtime helper loader. It's important that all usages of that module use at least beta.56 as well, as can be seen in this commit: 2fb46bd

rafeca:
We’ve found some changes in the generated sourcemaps (caused by https://github.com/babel/babel/pull/8380). The changes seem to be good but we’ve needed some time to verify things. I expect to be able to merge this soon 😃

rafeca:
giphy

dulmandakh:
@LinusU today they released v7.0.0-rc.1 😄

PR #219, opened 2018-08-09
Fix postProcessBundleSourcemap option.

gdborton:
**Summary**
This adds support back in for the `postProcessBundleSourcemap` option, which was dropped here: https://github.com/facebook/metro/commit/cd67cd47941a942509a360bac3d8a11c5ce070ff#diff-660c08c7ccb569e12d8a268bd8fa2011?w=1

I had to drop the outFilename from the params sent to `postProcessBundleSourcemap`, as not all the code paths leading to a `jsPlainBundle` or `sourceMapString` were writing to a file on disk (`createConnectMiddleware` for example).

I need this for targeting different environments (node vs web), as I need to wrap the output in different way.

**Test plan**
I updated the tests for plainJSBundle and sourceMapString, and embedded them into codeAndMap. Then tested that the option is getting called, and affecting the output as expected.


cpojer:
cc @mjesun

vitorveiga:
Hi,

Can anyone tell me if this issue is still happening in latest versions?

sergio-domingues:
Any updates on this? It would be really helpful to have access to this option, as it is documented on metro API.

TheHighriser:
Why did you remove the PR? I think this is still not working isnt it?

gdborton:
This PR is more than a year old at this point, and I'm unlikely to return to it. You can reopen with your own branch if you want.

PR #223, opened 2018-08-17
Pass `resetCache` option to `jest-haste-map`

rozele:
When the `resetCache` option is set on the bundler, it should also be passed to the `DependencyGraph` and `jest-haste-map` to clean up any cached files there.

**Summary**

While trying to get `react-native-windows` working against the `react-native@0.57` RC, I ran into a series of issues because the `jest-haste-map` cache was not cleaned up when the `--reset-cache` flag was set on the `react-native` CLI. Not only was the option not being passed to `metro` (see https://github.com/facebook/react-native/pull/20706), but `metro` was not passing the `resetCache` option to the `DependencyGraph` and `jest-haste-map`.

**Test plan**

1. yarn test
2. Validated in `react-native-windows`, where the problem was occurring. I had a case where if my `hasteImpl` implementation from `metro.config.js` had a bug. Even after fixing the bug, the `jest-haste-map` was cached with missing haste modules. Only after deleting that file could I get back to a state where the fixed `metro.config.js` would work. After making this change, using `--reset-cache` from the `react-native` CLI could clear the haste map cache.


rafeca:
Thanks! I'm landing this 👍

PR #224, opened 2018-08-18
Fix failure to resolve module if it looks like an asset (e.g. core-js $.html.js)

haggholm:
**Summary**

Fix https://github.com/facebook/react-native/issues/20721

Currently, the way the bundler checks if something is an asset or a source module is purely based on filename. This breaks in edge cases as found in `core-js` where there’s a file called `$.html.js` which is imported as `'./$.html'`. Based purely on filename, it looks like an asset; in fact, it’s a JS file, but Metro fails to find it.

**Test plan**

I’ll be honest here, I do not really understand how the Metro tests are written. I’ve verified it ‘manually’ by checking that Metro makes it past the `core-js` error, but I would be grateful for help with a test.


designtesbrot:
bump

cpojer:
cc @mjesun

cpojer:
Could you rebase this?

cpojer:
Closing due to inactivity.

Issue #225, opened 2018-08-20
Getting "Unknown named module: " errors in UMD module dependencies


adamchel:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
I’m working on creating an NPM package (“mongodb-stitch-react-native-sdk”) that should be usable by React Native applications, and I’m running into an issue where the bundle produced by the metro packager has an immediate runtime error when requiring this module.

Once I start the metro bundler (I’ve been running `react-native run-android` in my project), and I run the app with my Android device, the bundling phase seems to complete successfully, but I get an error that says `Unknown named module: ‘mongodb-stitch-react-native-core’`, that seems to be coming from `metroRequire`.

![image](https://user-images.githubusercontent.com/1976694/44362092-f6a2ab80-a48d-11e8-92f5-c3933ef3e67a.png)

This “mongodb-stitch-react-native-sdk” module is simply a UMD module (compiled by a TypeScript compiler) that imports other UMD modules that exist in `node_modules` as dependencies. The index.js of the module looks like this:

```
(function (factory) {
    if (typeof module === "object" && typeof module.exports === "object") {
        var v = factory(require, exports);
        if (v !== undefined) module.exports = v;
    }
    else if (typeof define === "function" && define.amd) {
        define(["require", "exports", "mongodb-stitch-react-native-core", "mongodb-stitch-react-native-services-mongodb-remote"], factory);
    }
})(function (require, exports) {
    "use strict";
    function __export(m) {
        for (var p in m) if (!exports.hasOwnProperty(p)) exports[p] = m[p];
    }
    Object.defineProperty(exports, "__esModule", { value: true });
    __export(require("mongodb-stitch-react-native-core")); // THIS IS THE LINE THAT IS FAILING
    __export(require("mongodb-stitch-react-native-services-mongodb-remote"));
});
//# sourceMappingURL=index.js.map
```

I’ve pointed out the line that causes the failure in the screenshot. The error seems to be occurring because the `mongodb-stitch-react-native-core` dependency doesn’t exist in the `verboseNamesToModuleIds` object map that exists in Metro’s `require` polyfill. 

I’ve tried to include the “mongodb-stitch-react-native-core” and the "mongodb-stitch-core-sdk" dependencies directly instead of using this module, but then I get a similar error because its index.js files have lines of code that look like this:

```
    var AuthInfo_1 = __importDefault(require("./auth/internal/AuthInfo"));
[...]

PR #226, opened 2018-08-21
Process locally scoped requires when they are likely part of a UMD module

adamchel:
**Summary**

Per the discussion in https://github.com/facebook/metro/issues/225, this would fix the dependency collection worker script so that it would properly interpret require calls in UMD modules. Even if a `require` is bound to a local variable, if it exists with an `exports` local variable binding in the same scope, it is likely to be a normal `require` call in a UMD module.

**Test plan**

I've updated the test for collecting dependencies to add a require call in a scope similar to one that would exist in a UMD module. I ran `lerna bootstrap` and `npm test` and everything passes for me locally. I also tested a similar change in the environment in which I reproduced the bug discussed in the  linked issue (this was an older version of Metro), and it allowed the UMD module to be properly bundled and executed without any other issues.


cpojer:
Closing as per the original issue, this change is no longer necessary for you.

PR #229, opened 2018-08-24
Bump react-deep-force-update

gaearon:
I just cut a patch for compat with some future changes in React. It's backwards compatible.

Haven't tested, but would appreciate if you could verify this didn't break hot reloading.


gaearon:
I'll need to bump this to 1.1.3 because there was a bug in 1.1.2.

PR #232, opened 2018-08-24
feat: add `resolver.moduleSystem` config property

aleclarson:
Using a locally developed Metro triggers a complaint about the `require` polyfill not existing in the project root.

To avoid this, you must copy the `require` polyfill into your project root and set the `resolver.moduleSystem` property (in your `metro.config.js` module) to an absolute path.


cpojer:
Thanks for this PR. I don't think this is actually necessary – instead what you can do is add the path to require.js to the roots of your config. This should probably be automatically done, though.

Issue #230, opened 2018-08-24
Switch minifier from uglify-es to terser


haggholm:
**Do you want to request a *feature* or report a *bug*?**

Feature

**What is the current behavior?**

Currently, `metro-minify-uglify` uses `uglify-es`, a branch of `uglify-js`. Unfortunately, `uglify-es` is not particularly well maintained and contains various bugs. Cf.

* https://github.com/mishoo/UglifyJS2/issues/3205
* https://github.com/mishoo/UglifyJS2/issues/3202
* https://github.com/mishoo/UglifyJS2/issues/3196
* https://github.com/mishoo/UglifyJS2/issues/3191

and others.

**What is the expected behavior?**

`metro-minify-uglify` could switch to `terser`. Its API is identical to `uglify-es` (except maybe some option index name? that might just be the Webpack plugin, though), so the effort should be negligible.

rafeca:
Hi! Thanks for posting the issue! The minifier is configurable in Metro so you can use anything different already.

If you want, feel free to create a PR to add a small compatibility package called metro-minify-terser, to make it super easy to switch from one to the other. I’ll be happy to merge that 😃

After that we can consider making it the default (we’ll need to do some tests internally to check that it does not regress the bundle size or build times)

rafeca:
Closing this issue, since your commit has landed and will be available in next Metro version

PR #243, opened 2018-09-12
Uglify terser

haggholm:
**Summary**

Add support for using [terser](https://github.com/fabiosantoscode/terser/). `terser` is a fork of `uglify-es`, which is not really maintained; and maintains API compatibility with `uglify-es`.

In order to avoid duplicating configuration like default options in `metro-minify-uglify`, this PR adds a parameter to `metro-minify-uglify` that allows passing in an alternative minifier module, and the new package `metro-minify-terser` calls this, passing in `terser`.

This is based on discussion in https://github.com/facebook/metro/issues/230.

**Test plan**

The tests are simply copied from `metro-minify-uglify`, since the API is and the results should be identical.

### DIFF
--- a/packages/metro-minify-terser/package.json
+++ b/packages/metro-minify-terser/package.json
@@ -13,6 +13,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "metro-minify-uglify": "^0.45.2",
     "terser": "^3.8.2"
   },
   "devDependencies": {}
--- a/packages/metro-minify-terser/package.json
+++ b/packages/metro-minify-terser/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "metro-minify-terser",
+  "version": "0.45.2",
+  "description": "🚇 Terser-based minifier alternative for Metro",
+  "main": "src/index.js",
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:facebook/metro.git"
+  },
+  "scripts": {
+    "prepare-release": "test -d build && rm -rf src.real && mv src src.real && mv build src",
+    "cleanup-release": "test ! -e build && mv src build && mv src.real src"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "terser": "^3.8.2"
+  },
+  "devDependencies": {}
+}
--- a/packages/metro-minify-terser/src/types.js.flow
+++ b/packages/metro-minify-terser/src/types.js.flow
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ * @format
+ */
+
+'use strict';
+
+import type {BabelSourceMap} from '@babel/core';
+import type {MinifyOptions} from 'metro/src/JSTransformer/worker';
+
+export type {MinifyOptions};
+
+export type ResultWithMap = {
+  code: string,
+  map: BabelSourceMap,
+  options?: MinifyOptions,
+};
+
+export type ResultWithoutMap = string;
+
+type MinifierWithSourceMap = (
+  code: string,
+  inputMap?: ?BabelSourceMap,
+  filename: string,
+  options?: MinifyOptions,
+) => ResultWithMap;
+
+type MinifierWithoutSourceMap = (
+  code: string,
+  options?: MinifyOptions,
+) => ResultWithoutMap;
+
+export type MetroMinifier = {
+  noSourceMap: MinifierWithoutSourceMap,
+  withSourceMap: MinifierWithSourceMap,
+};
--- a/packages/metro-minify-uglify/src/minifier.js
+++ b/packages/metro-minify-uglify/src/minifier.js
@@ -23,8 +23,9 @@ import type {BabelSourceMap} from '@babel/core';
 function noSourceMap(
   code: string,
   options?: MinifyOptions = {},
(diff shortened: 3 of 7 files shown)


packages/metro-minify-terser/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "type": "git",
+    "url": "git@github.com:facebook/metro.git"
+  },
+  "scripts": {
+    "prepare-release": "test -d build && rm -rf src.real && mv src src.real && mv build src",
+    "cleanup-release": "test ! -e build && mv src build && mv src.real src"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "metro-minify-uglify": "^0.45.2",
```
Salakar: Version needs to be hard set - e.g. remove the "^" - node CI tests should be good after this. 👍

packages/metro-minify-terser/src/types.js.flow
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export type ResultWithoutMap = string;
+
+type MinifierWithSourceMap = (
+  code: string,
+  inputMap?: ?BabelSourceMap,
+  filename: string,
+  options?: MinifyOptions,
+) => ResultWithMap;
```
Salakar: Missing `terser` param type definition as 5th arg.

packages/metro-minify-terser/src/types.js.flow
```diff
@@ -0,0 +1,41 @@
...
+
+type MinifierWithoutSourceMap = (
+  code: string,
+  options?: MinifyOptions,
+) => ResultWithoutMap;
```
Salakar: Missing `terser` param type definition as 3rd arg.

packages/metro-minify-uglify/src/minifier.js
```diff
@@ -23,17 +23,19 @@ import type {BabelSourceMap} from '@babel/core';
 function noSourceMap(
   code: string,
   options?: MinifyOptions = {},
+  uglifyModule: ?Object,
 ): ResultWithoutMap {
```
Salakar: Types need updating [here](https://github.com/facebook/metro/blob/master/packages/metro-minify-uglify/src/types.js.flow#L36) to add the extra `uglifyModule` arg.

packages/metro-minify-uglify/src/minifier.js
```diff
@@ -23,17 +23,19 @@ import type {BabelSourceMap} from '@babel/core';
...
-  return minify(code, undefined, options).code;
+  return minify(code, undefined, options, uglifyModule).code;
 }
 
 function withSourceMap(
   code: string,
   sourceMap: ?BabelSourceMap,
   filename: string,
   options?: MinifyOptions = {},
+  uglifyModule: ?Object,
```
Salakar: Types need updating [here](https://github.com/facebook/metro/blob/master/packages/metro-minify-uglify/src/types.js.flow#L31) to add the extra `uglifyModule` arg.

haggholm:
The reason I made it a parameter is that the alternative is to literally duplicate the code of `metro-minify-uglify`—certainly I could easily do this: it would literally involve just changing `/uglify(-es)?/` with `terser`. but it seems to me like an unnecessary maintenance burden, as any changes to preferred options would have to be copied between the `terser` and `uglify` versions. The parameter to `-uglify` is not exposed, but just an internal thing to allow for shared implementation. Keep in mind that `terser` is API compatible with `uglify-es` and `uglify-js@3`, so the calling code in Metro will be identical.

I suppose an arguably cleaner way would be to have some sort of `metro-minify-base` (never directly referenced by users) which is exposed via `-uglify` and `-terser` versions.

I don’t really care too much ;) – after all I’m not a maintainer! I just wanted to keep it DRY. But I’ll happily do either one: (a) the way I first did it, (b) making `-terser` a straight copy of `-uglify` (with the underlying module substituted), or (c) creating an ‘internal’ base module that is exposed via specialised versions.

A fourth alternative (d) might be to add a different option: keep *just* `metro-minify-uglify` but allow an override of the `uglify` implementation used. Though this might just look like unnecessary proliferation of options.

But either way, let me know which way you’d like it to go and I’ll gladly contribute it. (Or, obviously, if you’d rather do it, that’s fine—I just want to use `terser`!)

rafeca:
Hi @haggholm ! Thanks for the contribution! We're indeed exploring using `terser` ourselves, so this PR is really appreciated 😃

As @CompuIves mentioned, it would be better to make `metro-minify-terser` a first-class citizen and not make it go through `metro-minify-uglify`.

I agree that this means some duplication currently, but we are also going to simplify the API of the minifier so it'll only need to expsoe a single `minify` method (since the `noSourceMap` method is not used anywhere by Metro).

Also, we're going to have to move the flow types of the minifier API to the Metro repo, so they can be reused by any other potential minify implementation.

For now, I'm fine if you just copy the current `metro-minify-uglify` to create the `metro-minify-terser` package, and we'll do the cleaning later.

If you also want to do the cleaning, this is the two things that I would do (I'd appreciate a lot 😍):

1- Simplify the minifier API by getting rid of the `noSourceMap` method and renaming `withSourceMap` to `minify`.
2- Move the [flow types](https://github.com/facebook/metro/blob/master/packages/metro-minify-uglify/src/types.js.flow) of the minifier to the `metro` repository, so any package can import them easily.

Thanks again!

haggholm:
Since the changes from the current `master` branch are really quite different, I made the changes in a new PR to supercede this one: https://github.com/facebook/metro/pull/256

rafeca:
Closing this PR then

PR #246, opened 2018-09-13
[visualizer] make metro visualizer an optional dependency

Salakar:
**Summary**

This makes `metro-visualizer` an optional package of `metro` as currently it's quite heavy on dependencies and takes up a majority percentage of React Native's package install size; 

image

And: https://twitter.com/jamonholmgren/status/1040260845857656833?s=21

This will try require the package if the config option is enabled and warn if the visualizer package could not be found, but, still gracefully continues. This means its now up to user-land code whether to include this package in their project.

**Test plan**

All tests pass locally:

image

### DIFF
--- a/packages/metro/package.json
+++ b/packages/metro/package.json
@@ -77,7 +77,6 @@
     "metro-react-native-babel-preset": "0.45.2",
     "metro-resolver": "0.45.2",
     "metro-source-map": "0.45.2",
-    "metro-visualizer": "0.45.2",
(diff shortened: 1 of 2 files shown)


Salakar:
@rafeca If you'd like I can send up another PR after this one is merged to document the `enableVisualizer` option and also add a note about package requirement? 🙃 The option looks to be undocumented currently

rafeca:
Oh yes please, that would be awesome 😃

PR #250, opened 2018-09-15
fix: ensure "json" exists in `resolver.sourceExts`

aleclarson:
**Summary**

The resolver breaks when `json` is accidentally omitted from the `resolver.sourceExts` array, because Haste FS fails to recognize `package.json` files, which is crucial for module resolution.

**Test plan**

None yet.


rafeca:
Thanks a lot for the PR! can you rebase it? the tests should get fixed afterwards 😄

aleclarson:
Done 👍

PR #251, opened 2018-09-17
[metro] and support in reactNativeTransformer for .js babel config files

Salakar:
**Summary**

- **[feature]** Adds support for `.js` file babel configs, specifically:
  - `.babelrc.js`
  - `babel.config.js`
- **[typo]** Fix a typo in the error code checking `error.conde` -> `error.code`
- **[housekeeping]** default case now no longer reliant on the `rn-babelrc.json` file, moved this in-line due to it only needing a singular preset. This additionally removed the need for the `json5` dependency and has now been removed from the package.json.

As below, original case has been tested as working (`.babelrc`) - no breaking changes.

**Test plan**

Linting, flow checks and tests passing locally:
image

Additionally tested all 3 file types in a local project, with logging to indicate that the config had actually been applied (made each custom babel config require a preset that does not exist - making it error and log to indicate that it loaded the custom config)

**.babelrc**
image

**.babelrc.js**
image

**babel.config.js**
image

**without custom config**
builds bundle correctly:
image

 

Thanks 🙃

### DIFF
--- a/packages/metro/src/reactNativeTransformer.js
+++ b/packages/metro/src/reactNativeTransformer.js
@@ -47,61 +47,72 @@ const getBabelRC = (function() {
 
     babelRC = {plugins: []};
 
-    // Let's look for the .babelrc in the project root.
+    // Let's look for a babel config file in the project root.
     // In the future let's look into adding a command line option to specify
     // this location.
     let projectBabelRCPath;
+
+    // .babelrc
     if (projectRoot) {
       projectBabelRCPath = path.resolve(projectRoot, '.babelrc');
     }
 
-    // If a .babelrc file doesn't exist in the project,
-    // use the Babel config provided with react-native.
-    if (!projectBabelRCPath || !fs.existsSync(projectBabelRCPath)) {
-      babelRC = json5.parse(
-        fs.readFileSync(require.resolve('metro/rn-babelrc.json')),
-      );
+    // .babelrc.js
+    if (projectBabelRCPath && !fs.existsSync(projectBabelRCPath)) {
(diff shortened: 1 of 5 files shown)


packages/metro/src/reactNativeTransformer.js
```diff
@@ -47,61 +47,72 @@ const getBabelRC = (function() {
...
+      projectBabelRCPath = path.resolve(projectRoot, 'babel.config.js');
+    }
+
+    // We found a babel config file, extend our config off of it
+    if (projectBabelRCPath && fs.existsSync(projectBabelRCPath)) {
+      babelRC.extends = projectBabelRCPath;
+      return babelRC;
+    }
+
+    // Babel config file doesn't exist in the project,
```
rafeca: I wonder if we can get rid of all the following logic and just do `babelRC.extends = require.resolve('metro/rn-babelrc.json');`... would that work? (we may need to tweak the preset path in `metro/rn-babelrc.json`).

rafeca:
Wow that's nice, and much cleaner! 😄 Thanks a lot!

Check my comment, maybe you can clean the logic up even more 🎉

packages/metro/src/reactNativeTransformer.js
Salakar: I thought that as well but I wasn't sure if I should change it too much just in case - as you've mentioned it though I'll give it a go in the morning and push up👍

Salakar:
@rafeca updated as discussed here / on slack 👍 

I've also re-run all tests for the specific config file variants in my local project - have updated the `Test plan` section of the issue to reflect local testing output.

Thanks

PR #252, opened 2018-09-17
Bump emotion dependency

LinusU:
**Summary**

This newer version of emotion depends on the stable release of Babel 7, and avoids pulling in beta versions of Babel.

see emotion-js/emotion#852

**Test plan**

Ran `yarn test` locally, this is only a patch bump and shouldn't break anything.


LinusU:
Awesome 👏 

I'm slowly working my way thru all the dependencies ☺️

PR #267, opened 2018-09-26
Updates to UnableToResolveError to account for different platforms.

gdborton:
**Summary**
This error message is confusing when you get it on a mac, as /tmp/ isn't the
tmp directory that is being used.  The message should probably try to account
for settings that are configured by users as well, the Metro Bundler cache at
least is something that can be overwritten.

**Test plan**

Edited my node_modules/ folder for the metro-sample-app.

### DIFF
--- a/packages/metro/src/node-haste/DependencyGraph/ModuleResolution.js
+++ b/packages/metro/src/node-haste/DependencyGraph/ModuleResolution.js
@@ -13,6 +13,7 @@
 const Resolver = require('metro-resolver');
 
 const invariant = require('fbjs/lib/invariant');
+const os = require('os');
 const path = require('path');
 const util = require('util');
 
@@ -175,6 +176,14 @@ class ModuleResolver<TModule: Moduleish, TPackage: Packageish> {
           .concat(extraPaths);
 
         const hint = displayDirPaths.length ? ' or in these directories:' : '';
+        const metroCacheLocation = path.join(
+          os.tmpdir(),
+          'metro-bundler-cache-*',
+        );
+        const hasteCacheLocation = path.join(
+          os.tmpdir(),
+          'haste-map-metro-4-*',
+        );
         throw new UnableToResolveError(
           fromModule.path,
           moduleName,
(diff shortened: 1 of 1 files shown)


packages/metro/src/node-haste/DependencyGraph/ModuleResolution.js
```diff
@@ -175,6 +176,14 @@ class ModuleResolver<TModule: Moduleish, TPackage: Packageish> {
           .concat(extraPaths);
 
         const hint = displayDirPaths.length ? ' or in these directories:' : '';
+        const metroCacheLocation = path.join(
+          os.tmpdir(),
+          'metro-bundler-cache-*',
+        );
+        const hasteCacheLocation = path.join(
+          os.tmpdir(),
+          'haste-map-metro-4-*',
```
gdborton: This changed from `haste-map-react-native-packager-*`, but ideally should come from the location where it's configured: https://github.com/facebook/metro/blob/73dd5e8f1b203313e46974651d723ef84d59e5b2/packages/metro/src/node-haste/DependencyGraph.js#L91

rafeca: We could get the actual cache path from `jest-haste-map` if it was exposed as a getter ([from here](https://github.com/facebook/jest/blob/master/packages/jest-haste-map/src/index.js#L213)).

For now, this is perfectly fine, but I would change the second part of `hasteCacheLocation` to `haste-map-metro-*` so we don't have to update the error message if we bump the `JEST_HASTE_MAP_CACHE_BREAKER` variable.

gdborton: Good call on removing the version, I'll update.

We don't have access to the HasteMap instance right?  So we'd need to add it to module_map.js?

rafeca: Yeah, we could though change a bit the code here to have the HasteMap instance here once the getter for the cache folder gets eventually added to `jest-haste-map`.

rafeca: jest v24 is going to have [a getter for the cache path](https://github.com/facebook/jest/pull/7217) 😃 

The first alpha for v24 has just been published and we're going to update Metro to this alpha version soon

cpojer:
Closing because of inactivity.

PR #268, opened 2018-09-26
Resolves transformerPath before attempting to read contents.

gdborton:
**Summary**
Configuration paths are now supposed to be relative to allow them to be shared
on different paths/machines. This change makes sure that the new relative
transformerPath is require.resolved before being passed to fs.readFileSync.

**Test plan**
I edited node_modules directly in metro-sample-app to ensure that the fix got
me passed the unable to resolve error.


rafeca:
Yay! thanks! 😃

PR #278, opened 2018-10-02
[metro] add $FlowExpectedError annotation for optional visualizer dep

Salakar:
This adds a `$FlowExpectedError` comment annotation for the optional `metro-visualizer` dependency, resolves the following issue:

image

**Summary**

**Test plan**

image


Salakar:
It is strange, there was a dev dependency but removing that locally made no difference, I had to physically delete the metro-visuallizer directory in `packages/` locally to be able to reproduce the error - so must be a yarn workspaces thing, one to note for future I guess 🙈

PR #279, opened 2018-10-04
[build] Adds prettier to the build step to make output more readable.

gdborton:
**Summary**

This makes debugging errors that occur in builds easier as the build output is much more human readable.

**Test plan**
Ran a build and checked the output.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -29,6 +29,7 @@ const glob = require('glob');
 const micromatch = require('micromatch');
 const mkdirp = require('mkdirp');
 const path = require('path');
+const prettier = require('prettier');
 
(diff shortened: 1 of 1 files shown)


rafeca:
Ohhh this makes sense! 😄

Since we're not generating sourcemaps for the npm publsihed packages it should be fine.

Can you verify that prettier is not adding things like trailing commas in function calls? Although Metro officially requires minimum Node.js v8, I'd prefer to not break earlier versions due to the trailing commas.

gdborton:
@rafeca Checked for trailing commas, and there don't appear to be any. I'm not sure what other syntax would break node 6, but I don't expect anything breaking to be added by prettier here.

PR #280, opened 2018-10-04
[assets] Adds support for publicPath to enable serving assets from different locations.

gdborton:
Opening this PR to get some feedback/direction on the approach. Our use case requires that we serve our js/maps/image/css assets from `/static/`, which breaks some of the assumptions in the metro bundler.

This is a start toward allowing us to configure the path that compiled assets are served from. I think that publicPath makes the most sense as a server option, but it'll be needed by transformers and serializers to enable support for assets and source maps.

### DIFF
--- a/packages/metro/src/Assets.js
+++ b/packages/metro/src/Assets.js
@@ -177,8 +177,9 @@ async function getAssetData(
   localPath: string,
   assetDataPlugins: $ReadOnlyArray<string>,
   platform: ?string = null,
+  publicPath: ?string = '/assets',
 ): Promise<AssetData> {
-  let assetUrlPath = path.join('/assets', path.dirname(localPath));
+  let assetUrlPath = `${publicPath}/${path.dirname(localPath)}`;
 
   // On Windows, change backslashes to slashes to get proper URL path from file path.
   if (path.sep === '\\') {
--- a/packages/metro-config/src/defaults/index.js
+++ b/packages/metro-config/src/defaults/index.js
@@ -79,6 +79,7 @@ const getDefaultValues = (projectRoot: ?string): ConfigT => ({
     postMinifyProcess: x => x,
     transformVariants: {default: {}},
     workerPath: 'metro/src/DeltaBundler/Worker',
+    publicPath: '/assets',
   },
   cacheStores: [
     new FileStore({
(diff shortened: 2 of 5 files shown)


gdborton:
Some assumptions that will have to change to enable this:

Server assumes `/assets/`
https://github.com/facebook/metro/blob/master/packages/metro/src/Server.js#L479

https://github.com/facebook/metro/blob/master/packages/metro/src/Server.js#L546

gdborton:
cc: @raejin

rafeca:
The direction looks good to me. 

Regarding where to put the parameter, in order to make it available in the transformer you need to have it in the `transformer` object (I'm working on a change to make Metro pass any configured transformer option blindly).

Regarding the assets serialization, in the future the transformer should return both the JS code for the asset and the `AssetData` object as two different outputs, this way the serializer won't need to call `getAssetData()` again. This is a small refactor that needs to be done and will allow us to use the path in one less place.

packages/metro/src/Assets.js
```diff
@@ -177,8 +177,9 @@ async function getAssetData(
   localPath: string,
   assetDataPlugins: $ReadOnlyArray<string>,
   platform: ?string = null,
+  publicPath: ?string = '/assets',
 ): Promise<AssetData> {
-  let assetUrlPath = path.join('/assets', path.dirname(localPath));
+  let assetUrlPath = `${publicPath}/${path.dirname(localPath)}`;
```
rafeca: I think this is causing the end to end snapshot tests to fail

gdborton: Yeah I haven't addressed any tests yet, I need to get around the Server assumptions, but I'm not sure how to handle that.

cpojer:
@gdborton would you mind rebasing this PR and making the tests pass?

cpojer:
Closing because of inactivity.

relign:
I also need publicPath  to enable serving assets from different locations. why not merge ?

relign:
I also need publicPath  to enable serving assets from different locations. why not merge ?

raejin:
@relign I merged a PR that serves a similar purpose: https://github.com/facebook/metro/pull/299

relign:
> @relign I merged a PR that serves a similar purpose: #299

Thanks.

PR #307, opened 2018-11-02
do not inject es6 `default` syntax into dynamically imported modules

benjroy:
I believe there is a bug in the `asyncRequire` module.  It should resolve the promise with exactly what was returned from the `require(moduleID)` call.  It should not nest the module under another `default` key.

`babel-plugin-dynamic-import-node` rewrites `import(...)` statements in this way.  Here is some example output from some test fixtures: https://github.com/airbnb/babel-plugin-dynamic-import-node/blob/master/test/fixtures/dynamic-argument/expected.es2015.js

There is a snippet here https://developers.google.com/web/updates/2017/11/dynamic-import that shows explicit usage of dynamic import with an `es6` module.  The spe
```js
  (async () => {
    const moduleSpecifier = './utils.mjs';
    const module = await import(moduleSpecifier)
    module.default();
    // → logs 'Hi from the default export!'
    module.doStuff();
    // → logs 'Doing stuff…'
  })();
```

I have a line of code in a project i am working on.
```js
import('./migrations/accordion/accordion.2.0.0').then((module) => module.default)
```
the module at `'./migrations/accordion/accordion.2.0.0'` looks something like:
```js
export default function foo() { /* ... */ }
```

Without the patch in this PR, the import(...) call returns something like:
```js
{
  default: {
    default: [Function foo]
  }
}
```

With this patch, the import(...) call will return properly:
```js
{
  default: [Function foo]
}
```

### DIFF
--- a/packages/metro/src/lib/bundle-modules/asyncRequire.js
+++ b/packages/metro/src/lib/bundle-modules/asyncRequire.js
@@ -13,5 +13,5 @@
 // eslint-disable-next-line lint/flow-no-fixme
 const dynamicRequire = (require: $FlowFixMe);
 module.exports = function(moduleID: mixed): Promise<mixed> {
-  return Promise.resolve().then(() => ({default: dynamicRequire(moduleID)}));
+  return Promise.resolve().then(() => dynamicRequire(moduleID));
 };


rafeca:
Thanks a lot for the contribution!

You're right: the current behaviour is not correct. For some context, it was implemented to support importing commonJS modules (this is why we added the `default` key there), but indeed this does not work when importing ES6 modules.

In order to fix the issue, though, you'll need to add some small interop layer (the same way that [babel does](https://github.com/airbnb/babel-plugin-dynamic-import-node/blob/master/test/fixtures/dynamic-argument/expected.es2015.js)), this way you can fix importing ES6 modules while keeping the support to commonJS modules.

Basically, the logic that needs to be added is something similar to the one [we have in the require implementation](https://github.com/facebook/metro/blob/afee5b1a1021cf4ad6c43405a28325d26a0e4131/packages/metro/src/lib/polyfills/require.js#L225-L239).

In order to verify that the logic works well, I suggest adding an end to end test that verifies that a dynamic import to either a commonJS or an ES6 module works fine. You can add this to our current `import`/`export` test bundle: https://github.com/facebook/metro/blob/master/packages/metro/src/integration_tests/basic_bundle/import-export/index.js

mjesun:
I'll take care of that one, as it requires putting in place some interoperability functions to deal with ES5 :)

benjroy:
Looks like y'all took care of it with https://github.com/facebook/metro/commit/7b89dcd3524c0ec45c5f4cd722b32d79da7b336b

Thank you!

PR #309, opened 2018-11-08
Add custom serializer

raejin:
**Summary**
We would like to bundle an entryfile differently than it is currently done in Metro, specifically in development. Without a pluggable serializer, we wouldn't be able to perform bundle splitting in development, which is impactful to our development speed. 

This PR alleviates bundle sizing problem by introducing a customSerializer instead of using the `plainJSBundle`.

Some minor question, do I have to implement this in the `build` public method in the server? 

**Test plan**
- Wrote a `customSerializer` and test it against the new change in metro. Our bundle was able to do what we needed it to accomplish.

### DIFF
--- a/packages/metro/src/Server.js
+++ b/packages/metro/src/Server.js
@@ -664,6 +664,7 @@ class Server {
           }));
 
       const options = {
+        customSerializer: this._config.serializer.customSerializer,
         processModuleFilter: this._config.serializer.processModuleFilter,
         createModuleId: this._createModuleId,
         getRunModuleStatement: this._config.serializer.getRunModuleStatement,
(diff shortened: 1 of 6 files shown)


packages/metro/src/Server.js
```diff
@@ -689,7 +689,14 @@ class Server {
...
             })
           : Object.assign({}, options, {embedDelta: false}),
-      );
+      ];
+      let bundle;
+      const possibleCustomSerializer = this._config.serializer.customSerializer;
+      if (possibleCustomSerializer) {
+        bundle = possibleCustomSerializer(...serializerArguments);
+      } else {
+        bundle = plainJSBundle(...serializerArguments);
```
mjesun: Making the `if` a ternary should let (😄) you use `const`.

PR #315, opened 2018-11-14
Support for specifying a custom `require.js` file to be used.

codyi96:
**Summary**
In some cases, we may need to make some changes to the default polyfill file named `require.js`. This commit supports a custom `require.js` file.

**Test plan**
1. Specify a custom `require.js` and add it into the result of config.serializer.getPolyfills().
You should define options to override default config.
``` javascript
let defaultConfigOverrides = {
        resolver: {
          resolverMainFields: ['browser', 'main'],
          blacklistRE: createBlacklist([/.*\/__fixtures__\/.*/]),
        },
        serializer: {
          getModulesRunBeforeMainModule: () => [
            require.resolve('path/of/InitializeCore.js'),
          ],
          getPolyfills: () => [
            require.resolve('path/of/Object.es6.js'),
            ...
            require.resolve('path/of/require.js'),
          ]
        },
        transformer: {
          babelTransformerPath: require.resolve('metro/src/reactNativeTransformer'),
        },
        watchFolders: [],
      };
```
Then use it
``` javascript
const {loadConfig} = require('metro-config');
let config = await loadConfig(argv, defaultConfigOverrides);
```
2. Configure server with useCustomPolyfill whose value is true
3. Build bundle
4. Your code will replace the default code in the bundle.


cpojer:
Thanks for your PR.

Can you go into more details about why this is needed? I am not questioning that you may want to patch `require.js` but I'm wondering if there are other ways to accomplish what you are trying to do. The reason I'm hesitant to merge it is because the interface that our `require.js` provides is intrinsically relied on by all of Metro, as well as we optimized our `require.js` for performance quite a lot. I do not think it's a good idea to allow to customize such an integral part of infra and risk issues further down the road but I'm still curious about your use case before closing this PR.

codyi96:
For my use case, I want to change the default `require.js` so that i can get useful log and customize some functions for my modules.
For example:
```
...
function isMyAssetModule(moduleId) {
  var regex = /@codyi96\/[^\/]+\//;
  return new RegExp(regex).test(moduleId);
}
function define(factory,moduleId,dependencyMap)
{
  if (moduleId in modules && !isMyAssetModule(moduleId)) {
    // my asset module should be defined even if it has been in modules
    return;
  }
...
```

cpojer:
I would recommend you to add this kind of stuff directly to the modules via a custom babel plugin. That way it will be local to the modules you are changing instead of possibly having a very negative performance impact on your entire app.

Please note that we are planning on large changes to RN that will affect how modules are bundled. Unfortunately I cannot talk more about this right now, but please do understand that I'm trying to make sure you won't have more pain in the future.

PR #319, opened 2018-11-28
Pass caller option to babel when transforming. Fixes #190

### LINKED ISSUE #190: [META] conditional babel transpile
In version 0.56, iOS 9 is now the minimum required version. And on iOS 9, Webkit and JSC partially supports ES6 (https://webkit.org/blog/4054/es6-in-webkit/).

So it might be better to not transpile natively supported ES6 features to have native performance, smaller bundle size. Especially for release builds.

Relevant blog posts:

https://webkit.org/blog/4054/es6-in-webkit/
https://webkit.org/blog/6756/es6-feature-complete/
https://webkit.org/blog/7536/jsc-loves-es6/

satya164:
**Summary**

Passing this option allows us to conditionally return a different babel config according to the platform. For example:

```js
// babel.config.js
module.exports = api => {
  const platform = api.caller(caller => caller.platform);

  if (platform === 'android') {
    // return some config
  }

  // return some config
};
```

This is especially useful when using the new JSC on Android where we don't need to run all of the transforms.

Fixes #190

**Test plan**

Tested in a freshly inited React native app. The `caller` is correctly set to `{ caller: 'metro', platform: 'ios' }`

### DIFF
--- a/flow-typed/babel.js.flow
+++ b/flow-typed/babel.js.flow
@@ -192,6 +192,7 @@ type BabelCoreOptions = {|
   babelrc?: boolean,
   babelrcRoots?: Array<string>,
   envName?: string,
+  caller?: {name: string},
   code?: boolean,
   comments?: boolean,
   compact?: 'auto' | boolean,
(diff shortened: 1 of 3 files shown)


satya164:
cc @rafeca

gengjiawen:
I have verified on windows.

rafeca:
Thanks! 🎉

PR #321, opened 2018-11-29
Update to gh-pages Fixes Tagline and Home Page Image & Fix for Favicon

Developerayo:
**Summary**

Added the  "🚇"  to the title and Home Page name of facebook.github.io/metro/". Before I added it, it was showing to the public a Vertical Rectangle sign. Now it's showing the correct favicon  "🚇". And also the Favicon for the website was not showing on the live website because of incorect href linking.

When @mjesun came to Lagos,Nigeria  he talked so passionately about how Facebook is building Metro to help build faster Apps, so I decided to take a look at getting started with using Metro-Bundler.

**Test plan**

Once its merged the Tagline Favicon would should on the website at facebook.github.io/metro/ as "🚇 The JavaScript bundler for React Native" and also on the Home Page Header. Also Reload "https://developerayo.github.io/metro/" to see preview of the change!

## How It looks on my Fork with my Changes

metro2

metro

## How it looks currently of your End

metro_old1

metro_old


mjesun:
Yay! Thanks for contributing! ☺️ These pages are auto-generated, though. I think you will need to change the icon [here](https://github.com/facebook/metro/blob/master/website/siteConfig.js#L14), then regenerate them (don't use the emoji directly as we only accept ASCII characters in code - escape it using its unicode sequence).

Developerayo:
> Yay! Thanks for contributing! ☺️ These pages are auto-generated, though. I think you will need to change the icon [here](https://github.com/facebook/metro/blob/master/website/siteConfig.js#L14), then regenerate them (don't use the emoji directly as we only accept ASCII characters in code - escape it using its unicode sequence).

Hello @mjesun see [Issue #323 here](https://github.com/facebook/metro/pull/323) with that fix.

mjesun:
@Developerayo the doc changes are updated automatically, no need to update the branch manually! :)

You should already see it in: https://facebook.github.io/metro/.

PR #335, opened 2018-12-18
feat: Move dependencies command from react-native-cli to metro

Esemesek:
**Summary**
This command exists as a part of [react-native-cli](https://github.com/react-native-community/react-native-cli/blob/master/packages/local-cli/dependencies/dependencies.js). We agreed with @cpojer and @grabbou, that this command is out of the scope of `cli` and should be moved to `metro` repository.
Also not sure if I should `legacy API` - it was used in `react-native-cli`.

**Test plan**
This is moved from `react-native-cli` so I tested it by running commands in terminal.
screenshot 2018-12-18 at 18 36 37


cpojer:
I was hoping we could call it get-dependencies. Would you mind renaming it?

Esemesek:
I don't mind it. https://github.com/facebook/metro/pull/339 + added docs and fixed one description.

PR #362, opened 2019-02-22
[react-native] Default to lazy requires for all public RN exports

ide:
**Summary**

This commit changes `metro-react-native-babel-preset` to emit lazy `require` calls for all of RN's public exports. This is so that we can move RN to use ES import/export, which syntactically have no lazy version, while preserving the lazy behavior we have today by default.

If the developer specifies `lazyImportExportTransform: true`, all imports/exports will be lazy including RN's public exports, so they will remain lazy. (Also, this PR allows developers to specify non-boolean values for `lazyImportExportTransform` e.g. an array of whitelisted paths.)

Note: this commit must land and be published before RN can use ESM since the iOS E2E tests depend on the current lazy `require` calls.

See the draft PR for RN here: https://github.com/facebook/react-native/pull/23591

**Test plan**

Used this module inside of the RN E2E iOS tests and loaded the bundle successfully. Also used it in RNTester.


cpojer:
cc @mjesun 

We discussed this for RN's `react-native-implementation`, does the change to support this in Metro make sense to you?

mjesun:
Yeah, this makes sense to me; although my suggestion is to lazify everything :)

Metro has anyway native support for import/export, but this was mentioned for tests, where you’ll still need the Babel transform; so happy to get that in.

ide:
Would be nice to move towards making more things lazy (ex: if package.json contains a "lazyImport" flag). In the short term, it'd be nice to have this RN-specific change for RN 0.60. On Expo's side we can patch stuff but for people not using Expo it probably makes sense to land the same performance improvements (lazy imports -> opens the door for moving to import/export -> enables dead-export elimination).

matthargett:
Right now we have a Babel transform that does this conversion so tree shaking (and smaller bundle size) is possible. It would be great to be able to remove that transform for faster bundle times with RN 0.60.

ide:
> we have to find a way to keep this list up-to-date

Broadly speaking I agree it'd be better to update the list periodically but things will generally work even if the list gets out of date so I'm not so worried. If exports are removed from RN (ex: lean core) then those entries on this list will be no-ops, and if new exports are added to RN then they won't get transformed into lazy requires. Neither is ideal but also neither is a severe build breaker.

PR #364, opened 2019-02-25
Allow custom BABEL_ENV

angly-cat:
**Summary**

Allows custom non-dev `BABEL_ENV`s besides `'production'`. With respect to https://github.com/facebook/react-native/pull/6351#issuecomment-308820107.

**Test plan**

This change will break non-dev builds for people who for any reason has custom `BABEL_ENV` set but expects it to behave like `'production`' (current behavior). Though it can be easily fixed either by renaming their `env` from `production` to the value of custom `BABEL_ENV`, or by changing `BABEL_ENV` from custom to `'production'`.

Other than this unusual case, nothing should change.


angly-cat:
Supersedes https://github.com/facebook/metro/pull/183.

@cpojer please take a look.

PR #389, opened 2019-04-09
Removed unused code to compute worker execArgv in WorkerFarm

ms:
**Summary**

The execArgv is not passed in to JestWorker anymore but was not removed
(and I assume linting doesn't catch it because we call `.shift()` on it
which could have a side effect)

I discovered this as I was trying to fix a bug with `--inspect` being passed in to child workers because the code used to overwrite the JestWorker behavior when computing child `execArgv`

**Test plan**

Tests still pass!


ms:
I believe that the test failures are unrelated to this change. Let me know if I should push a rebased commit to trigger the build again.

Issue #395, opened 2019-04-24
Enable isTSX for .ts files


OzzieOrca:
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
Babel throws errors when using JSX in `.ts` files. `Unexpected token, expected "</>/<=/>="`

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**
Use JSX in a `.ts` file. (I have [`sourceExts`](https://facebook.github.io/metro/docs/en/configuration#sourceexts) set to `['jsx', 'js', 'ts']`.)

**What is the expected behavior?**
Is there any way to override [`isTSX`](https://babeljs.io/docs/en/babel-plugin-transform-typescript#istsx) for `.ts` files?

Specifically is there any way to override it in the project's babel config? https://github.com/babel/babel/issues/8799 seems to indicate you can't and I wasn't able to add a babel config plugin entry for `@babel/plugin-transform-typescript` and get it picked up.

Or could I submit a PR for configuring both `.ts` and `.tsx` with `isTSX: true`? Or, if you still need to support typescript's legacy angle bracket for Type Assertion, is there a way to add a config option somewhere?

```diff
{
  test: isTypeScriptSource,
  plugins: [
-    [require('@babel/plugin-transform-typescript'), {isTSX: false}],
+    [require('@babel/plugin-transform-typescript'), {isTSX: true}],
  ],
},
```
https://github.com/facebook/metro/blob/master/packages/metro-react-native-babel-preset/src/configs/main.js#L175

Making the above change locally appears to work fine.

https://github.com/airbnb/javascript/pull/985#issuecomment-239145468 seems to be a great  justification for allowing JSX in `.js` files. The same argument should be applied to allowing JSX in `.ts` files.

**Please provide your exact Metro configuration and mention your Metro, node, yarn/npm version and operating system.**
Metro: 0.51.1

robhogan:
We’re triaging older issues. `metro-react-native-babel-preset` has been replaced by `@react-native/babel-preset`, so this question no longer applies to supported React Native versions.

PR #396, opened 2019-04-24
Add forceIsTSX option to support JSX in .ts files

OzzieOrca:
**Summary**
Add forceIsTSX option to support JSX in .ts files

Addresses #395

**Test plan**
Test that JSX code in `.ts` files no longer throws errors and the `isTSX` config option is enabled for `@babel/plugin-transform-typescript` when babel is configured like:
```js
module.exports = {
  presets: [['module:metro-react-native-babel-preset', { forceIsTSX: true }]],
};
```

### DIFF
--- a/packages/metro-react-native-babel-preset/src/configs/main.js
+++ b/packages/metro-react-native-babel-preset/src/configs/main.js
@@ -172,7 +172,10 @@ const getPreset = (src, options) => {
       {
         test: isTypeScriptSource,
         plugins: [
-          [require('@babel/plugin-transform-typescript'), {isTSX: false}],
+          [
+            require('@babel/plugin-transform-typescript'),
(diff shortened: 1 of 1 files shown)


cpojer:
I would recommend to manually add this preset in your own babel configuration with your own setting.

OzzieOrca:
You have any docs for doing that? I'm assuming that's not worth the effort of copying all the config out of the file I changed here and keeping it up to date.

Not the end of the world but this week I did stare at a syntax error in a `.ts` file for too long before realizing the problem was the file extension. `Parsing error: '>' expected` from eslint isn't a very developer friendly error. Oh well. That error is outside the scope of the project. But would be nice to avoid that completely and have consistent file extensions for all files regardless of whether they contain JSX.

PR #410, opened 2019-05-31
[breaking] Turn off providesModuleNodeModules by default

ide:
(This is a draft PR until we completely replace Haste in RN.)

**Summary**

Sets the default value of `providesModuleNodeModules` to an empty array, since Haste is going away from RN. **This is a breaking change and metro and metro-config's major semver needs to be bumped.**

**Test plan**

Modify my copy of metro-config under RN so that `providesModuleNodeModules` defaults to an empty array, start the packager with `--reset-cache`, and load the RNTester bundle successfully.


ide:
cpojer is taking this even further with a grander sacrifice to the grim code reaper, closing this out.

PR #425, opened 2019-06-26
add Chinese

wangliguang:
**Summary**

I am Chinese RN developer. When I  look for Metro docs, translating all metro docs. I hope can help more Chinese  developer reduce the metro the cost of study

**Test plan**

I don't have change any core js file.


wangliguang:
what happen ?

PR #428, opened 2019-06-27
👍 FEAT: add_chinese

wangliguang:
**Summary**
I am Chinese RN developer. When I look for Metro docs, translating all metro docs. I hope can help more Chinese developer reduce the metro the cost of study

**Test plan**

I don't have change any core js file.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -9,9 +9,9 @@ title: API
 
 ---
 
-## Quick Start
+## 快速开始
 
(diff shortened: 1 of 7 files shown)


cpojer:
Thank you! Unfortunately this will change the language for all these files for everyone. For now I recommend you maintain a fork with Chinese language :)

relign:
expect to add Chinese documents

PR #432, opened 2019-07-16
fix: bump jest-haste-map

SimenB:
**Summary**

This includes https://github.com/facebook/jest/pull/8558

**Test plan**

Dunno... Green CI? 😀


package.json
```diff
@@ -22,7 +22,7 @@
     "codecov": "^2.2.0",
     "debug": "^2.2.0",
     "eslint": "5.1.0",
-    "eslint-config-fb-strict": "22.1.0",
+    "eslint-config-fb-strict": "24.0.0",
```
SimenB: Slightly better...

Before:
image

After:
image

scripts/build.js
```diff
@@ -99,7 +99,7 @@ function buildFile(file, silent) {
       );
   } else {
     const transformed = prettier.format(babel.transformFileSync(file, {}).code, {
-      parser: 'babylon',
+      parser: 'babel',
```
SimenB: image

package.json
```diff
@@ -36,7 +36,7 @@
     "glob": "^7.1.1",
     "istanbul-api": "^1.1.0",
     "istanbul-lib-coverage": "^1.0.0",
-    "jest": "^24.7.1",
+    "jest": "^24.8.0",
```
SimenB: why not

cpojer:
Unfortunately we can't land this as is. Metro is tied together with RN at FB. I'll ask @scotthovestadt to do an update internally.

mikehardy:
@cpojer I would request a bump to 24.9.0 now but it appears that a PR for that would be closed based on your comment. Can I comment-request a bump to 24.9.0 then :-)

Without it, current node v12 startup time for react-native start is like 2.5 minutes vs 12 seconds so it's a dev experience issue. Unsure on correctness etc but as a minor it shouldn't impact things negatively to bump it (hopefully not famous last words). Thanks!

SimenB:
nice necropost codecov

Issue #414, opened 2019-06-08
npm reports vulnerability on dependency "mem"


jsamr:
**Do you want to request a *feature* or report a *bug*?**

*bug*

**What is the current behavior?**

As per last commit on master branch (bbd72a46d37fb5a872029bde3990f8f14b10258d), metro uses `yargs@^9.0.0` which matches versions `9.0.0` and `9.0.1`. Both versions are exposed to a vulnerability with two indirections: `yargs@9.0.1 › os-locale@2.1.0 › mem@1.1.0`. The vulnerability affects `mem` versions strictly inferior to `4.0.0`.
[Here is snyk report](https://app.snyk.io/vuln/npm:mem:20180117).

**Solution**

Upgrade to the latest 13.X yargs version.

robhogan:
We're triaging older issues. Metro now uses `yargs@17`, and the reported `yargs` - `os-locale` - `mem` dependency path is no longer present. This report no longer applies.

PR #438, opened 2019-07-30
Upgrading to yargs@13.0.0 to remove dependency on mem package

timglass:
**Summary**
Remove dependency on os-locale which in turn has a dependency on mem package (commit [4dfa19b](https://github.com/yargs/yargs/commit/4dfa19b3275cbd74ed686c437d392c4612e237a4)), which has a vulnerability (WS-2018-0236).

This PR resolves issue #414.

**Test plan**
Run `npm test` both prior to and after checking out this branch, to ensure no existing tests are broken


LaurentGoderre:
Can we get this merged please?

cpojer:
Unfortunately we can't merge package updates via GitHub and have to do them at FB. We have a ton of stuff depending on Yargs so we may not be able to update this in Metro soon :( Sorry about that.

henrikra:
@cpojer Looks like the current master nowadays has the updated version. In which React Native this will be released?

covolution:
There's no fix version or milestone on this, its still in the latest version of expo ```expo@37 -> react-native@0.61.4 -> metro@0.56.4``` . How can we tell when there is a release that fixes this?

cpojer:
It's in Metro 0.59.

PR #445, opened 2019-08-20
fix: bump jest-haste-map et al

thymikee:
**Summary**

Bumping `jest-haste-map` and all related Jest packages btw to 24.9, so that [this patch](https://github.com/facebook/jest/pull/8787) is applied. I know it's in the semver range, but doing so will invalidate lockfiles of React Native users, whose Metro is transitive dependency of.

Hopefully we can get this in 0.56 release :)

**Test plan**

CI passes.


cpojer:
Unfortunately we cannot merge this because Metro is entangled with all other RN stuff at FB. It will need to be updated in a single commit across all projects at once. cc @scotthovestadt

mikehardy:
Note that of the two issues I've had to patch-package for jest-haste-map one is fixed but the other (https://github.com/facebook/jest/pull/8558) - which was released as 24.8.1 still just does not come down in a react-native 0.60.5 environment by default for some reason. I understand this might be a dependency mess internally but I would ❤️ for this to be done for RN0.61. Thanks guys!

PR #448, opened 2019-09-04
update default.js  to support slug text files

tdvkiran:
slug text system is a new text system which is used to render on applications using gpu and cpu
can go through about Slug here: https://sluglibrary.com/

**Summary**

Slug renders character glyphs on the GPU directly from outline data composed of quadratic Bézier curves to produce crisp text at any scale or from any perspective. There are no precomputed texture images or signed distance fields. Slug uses a breakthrough mathematical algorithm that we invented to achieve perfect robustness with high performance, and it is the only existing GPU method that renders properly antialiased glyph shapes with no artifacts under both magnification and minification.

Allows to use slug files similar to ttf font files

### DIFF
--- a/packages/metro-config/src/defaults/defaults.js
+++ b/packages/metro-config/src/defaults/defaults.js
@@ -44,6 +44,7 @@ exports.assetExts = [
   // Font formats
   'otf',
   'ttf',
+  'slug',
   // Archives (virtual files)
   'zip',
 ];


cpojer:
Thanks for your PR. I don't think we need to add this to the defaults here as it doesn't seem to be a very popular thing that most people use at this time. You can always change your own Metro config to add this.

Issue #453, opened 2019-09-27
metro-config sharedBlacklist regexp without scape "\"


lucasfontesgaspareto:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
When i try to run react-native start an error is outputed
```
error Invalid regular expression: /(.*\\__fixtures__\\.*|node_modules[\\\]react[\\\]dist[\\\].*|website\\node_modules\\.*|heapCapture\\bundle\.js|.*\\__tests__\\.*)$/: Unterminated character class. Run CLI with --verbose flag for more details.
SyntaxError: Invalid regular expression: /(.*\\__fixtures__\\.*|node_modules[\\\]react[\\\]dist[\\\].*|website\\node_modules\\.*|heapCapture\\bundle\.js|.*\\__tests__\\.*)$/: Unterminated character class
    at new RegExp (<anonymous>)
    at blacklist (C:\Users\***\react\myApp\node_modules\metro-config\src\defaults\blacklist.js:34:10)
    at getBlacklistRE (C:\Users\***\react\myApp\node_modules\@react-native-community\cli\build\tools\loadMetroConfig.js:69:59)
    at getDefaultConfig (C:\Users\***\react\myApp\node_modules\@react-native-community\cli\build\tools\loadMetroConfig.js:85:20)
    at load (C:\Users\***\react\myApp\node_modules\@react-native-community\cli\build\tools\loadMetroConfig.js:121:25)
    at Object.runServer [as func] (C:\Users\***\react\myApp\node_modules\@react-native-community\cli\build\commands\server\runServer.js:82:58)
    at Command.handleAction (C:\Users\***\react\myApp\node_modules\@react-native-community\cli\build\cliEntry.js:160:21)
    at Command.listener (C:\Users\***\react\myApp\node_modules\commander\index.js:315:8)
    at Command.emit (events.js:210:5)
    at Command.parseArgs (C:\Users\***\react\myApp\node_modules\commander\index.js:651:12)
```

**Please provide your exact Metro configuration and mention your Metro, node, yarn/npm - =-- - version and operating system.**
- See above for metro config.
- Windows 10 x64
- node v12.11.0
- npm@6.11.3
- metro@0.56.0
- metro-config@0.56.0

halkliff:
This bug is caused due to Node's parsing of regular expressions in Node versions >=12.11.0.

These regular expressions should be fixed without any compatibility issues with older Node versions...

Leminur:
[...]

PR #454, opened 2019-09-27
Fix regex scape #453

lucasfontesgaspareto:
**Summary**
In the first index of sharedBlacklist variable the regex writed was without scape, it is a path with purpose to exclude react/dist of metro process.

**Test plan**
Adding the slash scape "\\" before slash used in the path
executing ```react-native start``` i got it normally


cpojer:

Issue #330, opened 2018-12-11
`SHA-1 for file ... is not computed` when using `resolver.resolveRequest`


vjpr:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

I am using a custom `resolver.resolveRequest ` function in `rn-cli.config.js`. See https://facebook.github.io/metro/docs/en/configuration#resolver-options.

The error message is being printed from `metro/src/node-haste/DependencyGraph.js#getSha1`, which is being called by `metro/src/Bundler.js#transformFile`:

_metro/src/node-haste/DependencyGraph.js#getSha1_

```
  getSha1(filename: string): string {
    const resolvedPath = fs.realpathSync(filename);
    const sha1 = this._hasteFS.getSha1(resolvedPath);
    if (!sha1) {
      throw new ReferenceError(`SHA-1 for file ${filename} is not computed`);
    }
    return sha1;
  }
```

The issue is that the file won't exist in the haste map if you are resolving it in the `resolveRequest` hook.

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**

https://github.com/vjpr/expo-v2-test-pnpm/tree/broken

You need to run `pnpm install` (the reason for the custom resolver is because symlinks are not supported by metro and hence cause problems for pnpm).

**What is the expected behavior?**

`getSha1` should compute hash if not in haste map.

**Please provide your exact Metro configuration and mention your Metro, node, yarn/npm version and operating system.**

metro@0.45.6
node@8.9.0
pnpm@2.21.1

## Workaround

Add to _metro/src/node-haste/DependencyGraph.js#getSha1_:

```
    if (!sha1) {
      return getFileHash(resolvedPath)
      function getFileHash(file) {
        return require('crypto')
          .createHash('sha1')
          .update(fs.readFileSync(file))
          .digest('hex')
      }
    }
```

I have released a patched version here: Use `metro-pnpm@0.45.6-vjpr.2`

Stephen2:
I got this when trying to get Expo to work with yarn workspaces, using `expo-yarn-workspaces`

pavermakov:
I've got the same issue when I tried to upgrade react-native to 0.57.0. I followed the steps mentioned the change log (https://github.com/react-native-community/react-native-releases/blob/master/CHANGELOG.md#057), but had no luck.

mareksuscak:
@Stephen2 me too, did you solve it?

[...]

PR #455, opened 2019-09-30
Update DependencyGraph.js

louisjacksonrees:
This is a PR to fix the issue outlined here: #330

### DIFF
--- a/packages/metro/src/node-haste/DependencyGraph.js
+++ b/packages/metro/src/node-haste/DependencyGraph.js
@@ -193,12 +193,12 @@ class DependencyGraph extends EventEmitter {
     const sha1 = this._hasteFS.getSha1(resolvedPath);
 
     if (!sha1) {
-      return getFileHash(resolvedPath)
+      return getFileHash(resolvedPath);
       function getFileHash(file) {
         return require('crypto')
           .createHash('sha1')
           .update(fs.readFileSync(file))
-          .digest('hex')
+          .digest('hex');
       }
     }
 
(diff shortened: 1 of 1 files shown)


crramirez:
Works for me thanks a lot

jamonholmgren:
Looks like the CI issue is just a linting error -- missing a semicolon.

DuncanWalter:
👋 Just bumping this PR. I've been running into issue #330 trying to work with linked packages, and would love to have this change or something equivalent published

BaderSerhan:
Hello. Thanks for the PR! I started to have this issue out of nowhere! Any idea on when will it be merged?

mjesun:
Hi! Original implementor of the Metro caching strategies here 🙂

Metro uses different techniques to watch for file changes in the filesystem. It does that by relying in `jest-haste-map`, which implements two strategies for watching: `watchman`, if it's available on your machine, and `fs.watch` otherwise. In either case, `jest-haste-map` provides you the SHA-1s of the files being watched.

When `jest-haste-map` *isn't watching a file*, the SHA-1 will not be returned. Of course we could have computed it the SHA ourselves at first in the `if` where this PR is adding it - and this would make Metro build -, but if you tried changing the file nothing will happen, because the file is not monitored. To make Metro monitor the file, you need to add the folder where it's placed to the `projectRoots` configuration setting.

I however agree that the message is not great. It explains the internal error but not why this is important or any actionable feedback; it should mention why the SHA-1 is important (for caching purposes, why it couldn't be obtained (because you require a file not being watched by Metro), and how to fix it (make your file(s) watchable by adding the folder in which they are to `projectRoots`).

cpojer:
Thank you for the thorough explanation @mjesun. I'll close this PR but I'm happy to merge a new PR that improves on this message.

PR #461, opened 2019-10-12
Fixing Regex Bug in Node.js v12.11.0+ on Windows and improve security with npm audit fix -force

ralic:
Referenced Info : 
https://stackoverflow.com/questions/58120990/question-getting-error-on-react-native-start#


cpojer:

PR #465, opened 2019-10-23
Pass console logs through the reporter interface

fson:
**Summary**

Sending `console` logs from the device where the code is running using HMRClient was implemented in https://github.com/facebook/metro/commit/2d590a6157ce2587a8737c59a4b17344385151ba. This is nice, but currently these logs are printed directly to console, without a chance to customize the output format when Metro is running inside another tool, such as [Expo CLI](https://github.com/expo/expo-cli).

There's already a nice custom [reporting interface](https://github.com/facebook/metro/blob/v0.56.1/packages/metro/src/lib/reporting.js) in Metro that allows customizing the output format. For example, in Expo CLI we use a custom reporter similar to [JsonReporter](https://github.com/facebook/metro/blob/v0.56.1/packages/metro/src/lib/JsonReporter.js) to get all the Metro logging as JSON and then implement things like a web UI for viewing logs on top of that. This would be really difficult, if we had to parse everything from the regular text based logs meant to be read by humans, not computers.

This PR changes the `console` logging introduced in Metro v0.56.0 to pass these logs through the reporting interface, so that custom reporters can format them as necessary. The default implementation in `TerminalReporter` remains the same, without changes to the log output.

**Test plan**

Verify that all the tests pass.

### DIFF
--- a/packages/metro/src/HmrServer.js
+++ b/packages/metro/src/HmrServer.js
@@ -18,7 +18,6 @@ const debounceAsyncQueue = require('./lib/debounceAsyncQueue');
 const formatBundlingError = require('./lib/formatBundlingError');
 const getGraphId = require('./lib/getGraphId');
 const hmrJSBundle = require('./DeltaBundler/Serializers/hmrJSBundle');
-const logToConsole = require('./lib/logToConsole');
 const nullthrows = require('nullthrows');
 const parseOptionsFromUrl = require('./lib/parseOptionsFromUrl');
 const splitBundleOptions = require('./lib/splitBundleOptions');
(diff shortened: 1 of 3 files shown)


koofka:
@fson thanks for contributing this update.

We are attempting to implement a custom reporter in order to quiet the logging in the metro console to no avail.  It appears to intercept the build related events, but the console.log calls from within React Native continue to appear in the metro terminal.  After the last `bundle_build_done` event, the application consoles appear to bypass the metro_config reporter and we haven't had any luck tracking down where they are being passed.  Any thoughts / advice? 

Thanks much for this.

Issue #355, opened 2019-02-07
Watch option is ignored when creating JestHasteMap


hfter789:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
watch is hardcoded to true when creating JestHasteMap: https://github.com/facebook/metro/blob/master/packages/metro/src/node-haste/DependencyGraph.js#L96

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**
https://github.com/hfter789/metro
`jest packages/metro/src/integration_tests/__tests__/buildGraph-test.js`

**What is the expected behavior?**
JestHasteMap should receive the `watch` value coming from metro config.

FLGMwt:
This seems to have been introduced by FB internal changes here: https://github.com/facebook/metro/commit/0d6c135045a33f2b9effa8ca31b648e6b29ec24c . We were hoping to use the `watch: false` option in our React Native release build to prevent having to set `fs.inotify.max_user_watches` on our CI server.

I'm not very clear on the metro/haste-map internals, but it seems odd that a static build (when doing `bundle`) would need file watching.

Is there any context as to why this config option is ignored as of 6 months ago?

hypest:
> We were hoping to use the `watch: false` option in our React Native release build to prevent having to set `fs.inotify.max_user_watches` on our CI server.

Sorry for the possible noise on a ticket that might be unrelated but, wanted to add that similar to @FLGMwt , [our project](https://github.com/wordpress-mobile/gutenberg-mobile) needs to have the `bundle` command not watch for files on CI (JitPack in our case). Would appreciate an pointers on how to suppress file watching in `react-native bundle` if possible. 😬 🙇

debben:
I know nothing of react-native but have been tasked with figuring out the same max_user_watches issue when building a react-native project in CI. I can't think of a reason file watchers for js builds should be needed when just bundling for CI.

As I started digging through the rabbit hole of react-native, the cli, and these tools, I arrived at the watch metro config option. Despite 0d6c135 it would seem the docs still mention watch as an option. Are there any alternatives to short of the above suggestion about replacing DependencyGraph? I'm also interested in the context of why this option was removed @rafeca @mjesun

bonham000:
[...]

PR #491, opened 2019-12-05
Add watch flag to metro config

cdsanchez:
**Summary**

This is an attempt enable users of Metro to configure the `watch` flag that is passed to `JestHasteMap`. This addresses #355 

Without this change, we see failures on some of our internal build machines where we cannot easily configure `fs.inotify.max_user_watches`. 

**Test plan**

I have a corresponding PR to the react-native-cli: https://github.com/react-native-community/cli/pull/879. I used that change (via `yarn link`) to test this change in metro.

Invoked metro via `react-native start`
- Observed that the `watch` flag was true through `console.log`

Invoked metro via `react-native bundle`
- Observed that `watch` was set to false.

If there are any other test cases you would like me to run, please let me know!

### DIFF
--- a/packages/metro-config/src/__tests__/__snapshots__/loadConfig-test.js.snap
+++ b/packages/metro-config/src/__tests__/__snapshots__/loadConfig-test.js.snap
@@ -121,6 +121,7 @@ Object {
     "workerPath": "metro/src/DeltaBundler/Worker",
   },
   "transformerPath": "",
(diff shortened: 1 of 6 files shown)


cpojer:
Thank you for your PR. There is a similar PR with https://github.com/facebook/metro/pull/497. Let's use that one to discuss further.

PR #493, opened 2019-12-11
[fix] Use absolute path, not relative path, when calculating delta

noahsug:
**Summary**
When an absolute dependency path changes but the relative path stays the same (e.g. an extension change), DeltaBundler does not pick up the change. 

This ultimately leads to a "no such file or directory" error when trying to require a dependency with the incorrect extension.

To repro, create a file `foo.js` which imports `bar.js`:
```
// foo.js
import `bar`
```
 1. initialize the graph via `incrementalBundler.initializeGraph()`
 2. rename `bar.js` -> `bar.ts`
 3. update the graph via `incrementalBundler.updateGraph()`
 4. notice that the resulting graph still has `bar.js` listed as a dependency

**Test plan**

After apply these changes, I confirmed the delta now includes 1 added file (`bar.ts`), 1 modified file (`foo.js`) and 1 deleted file (`bar.js`) (where as before it only included the 1 modified file). I also confirmed that the "no such file or directory" error we were seeing at Airbnb went away.

### DIFF
--- a/packages/metro/src/DeltaBundler/traverseDependencies.js
+++ b/packages/metro/src/DeltaBundler/traverseDependencies.js
@@ -227,7 +227,11 @@ async function processModule<T>(
   }
 
   for (const [relativePath, dependency] of previousDependencies) {
-    if (!currentDependencies.has(relativePath)) {
+    if (
+      !currentDependencies.has(relativePath) ||
+      nullthrows(currentDependencies.get(relativePath)).absolutePath !==
+        dependency.absolutePath
+    ) {
       removeDependency(module, dependency.absolutePath, graph, delta);
     }
   }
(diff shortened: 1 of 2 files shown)


packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -227,7 +236,8 @@ async function processModule<T>(
   }
 
   for (const [relativePath, dependency] of previousDependencies) {
-    if (!currentDependencies.has(relativePath)) {
+    if (!hasSameAbsolutePath(
```
raejin: Should we remain the original logic, and adding an additional condition to check for absolute file path? This perhaps can clarify what this additional condition is doing. Since `hasSameAbsolutePath` does two checks.

Additionally, instead of calling it checking "absolute path", perhaps we want to rename it to something along the lines of "has the same extension"?

noahsug: I wanted to make it clear that the logic is changing from "check if the relative path has changed" to "check if the absolute path has changed"

packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -190,6 +190,16 @@ async function traverseDependenciesForSingleFile<T>(
   options.onDependencyAdded();
 }
 
+function absolutePathHasChanged(
+  dependencyA: Dependency,
+  dependencyB: Dependency,
+): Boolean {
+  if (dependencyA == null || dependencyB == null) {
```
raejin: Just wondering, when would `dependencyB` be `null`?

noahsug: I decided to remove this function and just do the check inline for simplicity

PR #495, opened 2019-12-12
Add mac as a platform in Platform.select inline

charpeni:
**Summary**

Related to https://github.com/facebook/react-native/pull/27484. Once this PR will be merged, I'll send a PR to also support `Platform.select` with `mac` in React Native.

This adds support for mac as a platform in `Platform.select` inline.

Mac is not properly a platform, but we should be able to target it with `Platform.select`. I'm not sure if this is a good idea or not, but it seems to me. 🤔 

#### Inlines `Platform.select` in the code to `ios` and `mac` if both are present and running on `ios` platform

```javascript
function a() {
  var a = Platform.select({ios: 1, android: 2, mac: 3});
}
```

Inlines into:

```javascript
function a() {
  var a = Platform.select({ios: 1, mac: 3});
}
```

#### Doesn't inline `Platform.select` if the current platform is `ios`, and `ios` and `mac` are in the properties

```javascript
function a() {
  var a = Platform.select({ios: 1, mac: 2});
}
```

Stays the same, as we're running on iOS, and we don't know if we're on macOS or not.

#### Doesn't inline `Platform.select` if the current platform is `ios` and `mac` is in the properties, but `ios` is not

```javascript
function a() {
  var a = Platform.select({mac: 1, default: 2});
}
```

Stays the same, as the `ios` platform could either be using `mac` or `default` for iOS.

**Test plan**

See changes in tests.

### DIFF
--- a/packages/metro/src/JSTransformer/worker/__tests__/inline-plugin-test.js
+++ b/packages/metro/src/JSTransformer/worker/__tests__/inline-plugin-test.js
@@ -296,6 +296,19 @@ describe('inline constants', () => {
     });
   });
 
+  it('inlines Platform.select in the code to ios and mac if both are present and running on ios platform', () => {
+    const code = `
+      function a() {
+        var a = Platform.select({ios: 1, android: 2, mac: 3});
+      }
+    `;
+
+    compare([inlinePlugin], code, code.replace(' android: 2,', ''), {
+      inlinePlatform: 'true',
+      platform: 'ios',
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


robhogan:
Sorry this didn't get a response, thanks for working on a fix. Closing as stale as the plugin has moved out of this repo to React Native as of 0.88 (https://github.com/facebook/react-native/pull/57848 ). 

The motivating case is still live though afaik, and now compounded with VR, Apple Watch and others where we have overlapping APIs on distinct platforms. The `platform` API probably needs a re-think but that's a bigger discussion best opened on RN.

PR #497, opened 2019-12-12
Add watch parameter back to config for jest haste map configuration

alanjoxa:
**Summary**
This PR will fix the regression caused by [this commit ](https://github.com/facebook/metro/commit/0d6c135045a33f2b9effa8ca31b648e6b29ec24c#diff-75dc57ca2d6ab410090adb917784434f)

In the [Metro documentation we have a `watch` field](https://facebook.github.io/metro/docs/en/configuration#watch), which is not working after the above commit. 

We cannot simply revert that commit because it will set the default value of the watch field to `false`, but the current default value is `true`. So I created this PR by reverting that commit and setting its default value to `true`. Which make this commit a no-op to current functionality, but fix the regression.

**Motivation** 

*Why this fix is required now ?*
Apart from fixing the bug, this enables react-native app developers to build there app in cloud servers, which has limit in no. of file watch at a time. This is a pain for people who have large dependencies, which causes the `node_modules` to have a large no. of files, which cause the build to fail with `ENOSPC` error.

**Test plan**
Apart from the completing the [metro pull request workflow](https://github.com/facebook/metro/blob/master/CONTRIBUTING.md#workflow-and-pull-requests)
I tested it by applying this patch to a sample react-native feature and it is working as expected. Also, since it is almost a revert of one of the previous commit, it has less foreign changes.

The test feature which I created is available here - https://github.com/alanjoxa/AwesomeProject
Test simulation instructions
```
git clone https://github.com/alanjoxa/AwesomeProject.git
cd AwesomeProject
npm install
npm start
```
This will start a metro server, now you can change the watch property in that [project's metro config](https://github.com/alanjoxa/AwesomeProject/blob/master/metro.config.js#L9), and see the difference, by making a build request at  by a GET at `http://localhost:8081/index.bundle?platform=ios`. 
When 

- `watch : true` - Metro server will watch for  all files in the project folder, including `node_modules` and apply any file changes to the bundle on the subsequent build request.
- `watch: false` - Will not watch for files, that means we need to restart the server for seeing the file changes. 

This file watching functionality will help in accelerating the local development, but that is not expected when we are building the project to release or in our CI pipelines, where the build is happening in some docker containers or some ec2 servers, where sometimes we do not have access to set the `fs.inotify.max_user_watches`. So people try the workarounds like, npm script for deleting the folders in the node_modules which are not required in the bundling process.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -49,4 +49,4 @@ untyped-import
 untyped-type-import
 
 [version]
-^0.113.0
+^0.114.0
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "eslint-plugin-react": "7.6.1",
     "eslint-plugin-relay": "1.4.1",
     "fbjs-scripts": "^1.1.0",
-    "flow-bin": "^0.113.0",
+    "flow-bin": "^0.114.0",
     "glob": "^7.1.1",
     "istanbul-api": "^1.1.0",
     "istanbul-lib-coverage": "^1.0.0",
--- a/packages/metro-config/src/__tests__/__snapshots__/loadConfig-test.js.snap
+++ b/packages/metro-config/src/__tests__/__snapshots__/loadConfig-test.js.snap
@@ -121,6 +121,7 @@ Object {
     "workerPath": "metro/src/DeltaBundler/Worker",
   },
   "transformerPath": "",
+  "watch": true,
   "watchFolders": Array [
     "/",
   ],
@@ -248,6 +249,7 @@ Object {
     "workerPath": "metro/src/DeltaBundler/Worker",
   },
   "transformerPath": "",
+  "watch": true,
   "watchFolders": Array [
     "/",
   ],
--- a/packages/metro-config/src/__tests__/convertConfig-test.js
+++ b/packages/metro-config/src/__tests__/convertConfig-test.js
@@ -47,6 +47,7 @@ describe('convertConfig', () => {
       minifierPath: DEFAULT_METRO_MINIFIER_PATH,
       port: 8080,
       reporter: new TerminalReporter(new Terminal(process.stdout)),
+      watch: true,
     });
 
     expect(prettyFormat(convertedConfig)).toEqual(prettyFormat(defaultConfig));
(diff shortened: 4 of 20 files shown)


cpojer:
Thank you for this PR. Instead of adding back this config option, could we instead change it so that Metro detects whether it is run via CI, and switches the watch flag based on that? I would much prefer that as a solution and there are packages like `is-ci` which we could use for this.

cdsanchez:
@cpojer I am not sure if `is-ci` would detect our internal build server. We would need to set `ci.isCI` to true manually, which less ideal than having the config colocated with metro IMO.

In #491 I considered "building an offline bundle" as the signal for flipping off `watch`. One benefit is that it will also disable file watchers for users building offline bundles on their workstation (IDE or TTY) in addition to CI. It wasn't clear could metro know it's building an offline bundle on its own, so it's a config flag passed down by the RN cli that is only set in the offline bundle scenario.

alanjoxa:
> Thank you for this PR. Instead of adding back this config option, could we instead change it so that Metro detects whether it is run via CI, and switches the watch flag based on that? I would much prefer that as a solution and there are packages like `is-ci` which we could use for this.

That is an excellent idea, because majority of the metro users(the react-native feature developers) don't have much idea about these metro configurations. And getting an error like `ENOSPC` in their CI servers and which are not reproducible in their local system will be frustrating. Having a detection of such CI environment and auto setting the watch flag to false will be really a win. 

Thank you for that direction.

I just looked at the [`is-ci`](https://github.com/watson/is-ci) which internally uses the [`ci-info`](https://github.com/watson/ci-info), is using a whitelist of major CI vendors in a [json file](https://github.com/watson/ci-info/blob/master/vendors.json) for identifying the CI environment. So by making use of this `ci-info` in metro will support all the CI's listed in it whitelist. Which covers the most of the cases and those who need to use any other server can use the watch flag.  

I will update this PR to support this auto CI detection as well.

alanjoxa:
Following up in a new PR : https://github.com/facebook/metro/pull/502 , because this PR got some merge issues.

PR #498, opened 2019-12-17
Fix invalid regular expression on Windows

t0rr3sp3dr0:
**Summary**

When invoking `react-native run-windows`, an `Invalid regular expression` error is thrown by Metro due to path separator being `\` on Windows. Metro tries to replace every `/` with the path separator, but this makes expressions like `[/\\]` in `[\\\]`, an invalid regular expression.

A solution would be replace only escaped `/` with the actual escaped path separator. instead of replacing any `/`:

https://github.com/facebook/metro/blob/d88295cc17453c74ffcb72627dd5e0f97886a55a/packages/metro-config/src/defaults/blacklist.js#L26

**Test plan**

```powershell
react-native init PoC --version 0.60.6
cd PoC
yarn add rnpm-plugin-windows
react-native windows --template vnext
yarn start
```

*Currently*

image

*Expected*

image

### DIFF
--- a/packages/metro-config/src/defaults/blacklist.js
+++ b/packages/metro-config/src/defaults/blacklist.js
@@ -22,7 +22,8 @@ var sharedBlacklist = [
 
 function escapeRegExp(pattern) {
   if (Object.prototype.toString.call(pattern) === '[object RegExp]') {
-    return pattern.source.replace(/\//g, path.sep);
+    // convert the '/' into an escaped local file separator
+    return pattern.source.replace(/\\\//g, '\\' + path.sep);
   } else if (typeof pattern === 'string') {
     var escaped = pattern.replace(/[\-\[\]\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
     // convert the '/' into an escaped local file separator


cpojer:
I don't think this is the right fix for your issue. Most likely your Metro configuration has an issue. Could you paste your config here so we can take a look?

t0rr3sp3dr0:
@cpojer, here follows the `metro.config.js`:

```js
/**
 * Metro configuration for React Native
 * https://github.com/facebook/react-native
 *
 * @format
 */
const fs = require('fs');
const path = require('path');
const blacklist = require('metro-config/src/defaults/blacklist');

const rnPath = fs.realpathSync(
  path.resolve(require.resolve('react-native/package.json'), '..'),
);
const rnwPath = fs.realpathSync(
  path.resolve(require.resolve('react-native-windows/package.json'), '..'),
);

module.exports = {
  resolver: {
    extraNodeModules: {
      // Redirect react-native to react-native-windows
      'react-native': rnwPath,
      'react-native-windows': rnwPath,
    },
    // Include the macos platform in addition to the defaults because the fork includes macos, but doesn't declare it
    platforms: ['ios', 'android', 'windesktop', 'windows', 'web', 'macos'],
    providesModuleNodeModules: ['react-native-windows'],
    // Since there are multiple copies of react-native, we need to ensure that metro only sees one of them
    // This should go in RN 0.61 when haste is removed
    blacklistRE: blacklist([
      new RegExp(
        `${(path.resolve(rnPath) + path.sep).replace(/[/\\\\]/g, '[/\\\\]')}.*`,
      ),

      // This stops "react-native run-windows" from causing the metro server to crash if its already running
      new RegExp(
        `${path
          .resolve(__dirname, 'windows')
          .replace(/[/\\\\]/g, '[/\\\\]')}.*`,
      ),
    ]),
  },
  transformer: {
    getTransformOptions: async () => ({
      transform: {
        inlineRequires: false,
      },
    }),
  },
};
```

t0rr3sp3dr0:
@cpojer, this problem was partially solved on #498. The patch made was actually just a bypass on `blacklist` function to avoid the `[/\\]` expression. Previously when calling this function on Windows, even with an empty array, would cause the same exception for the same reason. Now it just triggers the exception when someone explicitly pass expressions like `[/\\]` as parameter to the `blacklist` function. Expressions that is indeed a valid regular expressions and should be accepted.

PR #503, opened 2019-12-29
Fix the RN bundling in CI servers by checking for CI servers and turn…

alanjoxa:
**Summary**

*Issue : https://github.com/facebook/metro/issues/355*
*Follow up on my old PR : https://github.com/facebook/metro/pull/497*

The RN app build process is failing in CI servers due to the always-on file watch of metro. Mostly the CI servers will have a limit on no. of file watch at a time. This is a problem for RN apps with large no. of dependencies, and thus with a large no. of files in `node_modules`. Which cause the build to fail with `ENOSPC` error. This PR will detect the CI servers with [`ci-info`](https://github.com/watson/ci-info/) module and turn off the [file-watching of JestHasteMap in  metro](https://github.com/facebook/metro/blob/f6314e43071ae498b41d9e5579ad4a13941a4baa/packages/metro/src/node-haste/DependencyGraph.js#L95).
~~Also in the [Metro documentation we have a `watch` field](https://facebook.github.io/metro/docs/en/configuration#watch), but it is not working since [this commit ](https://github.com/facebook/metro/commit/0d6c135045a33f2b9effa8ca31b648e6b29ec24c#diff-75dc57ca2d6ab410090adb917784434f). So this PR will also make it working, so that people who wish to turn off the watch can do that with their metro config.~~ And this change is a no-op to current metro users, as it will maintain the default watch behavior.

Also we can manually turn off the file watch by setting the `CI=true` environment variable.

**Motivation** 

*Why this fix is required now ?*
This file watching functionality will help in accelerating the local development, but that is not expected when we are building the project to release in our CI pipelines, where the build is happening in some docker containers or some ec2 servers, where sometimes we do not have access to set the `fs.inotify.max_user_watches`. So people try the workarounds like, npm script for deleting the folders in the node_modules which are not required in the bundling process. 

**Test plan**
Completing the [metro pull request workflow](https://github.com/facebook/metro/blob/master/CONTRIBUTING.md#workflow-and-pull-requests).
And I tested the functionality by creating a [sample RN app](https://github.com/alanjoxa/AwesomeProject) by applying this [patch](https://github.com/alanjoxa/AwesomeProject/tree/master/patches) to its metro dependencies in node-modules, as part of its [postinstall script](https://github.com/alanjoxa/AwesomeProject/blob/e52b92fd04d61e9ed8530d5fdf7f59657b1a8a2e/package.json#L11). It is working as expected.

**Test simulation instructions**
```
git clone https://github.com/alanjoxa/AwesomeProject.git
cd AwesomeProject
npm install
npm start
```
This will start a metro server and serve the bundles by watching all files in the project root. We can turn off the file watch by restarting it by `CI=true npm start`. 
A sample bundle can be triggered by making a GET request at `http://localhost:8081/index.bundle?platform=ios`. 
When in

- `npm start` - Metro server will watch for  all files in the project folder, including `node_modules` and apply any file changes to the bundle on the subsequent build request.
- `CI=true npm start` - Will not watch for files, that means we need to restart the server for seeing the file changes.


cpojer:
This is great, thank you so much for iterating on your approach. `ci-info` detects various environment variables to verify whether it is run on "CI": https://github.com/watson/ci-info/blob/master/index.js#L52

Do you have a special need for using the `watch` config option via Metro or could you instead set the CI variable, like `CI=true <run-metro>` and could we do away with the configuration option?

alanjoxa:
That make sense, and it works perfectly fine for my case as well. Just that I tend to retain the `watch` flag under the impression that it is something missing as per the documentation.

But now I understand that it is kind of a redundant config when the `CI` env variable of `ci-info` is available for the same purpose.

I will update the PR to remove the watch config completely, even from the documentation.
Let me know if that works ?

cpojer:
That sounds great to me. Metro isn't supposed to have a watch flag, and it shouldn't watch on CI environments. Seems like we can get this to work with the minimal amount of code and configuration :)

alanjoxa:
Awesome. Thanks, will get that done asap.

cdsanchez:
@cpojer CI isn't the only case where file watchers should be disabled. It should also be disabled when running the react-native CLI `bundle` subcommand. What's the recommendation for disabling it in the RN CLI, which invokes `metro` programmatically?

alanjoxa:
The `CI` env variable can  be set programmatically by `process.env.CI = true` before calling the metro's bundle API. An example for it is there in [metro's babel-transformer](https://github.com/facebook/metro/blob/984aab84358aeef8de13ba86c33834a1c1bf6926/packages/metro-react-native-babel-transformer/src/index.js#L193), where it is setting the babel env before calling the babel transform.

cdsanchez:
> The `CI` env variable can be set programmatically by `process.env.CI = true` before calling the metro's bundle API. An example for it is there in [metro's babel-transformer](https://github.com/facebook/metro/blob/984aab84358aeef8de13ba86c33834a1c1bf6926/packages/metro-react-native-babel-transformer/src/index.js#L193), where it is setting the babel env before calling the babel transform.

Yep, I'm aware that it can be set programmatically. However, I would like to confirm that we are indeed going with the `CI` env var even for cases that aren't really "CI" (i.e. a dev workspace / IDE). Beyond the conceptual misalignment, we should be aware that setting `CI=true` could potentially have other unintended side effects since it's effectively a shared global variable.

If we're good with that, I can update my PR https://github.com/react-native-community/cli/pull/879

cpojer:
Yeah that’s a good point, and I was thinking about that overnight as well. I think I would prefer the bundle command to force the the watch flag to be false but without adding a globally configurable watch flag. That can be done in a follow-up PR to this one.

cdsanchez:
> Yeah that’s a good point, and I was thinking about that overnight as well. I think I would prefer the bundle command to force the the watch flag to be false but without adding a globally configurable watch flag. That can be done in a follow-up PR to this one.

Makes sense @cpojer. I'll work with @alanjoxa on that follow-up PR.

alanjoxa:
Yeah that make sense. There is no reason for the metros's [runBuild](https://facebook.github.io/metro/docs/en/api#async-runbuildconfig-options) API to start a file watcher, as it is a one time build process, where file watching is relevant only in the case of a persistent bundle server.

Issue #500, opened 2019-12-20
Circular module dependency causes incorrect dependency removal calculation


raejin:
Disclaimer: We're not using the latest version of Metro, which is on v0.55.0. But  this behavior remains the same with the latest release, as demonstrated in the example repo.

I have a proposed fix: https://github.com/facebook/metro/pull/506

**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When a dependency graph with entrypoint `E` is removing a dependency `B` from the dependency graph. If the dependency `B` has inverse dependency after removing `E` from the inverse dependency, the resulting dependency graph with entrypoint `E` will be incorrect.

Example dependency graph with entrypoint `E`
![image](https://user-images.githubusercontent.com/1754430/71216742-2fb08300-2270-11ea-8ac5-2ba47d26ed6f.png). 

If we're removing dependency B from the graph, we'd end up with no change in dependency graph. And Metro will produce graph revision looks like the following:

```
{
  "deltaSize":{"added":4,"modified":1,"deleted":0},
  "dependencies":[
    "E.js",
    "B.js",
    "U.js",
    "I.js",
    "A.js"
  ],
  "entryPoints":[
    "E.js"
  ]
}
```

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**
https://github.com/raejin/remove-dependency-metro-bug

**What is the expected behavior?**
 The correct dependency graph output should've been:
```
{
  "deltaSize":{"added":0,"modified":1,"deleted":2},
  "dependencies":[
    "E.js",
    "U.js",
    "A.js"
  ],
  "entryPoints":[
    "E.js"
  ]
}
```

Which should match the removal:
![image](https://user-images.githubusercontent.com/1754430/71217012-2e338a80-2271-11ea-8562-226fa2737c28.png)

**Please provide your exact Metro configuration and mention your Metro, node, yarn/npm version and operating system.**

Metro 0.55.0 & 0.57.0
npm 6.13.4
node v10.16.0

PR #506, opened 2020-01-03
Handle circular dependency upon dependency removal

raejin:
**Summary**
This is a proposed fix for the issue I submitted https://github.com/facebook/metro/issues/500. For problem statement, please refer to the issue linked to understand the scope of the bug.

Currently Metro does not handle dependency removal correctly in the presence of certain circular dependency. Since we rely on Metro server to produce correct dependency graph in hopes to bundle only the needed modules. This is especially an issue when a dependency is removed from the entry point, however, due to it having circular dependency, the updated dependency graph will leave out all its dependencies in the graph.

I added the following two more test cases to ensure the correctness of the algorithm:

1. Remove `B` from `E`: `removes a dependency with transient cyclic dependency`
image

2. Remove `B` from `E`: `removes a cyclic dependency which is both inverse dependency and direct dependency`
image

3. Remove `B` from `E`: `removes a sub graph that has internal cyclic dependency`
image

Feel free to propose more tests to ensure the correctness of the algorithm. I'm aware that this may introduce more expensive graph updates for certain scenarios, but I believe that ensuring the correctness is far more important for dependency graph updates.

## Implementation
Previously, our implementation will stop removing circular dependency due to it having remaining `inverseDependencies`:

https://github.com/facebook/metro/blob/984aab84358aeef8de13ba86c33834a1c1bf6926/packages/metro/src/DeltaBundler/traverseDependencies.js#L328-L330

This can be illustrated by this example: 
image

When removing `B` from `E`, `B` will still have an inverse dependency `A` left. Henceforth, the rest of the graph remain untouched. My proposed solution is to have this async function `canSafelyRemoveFromParentModule` recursively checking all the inverse dependencies. In this example, we will look up inverse dependencies of `A` all the way to the end until there is no inverse dependency. We can only safely remove this dependency if and only if its end inverse dependency (in this case `A` will have `B` as its end inverse dependency) only has one path and the path is the same as the parent path. 

**Test plan**
- Updated 1 existing test, with 2 additional unit tests to ensure the correctness of the dependency removal logic handling with circular dependency.

```
yarn run jest packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
```

### DIFF
--- a/packages/metro/src/DeltaBundler/traverseDependencies.js
+++ b/packages/metro/src/DeltaBundler/traverseDependencies.js
@@ -226,24 +226,22 @@ async function processModule<T>(
     module.dependencies.set(relativePath, dependency);
   }
 
-  await Promise.all(
-    Array.from(previousDependencies.entries())
-      .filter(
-        ([relativePath, dependency]) =>
-          !currentDependencies.has(relativePath) ||
-          nullthrows(currentDependencies.get(relativePath)).absolutePath !==
-            dependency.absolutePath,
-      )
-      .map(([relativePath, dependency]) =>
-        removeDependency(
-          module,
+  Array.from(previousDependencies.entries())
+    .filter(
+      ([relativePath, dependency]) =>
+        !currentDependencies.has(relativePath) ||
+        nullthrows(currentDependencies.get(relativePath)).absolutePath !==
           dependency.absolutePath,
-          graph,
-          delta,
-          new Set(),
-        ),
+    )
+    .forEach(([relativePath, dependency]) =>
+      removeDependency(
+        module,
+        dependency.absolutePath,
+        graph,
+        delta,
+        new Set(),
       ),
-  );
+    );
 
   // Check all the module dependencies and start traversing the tree from each
   // added and removed dependency, to get all the modules that have to be added
@@ -368,14 +366,14 @@ function canSafelyRemoveFromParentModule<T>(
   delta: Delta,
 ): boolean {
   const visited = new Set();
-  const result = getAllTopLevelInverseDependencies(
+  const topInverseDependencies = getAllTopLevelInverseDependencies(
     inverseDependencies,
     graph,
-    '',
+    '', // current module name
     visited,
   );
 
-  if (!result.size) {
+  if (!topInverseDependencies.size) {
     /**
      * This happens when parentModule and inverseDependencies has circular dependency,
      * this will eventaully became an empty set due to `visited` Set being the base case
@@ -384,9 +382,9 @@ function canSafelyRemoveFromParentModule<T>(
     return true;
   }
 
-  const filterNotDeletedResult = Array.from(result).filter(
-    x => !delta.deleted.has(x),
-  );
+  const undeletedInverseDependencies = Array.from(
+    topInverseDependencies,
+  ).filter(x => !delta.deleted.has(x));
 
   /**
    * We can only mark `visited` set of modules to be safely removable if
@@ -394,12 +392,11 @@ function canSafelyRemoveFromParentModule<T>(
    *   this can happen when trying to see if we can safely remove from
    *   a module that was deleted. This is why we filtered them out with `delta.deleted`
    * 2. We have one top module and it is parentModule
-   *
    */
   const canSafelyRemove =
-    !filterNotDeletedResult.length ||
-    (filterNotDeletedResult.length === 1 &&
-      filterNotDeletedResult[0] === parentModule);
+    !undeletedInverseDependencies.length ||
+    (undeletedInverseDependencies.length === 1 &&
+      undeletedInverseDependencies[0] === parentModule);
 
   if (canSafelyRemove) {
     visited.forEach(mod => {
@@ -409,7 +406,7 @@ function canSafelyRemoveFromParentModule<T>(
   return canSafelyRemove;
 }
 
-async function removeDependency<T>(
+function removeDependency<T>(
   parentModule: Module<T>,
   absolutePath: string,
   graph: Graph<T>,
@@ -418,7 +415,7 @@ async function removeDependency<T>(
   // module(s) that we're sure can be removed. This will skip expensive
   // inverse dependency traversals.
   canBeRemovedSafely: Set<string> = new Set(),
-): Promise<void> {
+): void {
   const module = graph.dependencies.get(absolutePath);
 
   if (!module) {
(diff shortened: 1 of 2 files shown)


packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -309,12 +312,71 @@ async function addDependency<T>(
...
+  inverseDependencies: Set<string>,
+  parentModule: string,
+  graph: Graph<T>,
+): Promise<boolean> {
+  const result = getAllTopLevelInverseDependencies(
+    inverseDependencies,
+    parentModule,
+    graph,
+    '',
+    new Set(),
```
noahsug: Maybe make these args optional with default values

packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -309,12 +312,71 @@ async function addDependency<T>(
...
+      });
+      return acc;
+    }, new Set());
+}
+
+/**
+ * Given `inverseDependencies`, tracing back inverse dependencies to
+ * see if it only leads back to `parentModule`.
+ */
+async function canSafelyRemoveFromParentModule<T>(
```
noahsug: I think this function is synchronous? In which case we should remove `async`

packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -309,12 +312,71 @@ async function addDependency<T>(
...
+  visited.add(currModule);
+  if (!inverseDependencies.size || currModule === parentModule) {
+    return new Set([currModule]);
+  }
+
+  return Array.from(inverseDependencies)
+    .filter(inverseDep => graph.dependencies.has(inverseDep))
+    .reduce((acc, inverseDep) => {
+      const mod = graph.dependencies.get(inverseDep);
+      if (!mod) {
```
noahsug: won't this always be false due to the `filter` above?

packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -323,9 +385,17 @@ function removeDependency<T>(
...
   module.inverseDependencies.delete(parentModule.path);
 
-  // This module is still used by another modules, so we cannot remove it from
-  // the bundle.
-  if (module.inverseDependencies.size) {
+  // Even if there are modules still using parentModule, we want to ensure
+  // there isn't circular dependency. Thus, we check if it can be safely remove
+  // by tracing back the inverseDependencies.
+  if (
+    module.inverseDependencies.size &&
```
noahsug: Can do the `module.inverseDependencies.size` check in `canSafelyRemoveFromParentModule`

packages/metro/src/DeltaBundler/traverseDependencies.js
cpojer: I assume this is to make Flow happy, which isn't able to keep track of that information.

cpojer:
Thank you so much for this fix. I believe this looks good but do you mind answering @noahsug's questions? I am also curious if there is any measurable performance difference but overall I think it is probably fine – all this data is available in the graph already and no expensive I/O needs to be done for this. If it's less than ~100ms additional time spent for a graph of 10k modules with one change I think we can live with this.

packages/metro/src/DeltaBundler/traverseDependencies.js
raejin: Ahh, nice catch :)

packages/metro/src/DeltaBundler/traverseDependencies.js
raejin: Yes what @cpojer said

packages/metro/src/DeltaBundler/traverseDependencies.js
raejin: I think this may make it clear on what `canSafelyRemoveFromParentModule` function intends to do, as way to signify that if `module.inverseDependencies.size` is non zero then we need additional check for all of the inverse dependencies.

raejin:
@cpojer 
> Thank you so much for this fix. I believe this looks good but do you mind answering @noahsug's questions? I am also curious if there is any measurable performance difference but overall I think it is probably fine – all this data is available in the graph already and no expensive I/O needs to be done for this. If it's less than ~100ms additional time spent for a graph of 10k modules with one change I think we can live with this.

After running the initial solution with our entrypoint which amounts to 8730 modules, there was obvious performance bug with it (~10s for removing the entrypoint).

With more investigation, I implemented a memoized version which helps avoiding unnecessary DFS. With the updated approach, we're looking at 1~2 seconds by removing an entrypoint which has 8730 modules. I also added more tests around various edge cases that I came up with while debugging with our entrypoint.

packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -309,12 +318,107 @@ async function addDependency<T>(
...
+ */
+function canSafelyRemoveFromParentModule<T>(
+  inverseDependencies: Set<string>,
+  parentModule: string,
+  graph: Graph<T>,
+  canBeRemovedSafely: Set<string>,
+  delta: Delta,
+): boolean {
+  const visited = new Set();
+  const result = getAllTopLevelInverseDependencies(
```
noahsug: maybe rename `result` to `inverseDependencies`? Since result isn't actually the result the function returns

packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -309,12 +318,107 @@ async function addDependency<T>(
...
+  if (!result.size) {
+    /**
+     * This happens when parentModule and inverseDependencies has circular dependency,
+     * this will eventaully became an empty set due to `visited` Set being the base case
+     * for the recursive call.
+     */
+    return true;
+  }
+
+  const filterNotDeletedResult = Array.from(result).filter(
```
noahsug: could rename to something like `undeletedInverseDependencies`

packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -309,12 +318,107 @@ async function addDependency<T>(
...
+  /**
+   * We can only mark `visited` set of modules to be safely removable if
+   * 1. We do not have top level module to compare with parentModule,
+   *   this can happen when trying to see if we can safely remove from
+   *   a module that was deleted. This is why we filtered them out with `delta.deleted`
+   * 2. We have one top module and it is parentModule
+   *
+   */
+  const canSafelyRemove =
+    !filterNotDeletedResult.length ||
```
noahsug: I think we can remove this line

raejin: We want to set `canSafelyRemove` when
1. `filterNotDeletedResult` is empty
2. `filterNotDeletedResult` is not empty but only has one item.

I don't think we can remove this 🤔

noahsug: Woops I misread the logic. Definitely do not remove this line 😛

cpojer:
@raejin 
> With more investigation, I implemented a memoized version which helps avoiding unnecessary DFS. With the updated approach, we're looking at 1~2 seconds by removing an entrypoint which has 8730 modules. I also added more tests around various edge cases that I came up with while debugging with our entrypoint.

Just wanted to clarify this is not the usual case for every change, only when removing a module that has 8k+ dependencies of its own, right? Removing a module here and there won't have a big performance impact, is that correct?

raejin:
> @raejin
> 
> > With more investigation, I implemented a memoized version which helps avoiding unnecessary DFS. With the updated approach, we're looking at 1~2 seconds by removing an entrypoint which has 8730 modules. I also added more tests around various edge cases that I came up with while debugging with our entrypoint.
> 
> Just wanted to clarify this is not the usual case for every change, only when removing a module that has 8k+ dependencies of its own, right? Removing a module here and there won't have a big performance impact, is that correct?

Yes that is correct. Even with 8k module the latency is negligible in my opinion.

PR #507, opened 2020-01-09
Add watch option to Server and disable file watchers for runBuild

cdsanchez:
**Summary**

In this PR I am disabling file watching for bundle generation done through `runBuild`. A `watch` option is also added to `Server` in order to enable that (and also because `Server` is used by the React Native CLI instead of `runBuild`).

The goal of this is to solve `ENOSPC` errors related to excessive file watcher use, particularly in CI environments. See: https://github.com/facebook/metro/issues/355 

Context:
I previously attempted to fix this by resolving a regression in the existing `watch` config option: https://github.com/facebook/metro/pull/491
Which was closed in favor of a similar PR:  https://github.com/facebook/metro/pull/497
Which was then re-created in: https://github.com/facebook/metro/pull/503 (and landed)

The above PR was modified to automatically set the `watch` value depending on whether it was built in a CI environment. There was a desire to remove the `watch` config option altogether.

At the end, we concluded that there is still a need for a caller to turn off file watching explicitly, and not only through CI detection.

Therefore, I'm threading it through the code as an option that is passed through programmatically and *not* through the config.

Once/if this lands, I will update the React Native CLI to use this option.

**Test plan**

I used the rn-cli as a test harness of sorts.

Invoked metro via `react-native start`
- Observed that the `watch` flag was true through `console.log`

Invoked metro via `react-native bundle`
- Observed that `watch` was set to false.

Relatively new to this project, so please let me know if I should be taking a look at anything else.

### DIFF
--- a/packages/metro/src/Bundler.js
+++ b/packages/metro/src/Bundler.js
@@ -17,12 +17,16 @@ import type {TransformOptions} from './DeltaBundler/Worker';
 import type {TransformResultWithSource} from './DeltaBundler';
 import type {ConfigT} from 'metro-config/src/configTypes.flow';
 
+export type BundlerOptions = {|
+  +watch?: boolean,
+|};
+
 class Bundler {
   _depGraphPromise: Promise<DependencyGraph>;
   _transformer: Transformer;
 
-  constructor(config: ConfigT) {
-    this._depGraphPromise = DependencyGraph.load(config);
+  constructor(config: ConfigT, options: BundlerOptions = {}) {
+    this._depGraphPromise = DependencyGraph.load(config, options);
 
     this._depGraphPromise
       .then((dependencyGraph: DependencyGraph) => {
--- a/packages/metro/src/Bundler.js
+++ b/packages/metro/src/Bundler.js
@@ -25,7 +25,7 @@ class Bundler {
   _depGraphPromise: Promise<DependencyGraph>;
   _transformer: Transformer;
 
-  constructor(config: ConfigT, options: BundlerOptions = {}) {
+  constructor(config: ConfigT, options?: BundlerOptions) {
     this._depGraphPromise = DependencyGraph.load(config, options);
 
     this._depGraphPromise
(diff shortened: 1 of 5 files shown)


cdsanchez:
@cpojer is this more inline with how we can turn off `watch`?

packages/metro/src/node-haste/DependencyGraph.js
```diff
@@ -93,17 +95,20 @@ class DependencyGraph extends EventEmitter {
...
+    config: ConfigT,
+    options?: {|+watch?: boolean|},
+  ): Promise<DependencyGraph> {
     const initializingMetroLogEntry = log(
       createActionStartEntry('Initializing Metro'),
     );
 
     config.reporter.update({type: 'dep_graph_loading'});
-    const haste = DependencyGraph._createHaste(config);
+    const haste = DependencyGraph._createHaste(config, (options || {}).watch);
```
cpojer: What about `(options && options.watch)`?

packages/metro/src/Bundler.js
```diff
@@ -17,12 +17,16 @@ import type {TransformOptions} from './DeltaBundler/Worker';
 import type {TransformResultWithSource} from './DeltaBundler';
 import type {ConfigT} from 'metro-config/src/configTypes.flow';
 
+export type BundlerOptions = {|
```
cpojer: Could you add `$ReadOnly`?

packages/metro/src/node-haste/DependencyGraph.js
```diff
@@ -73,7 +73,9 @@ class DependencyGraph extends EventEmitter {
     this._createModuleResolver();
   }
 
-  static _createHaste(config: ConfigT): JestHasteMap {
+  static _createHaste(config: ConfigT, watch?: boolean): JestHasteMap {
+    const shouldWatch = watch === undefined ? !ci.isCI : watch;
```
cpojer: Could you inline this below instead of using the variable? It's a simple expression. Also, let's use `watch == null` instead of checking for undefined.

packages/metro/src/Bundler.js
cdsanchez: Good call. I applied that to a couple of other places where I was doing something similar.

PR #508, opened 2020-01-16
fix regex

joeyfenny:
**Summary**

fixes this error: https://github.com/expo/expo-cli/issues/1074

**Test plan**

expo start => begins metro bundler without fatally crashing due to this regex error. Introducted when upgrading to later versions of node.js


cpojer:
cc @brentvatne

I do not think we need to merge this change as we recently published a fix for this exact issue on Windows. I'm going to close this but maybe Brent can chime in if this is a separate issue.

brentvatne:
yeah this should be resolved already. @joeyfenny try deleting your node_modules / lockfiles and running npm install again to get the latest compatible versions

PR #509, opened 2020-01-17

janicduplessis:
**Summary**

**Test plan**

```js
    transformer: {
      getTransformOptions: () => {
        return {
          transform: {
            inlineRequires: true,
          },
        };
      },
    },
```

#### prod, no minify:
size: 10,5mb
execution: 322ms

#### prod, minify:
size: 4.9mb
execution: 240ms

#### prod, no minify:
size: 9,5mb
execution: 308ms

#### prod, minify:
size: 4.6mb
execution: 226ms

TODO:

- Run more tests on iOS 10 / Android (JSC + Hermes)

### DIFF
--- a/packages/metro-react-native-babel-preset/package.json
+++ b/packages/metro-react-native-babel-preset/package.json
@@ -28,6 +28,7 @@
     "@babel/plugin-syntax-export-default-from": "^7.0.0",
     "@babel/plugin-syntax-flow": "^7.2.0",
     "@babel/plugin-transform-arrow-functions": "^7.0.0",
+    "@babel/plugin-transform-async-to-generator": "^7.0.0",
     "@babel/plugin-transform-block-scoping": "^7.0.0",
     "@babel/plugin-transform-classes": "^7.0.0",
     "@babel/plugin-transform-computed-properties": "^7.0.0",
--- a/packages/metro-react-native-babel-preset/package.json
+++ b/packages/metro-react-native-babel-preset/package.json
@@ -18,6 +18,7 @@
   ],
   "license": "MIT",
   "dependencies": {
+    "@babel/preset-modules": "^0.1.2",
     "@babel/plugin-proposal-class-properties": "^7.0.0",
     "@babel/plugin-proposal-export-default-from": "^7.0.0",
     "@babel/plugin-proposal-nullish-coalescing-operator": "^7.0.0",
--- a/packages/metro-react-native-babel-preset/src/configs/main.js
+++ b/packages/metro-react-native-babel-preset/src/configs/main.js
@@ -22,7 +22,6 @@ function isTSXSource(fileName) {
 const defaultPlugins = [
   [require('@babel/plugin-syntax-flow')],
   [require('@babel/plugin-proposal-optional-catch-binding')],
-  [require('@babel/plugin-transform-block-scoping')],
   [
     require('@babel/plugin-proposal-class-properties'),
     // use `this.foo = bar` instead of `this.defineProperty('foo', ...)`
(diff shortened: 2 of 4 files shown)


janicduplessis:
Note that this will probably also require changing from `uglify-es` to `terser` as I'm currently getting an error (`name.definition is not a function`) when building minified builds with no es6 transforms.

janicduplessis:
Depends on some variation of #510 landing.

janicduplessis:
If someone is interested in helping test this I've published the relevant metro packages to git tags.

Using yarn resolutions:

```
"resolutions": {
    "metro-config": "janicduplessis/metro#metro-config-v0.57.0-gitpkg",
    "metro-minify-terser": "janicduplessis/metro#metro-minify-terser-v0.57.0-gitpkg",
    "metro-react-native-babel-preset": "janicduplessis/metro#metro-react-native-babel-preset-v0.57.0-gitpkg"
}
```

developit:
Hi there! This is a great initiative. One thing though - I would *strongly* recommend applying `@babel/preset-modules` if output needs to support iOS 10 and/or 11. The overhead is near-zero and it fixes some pretty subtle but important bugs in JSC.

janicduplessis:
@developit Looks like the minimum iOS version to use it would be iOS 10.3 or 11, right now we need to support 10.0 (I think). It could be possible to use individual plugins instead of the preset.

janicduplessis:
I think we could use https://github.com/babel/preset-modules/blob/master/src/plugins/transform-tagged-template-caching/index.js and remove `transform-template-literals` right? According to `babel-preset-env` data it requires iOS 13 but I assume this is because of the bugs this transform fixes?

janicduplessis:
@developit Any chance the plugins could be published individually? I don't really mind reaching out to the individual files from the preset but it might help getting this merged.

janicduplessis:
@cpojer There is just one issue left to address, currently hermes does not support all syntax needed to remove all the es6 transforms. It is missing support for class, block scope (let/const), shorthand/computed props and there seems to be a bug with generators when using the async to generator transform.

If it is possible to prioritize those with the hermes team that’d be great.

For now we can either wait to land this or have a way to know the target engine is hermes and apply different transforms.

cpojer:
The Hermes team is working on full ES2015 support and that'll be required for us to switch on this flag as well. We can land this and when the Hermes team will roll out new features to production at Facebook we can extend the feature to have an `unstableHermes` or something flag in addition to this one that will disable the transforms gradually as we move to ES2015+ at FB. If you are excited about this, feel free to send a PR on top of this one to add the additional mode (we could turn the current option into an enum instead of a boolean).

janicduplessis:
I think an option that would work well here is engine: `{ [key: ‘ios’ | ‘android’: { ‘jsc’ | ‘hermes’ } }`. So you can do something like `{ ios: ‘jsc’, android: 'hermes' }`

cpojer:
I'm trying to land this at FB but somehow it keeps adding node-dev and then inotify which doesn't run on Mac and then Yarn bails. I can't seem to figure out which new package depends on node-dev. Do you know?

janicduplessis:
Strange, I haven’t had that issue and I am on macOS also. Not sure what node-dev is but it’s not anywhere in the yarn.lock changes.

cpojer:
I think I just got it to work. There was something super weird going on in our complex internal setup. I'll try to land this.

janicduplessis:
You probably want to hold on the minifier change to land first as it doesn’t work with uglify when the flag is on. Although it isn’t that big of a deal since the flag is off by default.

cpojer:
Heads-up: have to revert this because it did break something internally because of the funky dependency stuff. Can you re-send the PR once it is unlanded on master? Sorry for the trouble.

janicduplessis:
Sure, no problem

janicduplessis:
@cpojer #512

PR #510, opened 2020-01-18
Bring back terser minifier

janicduplessis:
**Summary**

This brings back `metro-minify-terser` which was removed because it wasn't used. Now to make sure it is used we'll make it the default.

`terser` is a fork of `uglify-es` which is not maintained. Updating this is required to remove es6 transforms since `uglify-es` currently crashes on in my app when not transforming es6. It would probably be possible to track down the bug but at this point it's probably easier and better long term to migrate to `terser`. It is now the default minifier for webpack and seem pretty stable.

**Test plan**

Tested in my app to make sure minifying works when disabling es6 transforms.

### DIFF
--- a/packages/metro-minify-terser/.npmignore
+++ b/packages/metro-minify-terser/.npmignore
@@ -1,0 +1,5 @@
+**/__mocks__/**
+**/__tests__/**
+build
+src.real
+yarn.lock
--- a/packages/metro-minify-terser/__tests__/minify-test.js
+++ b/packages/metro-minify-terser/__tests__/minify-test.js
@@ -1,0 +1,99 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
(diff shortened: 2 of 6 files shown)


janicduplessis:
@cpojer Makes sense, updated this PR to only add the plugin

chrisbobbe:
> Can you split this into two PRs:
> 
> * Add back all the code that was previously removed (+ upgrade terser and whatnot)
> * One PR that turns it on as the new default?
> 
> We can't switch this at FB in a single PR and I can land the former and have to make changes to the latter.

@janicduplessis, gentle bump on that second PR; I think that change hasn't been made yet? 🙂 See https://github.com/facebook/metro/blob/v0.66.2/packages/metro-config/src/defaults/defaults.js#L61.

janicduplessis:
I don't plan on working on it, feel free to open a PR to change the default :)

PR #511, opened 2020-01-18
add optional dependency support

connectdotz:
**Summary**

This PR added support for the common optional dependency pattern: enclose require() within try/catch block, which is already supported by webpack. 

While the lack of optional dependency support is not new, due to [lean-core](https://github.com/facebook/react-native/issues/23313) initiative, it has become obvious that we should just address this to provide better customer-experience. See the excellent discussion in react-native-community/discussions-and-proposals#120.

**Change Outline**
The changes are mainly in 3 areas: 
1. visit try/catch block and mark optional dependencies (`collectDependencies.js`)
1. during dependency resolve, if the optional dependency fails to resolve, just ignore it. (`traverseDependencies.js`)
1. add a config (`optionalDependency`, defaults to true) to disable and customize (exclude dependency from being optional) (`metro-config`)

The rest are just tunneling through the new config option, added/modified tests.

**Test plan**

tested the new create react-native app with `react-native-vector-icons`  and optional `@react-native-community/toolbar-android` (in try/catch block from `react-native-vector-icons` ) :
- without the optional dependency support, metro complained `@react-native-community/toolbar-android` not found and abort
- with the optional dependency, the app starts up without any problem.

**Discussion**
- if we found `import()` in the try block, should we also consider it "optional"? The common pattern is to use `require()` but one would think `import()` should be just the same, no? This PR considered any dependency found in the try block "optional", but that can be changed easily.
- I put the new config `optionalDependency` in the config.resolver, not sure if it is best to be in resolver or transformer?
- should we add some kind of warning message for omitted optional dependency? verbose flag?

### DIFF
--- a/packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
+++ b/packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
@@ -870,3 +870,113 @@ describe('reorderGraph', () => {
     ]);
   });
 });
+
+describe('optional dependencies', () => {
+  let localGraph;
+  let localOptions;
+  const getAllDependencies = () => {
+    const all = new Set();
+    mockedDependencyTree.forEach(deps => {
+      deps.forEach(r => all.add(r.name));
+    });
+    return all;
+  };
+  const assertResults = (dependencies, expectedMissing) => {
+    let count = 0;
+    const allDependency = getAllDependencies();
+    allDependency.forEach(m => {
+      const data = dependencies.get(`/${m}`);
+      if (expectedMissing.includes(m)) {
+        expect(data).toBeUndefined();
+      } else {
+        expect(data).not.toBeUndefined();
+      }
+      count += 1;
+    });
+    expect(count).toBeGreaterThan(0);
+    expect(count).toBe(allDependency.size);
+  };
+
+  beforeEach(() => {
+    mockedDependencies = new Set();
+    mockedDependencyTree = new Map();
+
+    entryModule = Actions.createFile('/bundle-o');
+    entryModule = Actions.createFile('/optional-c');
+
+    Actions.addDependency('/bundle-o', '/regular-a');
+    Actions.addDependency('/bundle-o', '/optional-b');
+    Actions.addDependency('/bundle-o', '/optional-c');
+    Actions.addDependency('/optional-c', '/optional-d');
+    Actions.addDependency('/optional-c', '/regular-e');
+
+    Actions.deleteFile('/optional-b');
+
+    localOptions = {
+      ...options,
+      transform(path) {
+        const result = options.transform.apply(this, arguments);
+        result.dependencies.forEach(dep => {
+          if (dep.name.includes('optional-')) {
+            dep.data.isOptional = true;
+          }
+        });
+        return result;
+      },
+    };
+    localGraph = {
+      dependencies: new Map(),
+      entryPoints: ['/bundle-o'],
+    };
+  });
+  describe('when allowOptionalDependencies config is true', () => {
+    beforeEach(() => {
+      localOptions.allowOptionalDependencies = true;
+    });
+    it('will ignore only the failed optional dependency', async () => {
+      const result = await initialTraverseDependencies(
+        localGraph,
+        localOptions,
+      );
+
+      const dependencies = result.added;
+
+      // Only `optional-b` is actually missing,
+      // the other optional modules `optional-c` and `optional-d` will
+      // still be included.
+      assertResults(dependencies, ['optional-b']);
+    });
+  });
+  describe('when allowOptionalDependencies config is false', () => {
+    beforeEach(() => {
+      localOptions.allowOptionalDependencies = false;
+    });
+    it('will not ignore failed optional dependency', async () => {
+      await expect(
+        initialTraverseDependencies(localGraph, localOptions),
+      ).rejects.toThrow('Dependency not found: /optional-b->optional-b');
+    });
+  });
+  describe('when allowOptionalDependencies with exclude', () => {
+    it('will not ignore the excluded optional dependency', async () => {
+      localOptions.allowOptionalDependencies = {exclude: ['optional-b']};
+      await expect(
+        initialTraverseDependencies(localGraph, localOptions),
+      ).rejects.toThrow('Dependency not found: /optional-b->optional-b');
+    });
+    it('but will still ignore the non-excluded optional dependency', async () => {
+      localOptions.allowOptionalDependencies = {exclude: ['optional-d']};
+      const result = await initialTraverseDependencies(
+        localGraph,
+        localOptions,
+      );
+
+      const dependencies = result.added;
+
+      // Only `optional-b` is actually missing,
+      // the other optional modules `optional-c` and `optional-d` will
+      // still be included.
+      assertResults(dependencies, ['optional-b']);
+    });
+  });
+});
--- a/packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
+++ b/packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
@@ -897,86 +897,55 @@ describe('optional dependencies', () => {
     expect(count).toBe(allDependency.size);
   };
 
+  const mockTransform = (notOptional?: string[]) => {
+    return function(path) {
+      const result = options.transform.apply(this, arguments);
+      result.dependencies.forEach(dep => {
+        if (notOptional && notOptional.includes(dep.name)) {
+          dep.data.isOptional = false;
+        } else {
+          dep.data.isOptional = dep.name.includes('optional-');
+        }
+      });
+      return result;
+    };
+  };
+
   beforeEach(() => {
     mockedDependencies = new Set();
     mockedDependencyTree = new Map();
 
     entryModule = Actions.createFile('/bundle-o');
-    entryModule = Actions.createFile('/optional-c');
 
     Actions.addDependency('/bundle-o', '/regular-a');
     Actions.addDependency('/bundle-o', '/optional-b');
-    Actions.addDependency('/bundle-o', '/optional-c');
-    Actions.addDependency('/optional-c', '/optional-d');
-    Actions.addDependency('/optional-c', '/regular-e');
 
     Actions.deleteFile('/optional-b');
 
-    localOptions = {
-      ...options,
-      transform(path) {
-        const result = options.transform.apply(this, arguments);
-        result.dependencies.forEach(dep => {
-          if (dep.name.includes('optional-')) {
-            dep.data.isOptional = true;
-          }
-        });
-        return result;
-      },
-    };
     localGraph = {
       dependencies: new Map(),
       entryPoints: ['/bundle-o'],
     };
   });
-  describe('when allowOptionalDependencies config is true', () => {
-    beforeEach(() => {
-      localOptions.allowOptionalDependencies = true;
-    });
-    it('will ignore only the failed optional dependency', async () => {
-      const result = await initialTraverseDependencies(
-        localGraph,
-        localOptions,
-      );
 
-      const dependencies = result.added;
+  it('missing optional dependency will be skipped', async () => {
+    localOptions = {
+      ...options,
+      transform: mockTransform(),
+    };
 
-      // Only `optional-b` is actually missing,
-      // the other optional modules `optional-c` and `optional-d` will
-      // still be included.
-      assertResults(dependencies, ['optional-b']);
-    });
+    const result = await initialTraverseDependencies(localGraph, localOptions);
+
+    const dependencies = result.added;
+    assertResults(dependencies, ['optional-b']);
   });
-  describe('when allowOptionalDependencies config is false', () => {
-    beforeEach(() => {
-      localOptions.allowOptionalDependencies = false;
-    });
-    it('will not ignore failed optional dependency', async () => {
-      await expect(
-        initialTraverseDependencies(localGraph, localOptions),
-      ).rejects.toThrow('Dependency not found: /optional-b->optional-b');
-    });
-  });
-  describe('when allowOptionalDependencies with exclude', () => {
-    it('will not ignore the excluded optional dependency', async () => {
-      localOptions.allowOptionalDependencies = {exclude: ['optional-b']};
-      await expect(
-        initialTraverseDependencies(localGraph, localOptions),
-      ).rejects.toThrow('Dependency not found: /optional-b->optional-b');
-    });
-    it('but will still ignore the non-excluded optional dependency', async () => {
-      localOptions.allowOptionalDependencies = {exclude: ['optional-d']};
-      const result = await initialTraverseDependencies(
-        localGraph,
-        localOptions,
-      );
-
-      const dependencies = result.added;
-
-      // Only `optional-b` is actually missing,
-      // the other optional modules `optional-c` and `optional-d` will
-      // still be included.
-      assertResults(dependencies, ['optional-b']);
-    });
+  it('missing non-optional dependency will throw', async () => {
+    localOptions = {
+      ...options,
+      transform: mockTransform(['optional-b']),
+    };
+    await expect(
+      initialTraverseDependencies(localGraph, localOptions),
+    ).rejects.toThrow();
   });
 });
--- a/packages/metro/src/DeltaBundler/traverseDependencies.js
+++ b/packages/metro/src/DeltaBundler/traverseDependencies.js
@@ -18,7 +18,6 @@ import type {
   Module,
   Options,
   TransformResultDependency,
-  AllowOptionalDependencies,
 } from './types.flow';
 
 type Result<T> = {
@@ -47,7 +46,6 @@ type InternalOptions<T> = $ReadOnly<{|
   resolve: $PropertyType<Options<T>, 'resolve'>,
   transform: $PropertyType<Options<T>, 'transform'>,
   shallow: boolean,
-  allowOptionalDependencies: AllowOptionalDependencies,
 |}>;
 
 function getInternalOptions<T>({
@@ -55,7 +53,6 @@ function getInternalOptions<T>({
   resolve,
   onProgress,
   experimentalImportBundleSupport,
-  allowOptionalDependencies,
   shallow,
 }: Options<T>): InternalOptions<T> {
   let numProcessed = 0;
@@ -63,7 +60,6 @@ function getInternalOptions<T>({
 
   return {
     experimentalImportBundleSupport,
-    allowOptionalDependencies,
     transform,
     resolve,
     onDependencyAdd: () => onProgress && onProgress(numProcessed, ++total),
@@ -475,12 +471,6 @@ function resolveDependencies<T>(
   dependencies: $ReadOnlyArray<TransformResultDependency>,
   options: InternalOptions<T>,
 ): Map<string, Dependency> {
-  const {allowOptionalDependencies} = options;
-  const required = (path: string) =>
-    allowOptionalDependencies === false ||
-    (Array.isArray(allowOptionalDependencies.exclude) &&
-      allowOptionalDependencies.exclude.includes(path));
-
   const resolve = (parentPath: string, result: TransformResultDependency) => {
     const relativePath = result.name;
     try {
@@ -492,8 +482,7 @@ function resolveDependencies<T>(
         },
       ];
     } catch (e) {
-      // If this is an optional dependency, ignore the resolution error
-      if (result.data.isOptional === true && !required(relativePath)) {
+      if (result.data.isOptional === true) {
         // should we output some warning? verbose mode?
         return undefined;
       }
--- a/packages/metro/src/DeltaBundler/traverseDependencies.js
+++ b/packages/metro/src/DeltaBundler/traverseDependencies.js
@@ -18,6 +18,7 @@ import type {
   Module,
   Options,
   TransformResultDependency,
+  AllowOptionalDependencies,
 } from './types.flow';
 
 type Result<T> = {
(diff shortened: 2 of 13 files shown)


packages/metro/src/lib/transformHelpers.js
```diff
@@ -71,6 +71,7 @@ async function calcTransformerOptions(
       onProgress: null,
       experimentalImportBundleSupport:
         config.transformer.experimentalImportBundleSupport,
+      optionalDependency: config.resolver.optionalDependency,
```
cpojer: Can we call this `allowOptionalDependencies`?

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -225,6 +258,8 @@ function processImportCall(
   }
 
   const dep = getDependency(state, name, options);
+  if (dep.data.isOptional) {
+  }
```
cpojer: Oops

cpojer:
It's a very busy time for me at Facebook right now so I don't have enough time to thoroughly review. I left some initial comments.

We also cannot make this a default in Metro, at least not in the same diff. Could you make it so the changes do not affect how Metro works in this PR for now?

packages/metro/src/ModuleGraph/worker/collectDependencies.js
connectdotz: 😊 that was sloppy, sorry...

packages/metro/src/lib/transformHelpers.js
connectdotz: sure

connectdotz:
> We also cannot make this a default in Metro, at least not in the same diff. 

I agree, it will be safer this way, the `allowOptionalDependencies` config now defaults to `false` so should not affect the current Metro's behavior.

> It's a very busy time for me at Facebook right now so I don't have enough time to thoroughly review.

I understand you have other important things to take care of, and thanks to help moving forward this issue during the busy time... feel free to suggest others to review if that would make it easier (?).

drew-gross:
# [Codecov](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=h1) Report
> Merging [#511](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=desc) into [master](https://codecov.io/gh/facebook/metro/commit/1a38e0097eedf61fb01780a215fe446336872ef0&el=desc) will **increase** coverage by `0.07%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #511      +/-   ##
==========================================
+ Coverage   83.98%   84.06%   +0.07%     
==========================================
  Files         175      176       +1     
  Lines        5865     5923      +58     
  Branches      973      987      +14     
==========================================
+ Hits         4926     4979      +53     
- Misses        827      832       +5     
  Partials      112      112              
```


| [Impacted Files](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [packages/metro-minify-terser/src/minifier.js](https://codecov.io/gh/facebook/metro/pull/511/diff?src=pr&el=tree#diff-cGFja2FnZXMvbWV0cm8tbWluaWZ5LXRlcnNlci9zcmMvbWluaWZpZXIuanM=) | `84.61% <0.00%> (ø)` | |
| [packages/metro-minify-terser/src/index.js](https://codecov.io/gh/facebook/metro/pull/511/diff?src=pr&el=tree#diff-cGFja2FnZXMvbWV0cm8tbWluaWZ5LXRlcnNlci9zcmMvaW5kZXguanM=) | `100.00% <0.00%> (ø)` | |

------

[Continue to review full report at Codecov](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=footer). Last update [1a38e00...1a42aba](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

ma-shop:
is there a way to install this on a current react-native project?

connectdotz:
>  is there a way to install this on a current react-native project?

not easily... react-native is behind metro version this PR is based on, once this PR is merged we still need to wait for react-native to adopt...

connectdotz:
this PR has been sitting for a while, @cpojer is there any concrete concern I can address so we can move this PR forward?

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -138,6 +140,42 @@ function collectDependencies(
   };
 
   const visitor = {
+    TryStatement(path: Path, state: State) {
+      path.get('block').traverse({
+        CallExpression(p) {
+          const getOptionalDependency = (): string | null => {
```
cpojer: Can you confirm this only evaluates one level deep in try-blocks and it will not work within nested-if statements or function bodies? I think this makes a lot of sense as it reduces complexity somewhat.

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
```diff
@@ -349,6 +349,42 @@ it('collects imports', () => {
   ]);
 });
 
+describe('optional dependency', () => {
+  it('only capture require(xxx) within try block', () => {
```
cpojer:
```suggestion
  it('only captures require(...) within a try block', () => {
```

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
```diff
@@ -349,6 +349,42 @@ it('collects imports', () => {
   ]);
 });
 
+describe('optional dependency', () => {
```
cpojer:
```suggestion
describe('optional dependencies', () => {
```

cpojer: Can you expand the test cases here to:
* Verify that nested if statements and function bodies in a try-catch block do not count dependencies as optional.
* Verify that the behavior of this extraction is the same as it is on master now with this option disabled.

packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -471,18 +475,47 @@ function resolveDependencies<T>(
...
+        // should we output some warning? verbose mode?
+        return undefined;
+      }
+      throw e;
+    }
+  };
 
-      return [relativePath, dependency];
-    }),
+  // const map = new Map(
```
cpojer: Remove this please

packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -471,18 +475,47 @@ function resolveDependencies<T>(
...
-    dependencies.map((result: TransformResultDependency) => {
-      const relativePath = result.name;
-
-      const dependency = {
-        absolutePath: options.resolve(parentPath, result.name),
-        data: result,
-      };
+  const {allowOptionalDependencies} = options;
+
+  const _resolve = (parentPath: string, result: TransformResultDependency) => {
```
cpojer:
```suggestion
  const resolve = (parentPath: string, result: TransformResultDependency) => {
```

packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -471,18 +475,47 @@ function resolveDependencies<T>(
...
+      throw e;
+    }
+  };
 
-      return [relativePath, dependency];
-    }),
+  // const map = new Map(
+  const resolved = dependencies.reduce(
+    (list: Array<[string, Dependency]>, result: TransformResultDependency) => {
+      const resolvedPath = _resolve(parentPath, result);
```
cpojer:
```suggestion
      const resolvedPath = resolve(parentPath, result);
```

packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -471,18 +475,47 @@ function resolveDependencies<T>(
...
+    try {
+      return [
+        relativePath,
+        {
+          absolutePath: options.resolve(parentPath, relativePath),
+          data: result,
+        },
+      ];
+    } catch (e) {
+      // if this is an optional dependency, ignore the resolve error
```
cpojer:
```suggestion
      // If this is an optional dependency, ignore the resolution error.
```

packages/metro/src/DeltaBundler/traverseDependencies.js
```diff
@@ -471,18 +475,47 @@ function resolveDependencies<T>(
...
+      if (
+        result.data.isOptional === true &&
+        allowOptionalDependencies !== false &&
+        !(
+          Array.isArray(allowOptionalDependencies.exclude) &&
+          allowOptionalDependencies.exclude.includes(relativePath)
```
cpojer: Should we instead invert this and require users to explicitly include the optional dependencies they want to use? That way we ensure users will update a configuration every time they use an optional dependency and that they will never end up with a broken bundle that is missing a dependency accidentally.

packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
```diff
@@ -870,3 +870,117 @@ describe('reorderGraph', () => {
     ]);
   });
 });
+
+describe('optional dependency', () => {
```
cpojer:
```suggestion
describe('optional dependencies', () => {
```

packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
```diff
@@ -870,3 +870,117 @@ describe('reorderGraph', () => {
...
+
+  beforeEach(() => {
+    // create our own data set
+    mockedDependencies = new Set();
+    mockedDependencyTree = new Map();
+
+    entryModule = Actions.createFile('/bundle-o');
+    // entryModule = Actions.createFile('/regular-a');
+    entryModule = Actions.createFile('/optional-c');
+    // entryModule = Actions.createFile('/regular-e');
```
cpojer: Can you clean up the three lines above that are unused?

packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
```diff
@@ -870,3 +870,117 @@ describe('reorderGraph', () => {
...
+        expect(data).not.toBeUndefined();
+      }
+      count += 1;
+    });
+    expect(count).toBeGreaterThan(0);
+    expect(count).toBe(allDependency.size);
+  };
+
+  beforeEach(() => {
+    // create our own data set
```
cpojer: Can you remove this comment?

packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
```diff
@@ -870,3 +870,117 @@ describe('reorderGraph', () => {
...
+    entryModule = Actions.createFile('/optional-c');
+    // entryModule = Actions.createFile('/regular-e');
+
+    Actions.addDependency('/bundle-o', '/regular-a');
+    Actions.addDependency('/bundle-o', '/optional-b');
+    Actions.addDependency('/bundle-o', '/optional-c');
+    Actions.addDependency('/optional-c', '/optional-d');
+    Actions.addDependency('/optional-c', '/regular-e');
+
+    // remove optional-b
```
cpojer: Remove this comment pls

packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
```diff
@@ -870,3 +870,117 @@ describe('reorderGraph', () => {
...
+    });
+    it('will ignore only the failed optional dependency', async () => {
+      const result = await initialTraverseDependencies(
+        localGraph,
+        localOptions,
+      );
+
+      const dependencies = result.added;
+
+      // only optional-b is actually missing,
```
cpojer:
```suggestion
      // Only optional-b is actually missing,
```

packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
```diff
@@ -870,3 +870,117 @@ describe('reorderGraph', () => {
...
+      // the other optional modules: optional-c and optional-d will
+      // still be included
```
cpojer:
```suggestion
      // still be included.
```

packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
```diff
@@ -870,3 +870,117 @@ describe('reorderGraph', () => {
...
+    it('will ignore only the failed optional dependency', async () => {
+      const result = await initialTraverseDependencies(
+        localGraph,
+        localOptions,
+      );
+
+      const dependencies = result.added;
+
+      // only optional-b is actually missing,
+      // the other optional modules: optional-c and optional-d will
```
cpojer:
```suggestion
      // the other optional modules `optional-c` and `optional-d` will
```

packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
```diff
@@ -870,3 +870,117 @@ describe('reorderGraph', () => {
...
+      // still be included
```
cpojer:
```suggestion
      // Only optional-b is actually missing,
      // the other optional modules `optional-c` and `optional-d` will
      // still be included.
```

packages/metro/src/DeltaBundler/traverseDependencies.js
connectdotz: maybe I didn't read your comment correctly... but that is exactly "allowOptionalDependencies.exclude" does: to turn optional dependency to be non-optional, if one chooses to.

But it's not required, because many 3rd-party library users might not know what is considered optional within those libraries... for those use cases, metro should just work (with standard syntax analysis - babel parser - taking the hints from code) without requiring extra user config, IMHO.

did I miss your point?

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
connectdotz: excellent questions...

> Verify that nested if statements and function bodies in a try-catch block do not count dependencies as optional.

How do we know the import/require within the if-statement or function bodies are not intended to be optional if they are within the **lexical** try-block ?  Note dependency added by a function declared outside of try-block, even if it is invoked within a try-block, will not be considered optional. If we are worrying about that developers accidentally included an import within the try block that does not mean to be optional, maybe we should just giving warning messages for all the unresolved optional packages?

> Verify that the behavior of this extraction is the same as it is on master now with this option disabled.

even though I have verified with tests, there is always the possibility for edge cases we haven't thought about... the only sure way is probably to turn off the whole `tryStatement` processing when the flag is off, I will do that...

packages/metro/src/ModuleGraph/worker/collectDependencies.js
connectdotz: "one-level deep"? in terms of runtime call-hierarchy or the code-structure? 

- code-structure: every statement within the **lexical** try-block will be parsed (by babel parser), so it is not shallow. 
- call-hierarchy: babel parser is a lexical analyzer so it does not evaluate the statements nor following function calls, so you can say it is shallow in terms of call-stack?

example:

```
function outside() {
  import('not-optional");
}
try {
  import('this-is-optional');
  if(true){
    if(false) {
        import('this-is-also-optional');
    }
  }
  outside();
} catch (e) {}
```

The example above is merely for illustration... usually, an optional dependency is enclosed in a much tighter try-block since it needs very specific "catch" logic to handle the missing module situation ...

drew-gross:
# [Codecov](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=h1) Report
> Merging [#511](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=desc) into [master](https://codecov.io/gh/facebook/metro/commit/d472f3964301b01fb64a254a61413b0f46566968&el=desc) will **increase** coverage by `0.06%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #511      +/-   ##
==========================================
+ Coverage   84.00%   84.07%   +0.06%     
==========================================
  Files         176      176              
  Lines        5897     5927      +30     
  Branches      981      989       +8     
==========================================
+ Hits         4954     4983      +29     
- Misses        831      832       +1     
  Partials      112      112              
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=footer). Last update [d472f39...a4261ba](https://codecov.io/gh/facebook/metro/pull/511?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

packages/metro/src/ModuleGraph/worker/collectDependencies.js
cpojer: Can we instead change it to check only for `require` calls directly within one-level of a try-block?

Example:
```
// Valid
let a;
try {
  a = require('module');
} catch {}

// Invalid

try {
  if ('maybeInclude') {
    require('module');
  }
} catch {}
```

You can achieve this by directly iterating over the items in `block`  instead of using `traverse`.

I think we can reduce complexity here – I don't see any case where a try block for an optional dependency should be more complex than assigning a variable.

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
cpojer: For the first question: see my answer above about changing the extractor.

packages/metro/src/ModuleGraph/worker/collectDependencies.js
connectdotz: > Can we instead change it to check only for require calls directly within one-level of a try-block?... I think we can reduce complexity here

we could certainly constraint to one-level, but what does 'complexity' mean?

```
try {
  // hundreds of line of code here
  ...
  a = require('module');
  // another hundred of line of code here
  ...
} catch {}
```

The example above is one-level, but is it considered complex? Maybe we should also consider limiting the try-block statement number to be 1 to make it not-complex? But a single statement can also be "complex", so goes the rabbit hole...

On the other hand, if  the goal of trying to "reduce complexity" is to make sure developers won't accidentally put the non-optional dependency in the try-block, then instead of using the level, or any other kinds of complexity measurement to **guess**, I wonder would it be more direct and fool-proof to use an explicit token, such as a comment to safeguard such intention:

```
try {
  // metro: optional-dependency
  a = require('this-is-optional');

  b = require('this-is-NOT-optional');
} catch {}
```

So all the existing dependency in try-block, even when the `allowOptionalDependencies` is turned on, will not be optional, unless developers explicitly added this comment directly above the call... Does that address your concern?

cpojer: I am fine if we limit it to variable assignments one-level deep:

```
try {
  a = require('someModule');
  b = require('someOtherModule');
} catch {}
```

Invalid:
```
try {
  a = require('someModule');
  runSomeCode();
  b = require('someOtherModule');
} catch {}
```

Any other code can be moved out of the try block and handled afterwards. That way we ensure optional dependencies can only be used via a very single pattern – it's an escape hatch with limited API support.

connectdotz: ok, I refactor the code to adopt the bottom-up approach and piggyback on the existing import/require expression processing to completely removed the TryStatement handler (top-down). The optional dependency check is now simply to see if the TryStatement is in its "immediate" parent chain after the import/require expression is vetted in the existing CallExpression handlers. 

However, I didn't implement any other sibling level constraint from the "invalid" example above. Sibling level constraint will make otherwise simple and linear parent-child traversal a lot more complicated, because it made the expression condition depends on all the other expressions/statements within the block... it didn't add any feature as far as I can see and certainly didn't further reduce the complexity in the extraction code, our ultimate goal...

Let me know if this makes sense and I will do the final rebase/merge.

cpojer: This sounds great actually. Can you rebase so I can take another look before importing this to FB?

connectdotz:
hmmm... after merged with the master, ci failed due to node version incompatibility, which is unrelated to this PR...
> error graphql@15.0.0-rc.2: The engine "node" is incompatible with this module. Expected version ">= 10.x". Got "8.17.0"
> error Found incompatible module.

The master built has been failing at least since the last commit as well: [a2738f3
](https://github.com/facebook/metro/commit/a2738f3f47386bf12a033d399c9b7b6226b59236)

thoughts?

connectdotz:
done rebase.

connectdotz:
sorry, added one more change to consolidate all optional-dependency validation logic into the `collectDependencies` to simplify resolve handling in `traverseDependencies` and reduce change scope a bit. 

I think this concluded my change, please let me know if you see anything else...

JoelMarcey:
@drew-gross Is it possible for you to go to your GitHub settings and revoke the OAuth token that is making these Codecov comments?

PR #512, opened 2020-01-23

janicduplessis:
Reland of #509

**Summary**

**Test plan**

```js
    transformer: {
      getTransformOptions: () => {
        return {
          transform: {
            inlineRequires: true,
          },
        };
      },
    },
```

#### prod, no minify:
size: 10,5mb
execution: 322ms

#### prod, minify:
size: 4.9mb
execution: 240ms

#### prod, no minify:
size: 9,5mb
execution: 308ms

#### prod, minify:
size: 4.6mb
execution: 226ms

### DIFF
--- a/packages/metro-react-native-babel-preset/package.json
+++ b/packages/metro-react-native-babel-preset/package.json
@@ -28,6 +28,7 @@
     "@babel/plugin-syntax-export-default-from": "^7.0.0",
     "@babel/plugin-syntax-flow": "^7.2.0",
     "@babel/plugin-transform-arrow-functions": "^7.0.0",
+    "@babel/plugin-transform-async-to-generator": "^7.0.0",
     "@babel/plugin-transform-block-scoping": "^7.0.0",
     "@babel/plugin-transform-classes": "^7.0.0",
     "@babel/plugin-transform-computed-properties": "^7.0.0",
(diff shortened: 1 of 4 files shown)


cpojer:
I'll get to this again later this week. Sorry for the delay.

rickhanlonii:
Heads up: this needed to be reverted because the `react-class-display-name` babel transform started failing on component like:

```
import {View} from 'react-native';
import * as React from 'react';

export default class extends React.Component {
  render(): React.Node {
    return <View />;
  }
}
```

The error was "Cannot read property 'name' of null".

Could we re-land this without any changes to the ordering of the plugins or any other semantics?

janicduplessis:
@rickhanlonii Thanks for the repro. I tried it but I am not getting the error. What is `react-class-display-name` transform? I don't see it in the preset and found https://github.com/researchgate/babel-plugin-react-class-display-name on npm but it is for babel 5.

motiz88:
That's actually a FB-internal plugin. However, I'm afraid the specific failure mode is just a symptom of deeper Babel weirdness triggered by this change somehow.

For instance, this plugin had a `ClassDeclaration` visitor but it was suddenly being called with `ClassExpression` nodes (which have a billable `id`, hence the error). When we had worked around that, we started seeing other exceptions in Babel plugins. (Something with `super` in async methods? I'd need to dig up the exact error)

The most viable way forward IMO is to ensure this has the _exact same_ transform behaviour when the flag is off, as Rick mentioned above.

janicduplessis:
Alright, will change the code structure a bit so it can include plugins in the same order as before.

PR #514, opened 2020-01-28
Set BABEL_ENV in output/bundle.js

jjd314:
First, let me apologize for the paucity of this pull request. I made this change as a patch on version 0.56.3, which is what got automatically installed for me in my react-native project. I thought I would submit it as this pull request just so someone who might know what they are doing on this project could see it, since my googling showed others who were hitting the same problem. I do not have the capacity right now to really become familiar enough with the code to be sure this is the right fix for the problem, nor to really test it thoroughly. If someone on the project wants it, great; if not and you ignore it, that's great, too. Again, my apologies.

When using metro from 'yarn android --variant=release' or 'yarn ios --variant=release', BABEL_ENV needs to be set to 'production' so that babel can find the right configuration in babel.config.js. I just copied the code from the other places in metro where this is done.

The only test I made was successfully running 'yarn android --variant=release' on my project, and then only on version 0.56.3.

### DIFF
--- a/packages/metro/src/shared/output/bundle.js
+++ b/packages/metro/src/shared/output/bundle.js
@@ -28,8 +28,8 @@ function buildBundle(
 }> {
   const OLD_BABEL_ENV = process.env.BABEL_ENV;
   process.env.BABEL_ENV = requestOptions.dev
-                        ? "development"
-                        : (process.env.BABEL_ENV || "production");
+                        ? 'development'
+                        : (process.env.BABEL_ENV || 'production');
   try {
     return packagerClient.build({
       ...Server.DEFAULT_BUNDLE_OPTIONS,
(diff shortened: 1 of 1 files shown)


packages/metro/src/shared/output/bundle.js
```diff
@@ -26,11 +26,26 @@ function buildBundle(
...
+  process.env.BABEL_ENV = requestOptions.dev
+    ? 'development'
+    : process.env.BABEL_ENV || 'production';
+  return packagerClient
+    .build({
+      ...Server.DEFAULT_BUNDLE_OPTIONS,
+      ...requestOptions,
+      bundleType: 'bundle',
+    })
+    .then(
```
cpojer: Could you change this to use `finally` instead? Then you don't need to worry about the return values etc.

(If CI still uses node 8, feel free to bump that to 10+)

cpojer:
Closing because of no activity for a month.

PR #515, opened 2020-01-31

janicduplessis:
Reland of #512

Make sure plugins ordering stays exactly the same as before.

**Summary**

**Test plan**

```js
    transformer: {
      getTransformOptions: () => {
        return {
          transform: {
            inlineRequires: true,
          },
        };
      },
    },
```

#### prod, no minify:
size: 10,5mb
execution: 322ms

#### prod, minify:
size: 4.9mb
execution: 240ms

#### prod, no minify:
size: 9,5mb
execution: 308ms

#### prod, minify:
size: 4.6mb
execution: 226ms

### DIFF
--- a/packages/metro-react-native-babel-preset/package.json
+++ b/packages/metro-react-native-babel-preset/package.json
@@ -18,6 +18,7 @@
   ],
   "license": "MIT",
   "dependencies": {
+    "@babel/preset-modules": "^0.1.2",
     "@babel/plugin-proposal-class-properties": "^7.0.0",
     "@babel/plugin-proposal-export-default-from": "^7.0.0",
     "@babel/plugin-proposal-nullish-coalescing-operator": "^7.0.0",
@@ -28,6 +29,7 @@
     "@babel/plugin-syntax-export-default-from": "^7.0.0",
     "@babel/plugin-syntax-flow": "^7.2.0",
     "@babel/plugin-transform-arrow-functions": "^7.0.0",
+    "@babel/plugin-transform-async-to-generator": "^7.0.0",
     "@babel/plugin-transform-block-scoping": "^7.0.0",
     "@babel/plugin-transform-classes": "^7.0.0",
     "@babel/plugin-transform-computed-properties": "^7.0.0",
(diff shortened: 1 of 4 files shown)


janicduplessis:
cc @cpojer @rickhanlonii @motiz88

rickhanlonii (approved):
LGTM, let's try this again

rickhanlonii:
This has the same issue internally, a babel plugin is throwing. We'll have to take a close look in order to fix and land this.

janicduplessis:
@rickhanlonii Is is possible that some project has the unstable_disableES6Transforms flag set? Otherwise I'm pretty confident this keeps the same plugin order.

cpojer:
@moti any thoughts about what we should do with this one?

motiz88:
At this point I'd add a new value to the `unstable_transformProfile` enum rather than an entirely new flag. This enum value would be used only by OSS.

It also makes sense to find bigger OSS codebases to try these changes on, since the type of issues we've run into internally can reflect real blockers but we probably can't keep landing and reverting this at FB :) keeping it as an experimental unstable flag might be sufficient though.

janicduplessis:
@motiz88 I agree, the PR was patched on top of an existing option but `unstable_transformProfile` seems like a good fit. One issue I see though is it would be nice to support specifying this per platform, somthing like: `unstable_transformProfile: {android: 'hermes-canary', ios: 'jsc-ios10'}`.

janicduplessis:
We might also want to organize this code differently if we're going to support more js engine profiles as I feel the current structure might get pretty nasty quickly.

newyankeecodeshop:
@janicduplessis @motiz88 Is there anything folks in the community can do to help move this along? It'd be nice to get this in now that modern JavaScriptCore is available on Android and iOS.

NE-SmallTown:
Any update on this? I think it will make sense if we could land this, and make this as the first step to implement the `preset-env`(targets) in Metro

janicduplessis:
I don't have time to work on this currently, if anyone wants to take on the project. I would suggest discussing it more here with @motiz88 and myself (see https://github.com/facebook/metro/pull/515#issuecomment-605374891).

janicduplessis:
I know hermes supports a lot more modern syntax now, I'd be interested to know if facebook leverages those yet internally.

robhogan:
Closing as this mechanism has been replaced by `unstable_transformProfile` and the preset itself (which takes transform profile as an input) has moved to `@react-native/babel-preset`

PR #523, opened 2020-02-25
Move pretty-format to dev dependencies

thymikee:
**Summary**

`pretty-format` is only used in tests, so there's no need for it to be included in dependencies.

**Test plan**

Green CI


packages/metro-config/package.json
```diff
@@ -17,10 +17,10 @@
     "jest-validate": "^24.9.0",
```
newyankeecodeshop: `jest-validate` 24.9 also has `pretty-format` in the runtime dependency list. Do we need to make this same change to their project?

PR #526, opened 2020-03-05
[Metro] update default blacklist

ShaMan123:
## Add ios/android build directories to blacklist
On windows rebuilding an app while metro is running causes both the build and metro to crash. 
This means that every time you rebuild your native code you need to restart metro and wait for it to complete bundling, which can be a nightmare when debugging native code.
This fixes this issues.

### DIFF
--- a/packages/metro-config/src/defaults/blacklist.js
+++ b/packages/metro-config/src/defaults/blacklist.js
@@ -18,6 +18,7 @@ var sharedBlacklist = [
   /website\/node_modules\/.*/,
   /heapCapture\/bundle\.js/,
   /.*\/__tests__\/.*/,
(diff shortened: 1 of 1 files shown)


cpojer:
I haven't heard about this before. I think it is likely that your setup is causing Metro to crash rather than an issue with Metro itself. I recommend using the `blacklist` option in your metro config to stop problematic folders from affecting your server.

I'm happy to consider merging this change if you can provide a repro and show that this is something happening to many users. Thank you!

ShaMan123:
This happens in every project I work on

בתאריך יום ה׳, 5 במרץ 2020, 14:31, מאת Christoph Nakazawa ‏<
notifications@github.com>:

PR #528, opened 2020-03-07
Delete appveyor.yml

imbsky:
**Summary**
This file is not used after all. If we want to test metro on Windows, we can do it with CircleCI or GitHub Actions.

**Test plan**
Not necessary for this PR.


imbsky:
👍

PR #533, opened 2020-03-16
Encode ..'s in asset paths to fix #290

### LINKED ISSUE #290: Asset paths (httpServerLocation) invalid when asset is in watchFolders
Bug

When I specify a directory in `watchFolders` from outside of my project root, the path reported via `httpServerLocation` is broken. For example, `/symlinked-watch-folder/subdir/image@3x.png`.

1. Create a React Native project in one directory.
2. Create a sibling directory called `symlinked-watch-folder` with a Node module in it named `symlinked-module`.
3. Add assets to this module, specifically one in `subdir/image@3x.png`.
4. In this module, reference the image like so: `<Image source={require('../subdir/image.png')} />` 
5. Now from the RN project directory, use npm to install the module: `npm install ../symlinked-watch-folder`.
6. Make a component in your project root that displays the module's component with the image in it.
7. Run the React Native project. The image will not appear and a request for the image will 404 because the path is invalid.

It looks like there are different valid locations, including `/assets/symlinked-module/subdir/image@3x.png` and even just `/subdir/image@3x.png`.

**More info**
Because this repo isn't very publicly documented, I can't tell exactly what's going on. But it looks like `getAssetData` is trying to `path.join` these two paths:
`/assets`
`../symlinked-watch-folder/subdir/image@3x.png`
https://github.com/facebook/metro/blob/bba48f068c6e709187d36170d3e5d49b7e9f6169/packages/metro/src/Assets.js#L181
The `../` wipes out the `/assets` and we end up with a result that isn't valid:
`/symlinked-watch-folder/subdir/image@3x.png`

I believe the relative path is coming from here:
https://github.com/facebook/metro/blob/bba48f068c6e709187d36170d3e5d49b7e9f6169/packages/metro/src/DeltaBundler/Serializers/getAssets.js#L44

The asset is being served correctly, it's just that [the `httpServerLocation` that's being generated](https://github.com/facebook/metro/blob/bba48f068c6e709187d36170d3e5d49b7e9f6169/packages/metro/src/Assets.js#L196) is not correct.
I can access the asset just fine if I correct the path to one of these:
`/assets/symlinked-module/subdir/image@3x.png`
`/subdir/image@3x.png`

If there is some way to handle this using the documented options, please let me know. (cc: @rafeca )

I've tried both Metro 0.47.1 and Metro 0.48.0
Node: 10.11.0
npm: 6.4.1
OS: macOS High Sierra

imownbey:
**Summary**
Android's http client will strip out '..'s from urls before sending a request. This causes projects using yarn workspaces with assets hosted outside of the project root to fail. For more information you can see #290 .

This fixes that by encoding ..'s as __'s inside `getAssetData` and then unencodes the __'s as ..'s in `getAsset`.

**Test plan**

Updated tests to reflect the changed URLs.

### DIFF
--- a/packages/metro/src/Assets.js
+++ b/packages/metro/src/Assets.js
@@ -191,12 +191,9 @@ async function getAssetData(
   platform: ?string = null,
   publicPath: string,
 ): Promise<AssetData> {
-  // If the path of the asset is outside of the projectRoot, we don't want to
-  // use `path.join` since this will generate an incorrect URL path. In that
-  // case we just concatenate the publicPath with the relative path.
-  let assetUrlPath = localPath.startsWith('..')
-    ? publicPath.replace(/\/$/, '') + '/' + path.dirname(localPath)
-    : path.join(publicPath, path.dirname(localPath));
+  // OKHTTP (Android HTTP client) strips out any '..' from a URL so we replace them with 2 _'s.
+  const escapedLocalPath = localPath.replace(/\.\.\//, '__/');
+  let assetUrlPath = path.join(publicPath, path.dirname(escapedLocalPath));
 
   // On Windows, change backslashes to slashes to get proper URL path from file path.
   if (path.sep === '\\') {
@@ -276,8 +273,8 @@ async function getAsset(
     relativePath,
     new Set(platform != null ? [platform] : []),
   );
-
-  const absolutePath = path.resolve(projectRoot, relativePath);
+  // Replaces __/ or __\ with ../ or ..\
+  const absolutePath = path.resolve(projectRoot, relativePath.replace(/__(\/|\\)/, '..$1'));
 
   if (!assetExts.includes(assetData.type)) {
     throw new Error(
@@ -290,8 +287,7 @@ async function getAsset(
       `'${relativePath}' could not be found, because it cannot be found in the project root or any watch folder`,
     );
   }
-
-  const record = await getAbsoluteAssetRecord(absolutePath, platform);
+  const record = await getAbsoluteAssetRecord(absolutePath, platform); // Matches __/ or __\
 
   for (let i = 0; i < record.scales.length; i++) {
     if (record.scales[i] >= assetData.resolution) {
--- a/packages/metro/src/__tests__/Assets-test.js
+++ b/packages/metro/src/__tests__/Assets-test.js
@@ -120,7 +120,7 @@ describe('getAsset', () => {
     ).toEqual(['b4 image', 'b4 ios image']);
   });
 
-  it('should find an image located on a watchFolder', async () => {
+  it('should find an image located on a watchFolder with encoded path', async () => {
     mkdirp.sync('/anotherfolder');
 
     writeImages({
@@ -129,7 +129,7 @@ describe('getAsset', () => {
 
     expect(
       await getAssetStr(
-        '../anotherfolder/b.png',
+        '__/anotherfolder/b.png',
         '/root',
         ['/anotherfolder'],
         null,
@@ -146,7 +146,7 @@ describe('getAsset', () => {
     });
 
     await expect(
-      getAssetStr('../anotherfolder/b.png', '/root', [], null, ['png']),
+      getAssetStr('__/anotherfolder/b.png', '/root', [], null, ['png']),
     ).rejects.toBeInstanceOf(Error);
   });
 });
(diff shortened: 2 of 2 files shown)


cpojer:
I'm not a fan of this change. Why not just add all assets as a root or watchFolder?

imownbey:
@cpojer Often people like to have assets live alongside the javascript, and when you have a project broken up into yarn workspaces that can mean that assets live above the root. I agree this isn't the best solution but it seemed to be the one with the smallest footprint.

The current behavior is ..'s will be in URLs in iOS and stripped out in Android so it is definitely broken behavior even if you disagree with people using workspaces and seems like one people will continue to run into. You can see other PRs like https://github.com/react-native-community/cli/pull/939#merged-event where this has had to be fixed in other parts of the workflow.

cpojer:
That doesn't answer my question, why not include all those files in roots/watch folders?

imownbey:
I believe the asset is always referenced from the projectRoot and not from watchFolders, although would be interested in understanding how I could fix this with watchFolders https://github.com/facebook/metro/blob/6f8b7b8357ec8e5b2ac35cecb99cef7417134374/packages/metro/src/Assets.js#L277

imownbey:
Maybe to better illustrate in our project we have multiple yarn workspaces, one for a design system `packages/designSystem/src/` and one for the app `packages/app/[src, ios, android, etc]`. We need to make projectRoot `packages/app/` so when something in say `packages/designSystem/src/Icons/X.jsx` does a `require('./icon.png')` and then is used inside of `app` the asset relative path is `../../designSystem/src/Icons/icon.png` but OKHTTP strips out those first two `../` so the path sent to metro is just `designSystem/src/Icons/icon.png` which 404s

janicduplessis:
@cpojer I think I pointed @imownbey towards this solution. This does happen for files included inside watchFolders (but not projectRoot). The case mentioned in the previous comment illustrates this. I guess an alternative to encoding `../` would be to have paths relative to the outer most watchFolder instead.

cpojer:
I'm in favor of merging this if we don't use a special symbol as part of the path that may cause conflicts. That seems hacky. Let's find an elegant solution that works.

janicduplessis:
@cpojer What about using some unique identifier like the asset hash (or whatever is the key of the asset in metro) and use that to build httpServerLocation. Maybe something like `/assets/<hash>.<ext>`. Then metro can extract that hash and lookup the path on disk to return the asset data.

cpojer:
I think that could work, yeah. I'm not sure how much work it is. We'd only have to do it for assets that are outside of the root, right?

imownbey:
Happy to look into that. Do you think the hash or moving the path to a param would be easier?

@cpojer I’d think it’d be better to request all assets the same way (by hash if moving to that) rather than changing based on if it’s in or out of the project dir, just for conformity sake. But if you think it’s better to one-off out of project files I can just do that. I just worry about bugs in one case and not the other being harder to track down.

belemaire:
@imownbey Thanks much for this PR :+1:
We are actually running into this issue, and are now hot patching metro with your fix while waiting for it to get merged, or a better solution offered (the hacky workaround suggested in https://github.com/facebook/metro/issues/290#issuecomment-543746458 won't work for us as we need to accommodate many different projects structures).
Just wanted to let you know that there is actually an issue with the code in your PR having to do with the replace statements. For example :

```javascript
// OKHTTP (Android HTTP client) strips out any '..' from a URL so we replace them with 2 _'s.
const escapedLocalPath = localPath.replace(/\.\.\//, '__/');
```

The problem here that you are mentioning that you `replace any '..' from an URL with __`, while the code will actually only replace the first occurence (i.e `../` will be replace by `__/` but `../../../` will be replaced by `__/../../`). The tests are actually only testing out a single level so I guess its slipped through.

Solution here is to simply add the global regex flag to the two replace statements you introduced in this PR. 
i.e 
```javascript
localPath.replace(/\.\.\//g, '__/');
``` 
and 
```javascript
relativePath.replace(/__(\/|\\)/g, '..$1')
```

imownbey:
Ah nice catch @belemaire. Updated test and added that. It might be good to only replace ..'s at the start of the path, happy to do that if people would prefer.

sibelius:
any progress on this?

rvion:
I'd love to have this too.
this is blocking for me as I share assets from a node_module above project root between several apps.

this allow assets to be found:

```js
  watchFolders: [path.resolve(__dirname, '../../../node_modules')],
```

BUT on android, it doesn't work, because

```
/assets/../../../node_modules/foo/bar
```

is reduced to 

```
/node_modules/foo/bar/
``` 

even before the request is dealt with. => so it ultimately is not found

rvion:
@imownbey @cpojer do you know a workaround for this ?

PR #545, opened 2020-04-18
New metro-hermes-compiler package @allow-large-files

cpojer:
Summary: This diff uses a WASM build of HBC to provide a simple JS interface that will be used by a follow-up diff to compile bytecode in Metro.

Differential Revision: D20792565


cpojer:
Just sharing the state of the code, will land the internal version.

PR #546, opened 2020-04-18
Add support for bytecode compilation & bytecode bundles

cpojer:
Summary:
This diff implements the proposal outlined in https://fb.quip.com/VIs3AvMWamUq: It adds bytecode compilation to Metro via a new `.bytecodebundle` endpoint. When a bundle is requested through this endpoint, two things happen:
* The transform worker returns both the compiled source code and bytecode (this can be changed as needed but both may be necessary for source maps and caching)
* The serializer will retrieve the bytecode output and concat the buffers together, as well as compiling the post modules on the fly using HBC. This is fast.

There are several upsides to this design:
* It is type safe (except for one cast) and doesn't add complexity to the existing bundle pipeline as this sort of loose-connectedness between the transform worker and serializer is part of Metro's supported design that also unlocks web features in Instagram and Airbnb.
* It did not require creating an abstraction on top of strings and buffers to treat them both as the same opaque type but rather intentionally chooses completely different code paths.
* This change doesn't affect the existing system and other endpoints like `.bundle` at all, meaning that we can land this change once I have iterated on the tests.

JavaScript source injects the `moduleId` and `dependencyMap` by slicing up the `__d` call to inject an array of numbers like this:
* Transformer generates `__d(module_factory)`
* Serializer slices off the `)` to inject `moduleId, dependencyMap)`

In order to make this work in Hermes bytecode we evaluated five solutions and picked the simplest one by injecting a global variable before each module: https://fb.quip.com/VIs3AvMWamUq#LPEACAtq3mY. This has a 1 MiB bundle size impact (Main JS source bundle (with bundle splitting): 8.8 MiB, base bytecode bundle: 6.7 MiB, bytecode with this diff: 7.7 MiB). This is an ok trade-off for now but I'll change the module headers to a custom JSON format either in this diff or a follow-up.

## ToDo (blocking)
* Write tests for the new bytecode bundle modules, Server.js, the worker and whatever else I touched.

## ToDo (non-blocking, rollout strategy)
* Understand how this will interact with Metro's existing global cache system and make the necessary changes
* Adapt mobile clients to be able to load a `.bytecodebundle` (see https://our.intern.facebook.com/intern/paste/P129069309 for an example from ReactVR)
* Actually test this end-to-end
* Manually measure bundle size, build times and "TTI" on device
* Add metrics and roll this out behind a GK
* Make this work with bundle splitting which currently uses `eval`
* Make this work with Fast Refresh (can likely keep using strings for the initial rollout)

Differential Revision: D21045629


cpojer:
Just sharing the state of the code, will land the internal version.

Issue #550, opened 2020-04-27
Option to disable constantFoldingPlugin in production builds


olivergeorge:
This is a feature request. 

Constant folding breaks some already minified code.  In order to do a release build with minified code its necessary to patch the worker.js file to comment out `plugins.push([constantFoldingPlugin, opts]);`.  

#291 describes the problem this feature would resolve.  This is still a problem as of 0.62.2.

Adding an option [here](https://github.com/facebook/metro/blob/95ee7bb69aeb1a9539772c2b7e7d2c74180d6fa8/packages/metro-transform-worker/src/index.js#L315) to disable constantFoldingPlugin would provide a simple configuration based solution.

robhogan:
Sorry this never got a response. The motivating #291 was fixed in v0.66.0. If folks still want to disable constant folding, please open a new issue.

Issue #291, opened 2018-10-12
Constant folding breaks some already minified code


chpill:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

See https://github.com/chpill/repro-bug-metro

The `index.js` file contains JS compiled from Clojurescript and JS optimized by google closure compiler. The minimalist app boots without issue when `__DEV__ == false` or when `minify == true`.
But when `__DEV == false && minify == true`, the app crashes with a JS error because of an undefined variable (`ReferenceError: $G__1978__22$$ is not defined`).

It turns out the issue comes from the constant folding optimization when applied on our already optimized/minified index.js

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**

To reproduce the bug:

* `git clone https://github.com/chpill/repro-bug-metro`
* `react-native start --reset-cache`
* `react-native run-android` then check that the app boots (you should see a green screen)
* Shake the device and go to "Dev settings", adjust options so that `__DEV == false && minify == true` and reload
* You should now see the exception `ReferenceError: $G__1978__22$$ is not defined`

To see that the problems comes from constant folding on our index.js:

* go to the metro package in the node_modules, and modify https://github.com/facebook/metro/blob/612d4d7ea3c1c96ec6eac42b12733fb425a181fc/packages/metro/src/JSTransformer/worker.js#L241-L244 with:
```
if (!options.dev) {
  if (filename === "index.js")
    console.log("======= CUSTOM PATCH @chpill ======== IGNORING CONSTANT FOLDING FOR", filename);
  else
    plugins.push([constantFoldingPlugin, opts]);

  plugins.push([inlinePlugin, opts]);
}
```

* stop the running packager if necessary, and launch `react-native start --reset-cache`
* reload the app
* Make sure you see the custom console log statement being issued while the packager is bundling the production minified artefact
* You should now see the application booting without issue.

**What is the expected behavior?**

The constant folding optimization should not produce invalid JS code, or there should be a way to opt out of it for specific files.

**Please provide your exact Metro configuration and mention your Metro, node, yarn/npm version and operating system.**

```
{
  "metro": "^0.48.1",
  "metro-babel-register": "^0.48.1",
  "metro-core": "^0.48.1",
  "metro-memory-fs": "^0.48.1"
}

yarn: 1.7.0
node: v8.11.3

OS: Linux 4.18.0-1-amd64
```

[...]

PR #552, opened 2020-04-28
Add option to disable constantFoldingPlugin

olivergeorge:
**Summary**

Constant folding breaks some already minified code.   

The current workaround is to patch the `worker.js` file to comment out 

```js
plugins.push([constantFoldingPlugin, opts]);
```

Adding an option to disable constantFoldingPlugin provides a configuration based solution.

Related issues:
#291 describes the problem this feature would resolve.
#550 is a request for an option to disable constantFoldingPlugin

**Test plan**

TBC

### DIFF
--- a/packages/metro-transform-worker/src/index.js
+++ b/packages/metro-transform-worker/src/index.js
@@ -312,7 +312,7 @@ module.exports = {
       ]);
     }
 
-    if (!options.dev) {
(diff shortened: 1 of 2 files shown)


olivergeorge:
Ah, this approach to an option option seems to require a lot of small changes across the code base.  

Closing for now.

PR #554, opened 2020-04-30
allow defaultConfigOverrides to work with configModule function

deecewan:
**Summary**

I was attempting to write a custom config for a `react-native` project (mostly for cache control), so I set up my `metro.config.js` to look something like this:

```js
module.exports = async (defaultConfig) => ({
  ...defaultConfig,
  cacheStores: [
    new FileStore(/* config */),
	...defaultConfig.cacheStores,
  ],
});
```

However, when I did that, I got compilation errors because of `flow` syntax not being compiled away. `@react-native-community/cli` [passes in it's own default config](https://github.com/react-native-community/cli/blob/e7d694672a1d4021267c9eda378282f4c97c193a/packages/cli/src/tools/loadMetroConfig.ts) as overrides when calling `loadConfig`.

Unfortunately, these are silently thrown away when you set `metro.config.js` to export a function.

This is one option for a "fix" (this will make my code work, but may not be generally liked). Another would be:

```diff
- const resultedConfig = await configModule(defaultConfig);
+ const resultedConfig = await configModule(defaultConfig, defaultConfigOverrides);
```

That one would be backwards compatible, because wouldn't touch the default config for older `configModule`s, and newer ones could accept the second argument.

**Test plan**

Unsure at this stage - if there is appetite for this change, I can look at adding tests.

### DIFF
--- a/packages/metro-config/src/loadConfig.js
+++ b/packages/metro-config/src/loadConfig.js
@@ -169,8 +169,10 @@ async function loadMetroConfigFromDisk(
   if (typeof configModule === 'function') {
     // Get a default configuration based on what we know, which we in turn can pass
     // to the function.
+    
+    const inputConfig = mergeConfig(defaultConfig, defaultConfigOverrides);
 
-    const resultedConfig = await configModule(defaultConfig);
+    const resultedConfig = await configModule(inputConfig);
     return resultedConfig;
   }
 


fson:
Without knowing about this pull request, I implemented something similar in https://github.com/facebook/metro/pull/559.

The difference is that I wanted to avoid passing the second `defaultConfigOverrides` argument to the config function (even if this is less backwards compatible), instead merging the `defaultConfigOverrides` with the Metro default values before passing them to the function. I think this might be preferrable since it avoids the confusion about which one you should use for any particular field (some fields might only be defined in the Metro default config).

cpojer:
Thanks for the PR! I believe https://github.com/facebook/metro/pull/559 will resolve the same issue, and it has tests, so I'll merge that one.

PR #555, opened 2020-04-30
Allow objects that only wrap react components

jfrolich:
**Summary**

ReasonML `reason-react` components are represented as an object with a single field `make`, that contains the actual javascript react component. This is due to language constraints. 

When a module is a react components this actually works with `fast-refesh` because in that case the only export is a function named `make` that is the component.

However this breaks down when there are more than one components defined in the file, in case we have to create a submodule, which again is an object with a single make function. This means we are exporting this object as well.

I wrote a small improvement that fast-refresh not only allows react components as exports, but also objects that just have components as keys.

**Test plan**

There are 5 failing tests now. It would be great if anyone can help tackle those.

Reference: https://github.com/reasonml/reason-react/issues/549#issuecomment-621557485


gaearon:
This really needs a test, we can’t merge without one. I can try to carve out time next week but in general you should be able to find an existing test (if this path is covered, that is) and copy paste it and then tweak it.

jfrolich:
I decided to fix it in `react-refresh` so the fix is not bundler specific. (See https://github.com/facebook/react/pull/20522.)

PR #556, opened 2020-04-30
[Fix] `DependencyGraph.resolveDependency` resolves to the same module when imported from `..`

raejin:
**Summary**
This is a part 2 to what I didn't anticipate from the previous PR: https://github.com/facebook/metro/pull/540

Essentially the previous PR didn't also account for the usage of `..`, this fixes the issue.

**Test plan**
- CI
- Test against our code base.

### DIFF
--- a/packages/metro/src/node-haste/DependencyGraph.js
+++ b/packages/metro/src/node-haste/DependencyGraph.js
@@ -252,6 +252,7 @@ class DependencyGraph extends EventEmitter {
     const isPath =
       to.includes('/') ||
       to === '.' ||
+      to === '..' ||
       from.includes(path.sep + 'node_modules' + path.sep);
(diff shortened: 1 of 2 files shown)


packages/metro/src/node-haste/DependencyGraph.js
```diff
@@ -252,6 +252,7 @@ class DependencyGraph extends EventEmitter {
     const isPath =
```
raejin, line 252: What essentially `isPath` is checking for?

cpojer: It's short-circuiting whether to use a "global" cache or a cache that's relative to the current path. At Facebook all module ids are global and engineers require them like `require('MyModule')` without a path. This code here optimizes the common use case of requiring a global module.

PR #557, opened 2020-05-02
[rn-babel-transformer]: use async transform

jrylan:
**Summary**

This change allows Babel to be configured using "babel.config.mjs", with native ES5 imports in Node.js >= 13. There is no impact if a project is still configured using "babel.config.js" or any other sync configuration.

**Test plan**

1. I reset Metro cache and succesfully recompiled the project a few times using a "babel.config.mjs" with a few native ESM custom plugins imported using standard import declarations.

2. I removed my "babel.config.mjs" file, and reverted back to the default babel.config.js supplied by React Native. I then cleared the metro cache (confirmed via the notice message) and recompiled the project several times successfully.

### DIFF
--- a/packages/metro-react-native-babel-transformer/src/index.js
+++ b/packages/metro-react-native-babel-transformer/src/index.js
@@ -72,6 +72,11 @@ const getBabelRC = (function() {
     }
 
     if (projectBabelRCPath) {
+      // babel.config.mjs
+      if (!fs.existsSync(projectBabelRCPath)) {
+        projectBabelRCPath = path.resolve(projectRoot, 'babel.config.mjs');
+      }
+      
       // .babelrc.js
       if (!fs.existsSync(projectBabelRCPath)) {
         projectBabelRCPath = path.resolve(projectRoot, '.babelrc.js');
--- a/packages/metro-react-native-babel-transformer/src/index.js
+++ b/packages/metro-react-native-babel-transformer/src/index.js
@@ -18,7 +18,7 @@ const inlineRequiresPlugin = require('babel-preset-fbjs/plugins/inline-requires'
 const makeHMRConfig = require('metro-react-native-babel-preset/src/configs/hmr');
 const path = require('path');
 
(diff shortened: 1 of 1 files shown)


jrylan:
I have no idea why the test fails for using transformFromAsync and such, most likely Babel needs to be updated to a recent version. I can add that to this PR or a separate PR if someone lets me know.

The failing Node 10 errors seem to be random bits that are completely unrelated to my change.

Also thanks to @zpao for resolving the CLA issue for me (I had already signed it in the past).

packages/metro-react-native-babel-transformer/src/index.js
```diff
@@ -161,7 +166,7 @@ function buildBabelConfig(filename, options, plugins?: BabelPlugins = []) {
   return Object.assign({}, babelRC, config);
 }
 
-function transform({filename, options, src, plugins}: BabelTransformerArgs) {
+async function transform({filename, options, src, plugins}: BabelTransformerArgs) {
```
cpojer, line 169: What about the callsite of this? Does it already `await` or work with Promises?

jrylan: Hi @cpojer! Thanks for taking a look. The answer is yes, it does!

I put together a demo repo here that uses `patch-package` to apply these changes.

https://github.com/jrylan/metro-react-native-babel-transformer-async-demo/tree/master

I've added notes in the repo README that shows that it works.

cpojer:
This diff breaks React Native on master, and it has incorrect flow type definitions.

Could you fix the flow issue (we have custom babel types in this repo which you'll need to adjust) and ensure this works with the master version of both Metro and React Native? Otherwise we won't be able to land this.

jrylan:
Sure, I’ll take a look. I’m eager to get this landed, in my testing the async transform had a pretty noticeable build speed improvement.

cpojer:
Why would this affect performance positively? Can you share some numbers?

PR #567, opened 2020-06-09
Prevent unhandled rejection with `buildCodeFrameMessage`

raejin:
We encounter regular resolution issue but was not captured by throwing
proper resolution error, this handles the specific case we're seeing.

```
(node:309) UnhandledPromiseRejectionWarning: Error: EISDIR: illegal operation on a directory, read
    at Object.readSync (fs.js:498:3)
    at tryReadSync (fs.js:332:20)
    at Object.readFileSync (fs.js:369:19)
    at UnableToResolveError.buildCodeFrameMessage 
```

Ever since upgraded, we've seen the above error messages representing in fact an actual resolution error that should have been represented by `Unable to resolve module %s from %s: %s` messages.

**Test plan**
- CI


cpojer (approved):
Thank you 👍

PR #574, opened 2020-06-29
metro-source-map: propagate sourcesContent when composing source maps

motiz88:
Summary:
* Implements the `sourceContentFor` API from `source-map` in Metro's `Consumer`.
* Copies source contents in `composeSourceMaps` if they exist.

Related issues: https://github.com/facebook/react-native/issues/26086, https://github.com/facebook/hermes/issues/85. We'll be able to close the RN issue once this lands in the version of `metro-source-map` used in RN master.

Differential Revision: D22284159

### DIFF
--- a/packages/metro-source-map/src/Consumer/AbstractConsumer.js
+++ b/packages/metro-source-map/src/Consumer/AbstractConsumer.js
@@ -54,9 +54,13 @@ class AbstractConsumer implements IConsumer {
     }
   }
 
-  // flowlint unsafe-getters-setters:off
+  // flowlint-next-line unsafe-getters-setters:off
   get file(): ?string {
     return this._sourceMap.file;
+  }
+
+  sourceContentFor(source: string, nullOnMissing: true): ?string {
+    invariant(false, 'Not implemented');
   }
 }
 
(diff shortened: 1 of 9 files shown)


fbartho:
@motiz88 -- will this detect Sourcemaps when using the TypeScript compiler in a standalone way?

Example:
-`tsc` builds to `dist/`
- Metro/RN configured to load `dist/` for the application.

Alternatively, does this require babel-plugin-typescript? or is this Hermes-only?

motiz88:
Hey @fbartho! Metro by itself does not read source maps from its input.

The `metro-source-map` package does, however, export the [`composeSourceMaps`](https://github.com/facebook/metro/blob/master/packages/metro-source-map/src/composeSourceMaps.js) function (which is the context for this PR). This in turn is used in the React Native + Hermes integration (see RN's [`compose-source-maps` script](https://github.com/facebook/react-native/blob/master/scripts/compose-source-maps.js)) to generate a correct source map for that scenario:

1. Metro builds a JS bundle with a source map.
2. Hermes compiles the JS bundle to bytecode and creates its own (partial) source map.
3. The build script invokes `composeSourceMaps` to combine the source maps from (1) and (2).

It sounds like you may be able to use `composeSourceMaps` similarly in your custom build. If you run into problems using this function, feel free to open a new issue.

fbartho:
Ah! -- It looks like RN's compose-source-maps script explicitly doesn't support the `RAM Bundle Format`. So I guess I'm out-of-luck there -- but that's not obviously a problem that I should request help with here.

Thanks for your prompt response @motiz88 -- and at least this is a stepping stone to my dreamland :P 

https://github.com/facebook/react-native/blob/753fdebcd4e03f4b809051ddbb23fd09413bbf35/scripts/compose-source-maps.js#L42-L54

PR #592, opened 2020-10-05
Update Configuration.md

liorJuice:
Having `watchFolders` along with `extraNodeModules` options, led us to believe `watchFolders` are there only to signal when to rebuild and you need to also include them in `extraNodeModules` if you want them in your bundle. which is, as far as i understand, not correct. so i suggest this clarification. if you can confirm the clarifications are correct, please add them to the docs.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -59,8 +59,8 @@ The root folder of your project.
 
 Type: `Array<string>`
 
-Specify any additional (to projectRoot) watch folders, this is used to know which files to watch.
-(By default the file watching is disabled in CI environments. Also it can be manually disabled by setting the env variable `CI=true`)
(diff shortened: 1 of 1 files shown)


Jesus75:
> Having `watchFolders` along with `extraNodeModules` options, led us to believe `watchFolders` are there only to signal when to rebuild and you need to also include them in `extraNodeModules` if you want them in your bundle. which is, as far as i understand, not correct. so i suggest this clarification. if you can confirm the clarifications are correct, please add them to the docs.

Sms

L10s,m
All good

cpojer (approved):
Thank you!

cpojer:
Could you sign the CLA?

rh389:
Closing as we never received a CLA

PR #593, opened 2020-10-06
Remove JSON from default asset types

### LINKED ISSUE #487: JSON files from project and `node_modules` dependencies are duplicate to the js bundle and to android resources `res/raw`
When building android bundle, JSON files from project and `node_modules` dependencies are duplicate to the js bundle and to android resources `res/raw`. 
bundle

My project have big JSON data ~13MB (compressed) `PROJECT_ROOT/src/assets/words.json`, which are needed only in JS. When duplicating the size of the bundle increases several times.
Faced this issue after updating from RN 0.59 to RN 0.61. 

I think the problem became relevant after this commit https://github.com/facebook/metro/commit/dcb41e39f0df6ae1e0b3dfeff9ef5a69128830d5#diff-235a3e5d21175615e1cc254dd8b17eb2 , because if i override `assetExts` option and remove `json` everything works well and as before.

Is this problem general or mine? It does not seem to me a good idea to copy all json from all dependencies to android resources. If this is normal behavior, is it possible to exclude json files from my project?

- Ubuntu 18.04.3 LTS
- node v10.17.0
- npm@6.12.1
- yarn@1.19.1
- metro@0.56.3
- metro-config@0.56.3
- RN@0.61.4

janicduplessis:
**Summary**

JSON files are currently considered both source and assets so they are duplicated in the js bundle and in the assets copied in the final app. I think this is a better default and if someone relies on json files being also included in the bundle they can add it back in their metro config.

Fixes #487

**Test plan**

Test that json files are no longer present in my app assets and the app still works.


cpojer:
We actually rely on this at FB as we access JSON files from native 🤔

janicduplessis:
Hmm so native looks for the file in assets and relies on RN to bundle it? Seems like quite the hack, but I don't see another easy way for native to access it without JS sending it though the bridge. Do you think it'd be possible to update the facebook metro config to add back JSON in asset types and remove it from the default OSS config?

cpojer:
Yeah we may be able to do that, I can take a look. It may sound hacky but it allows us to do a ton of work in native to preload stuff instead of initializing a JS VM.

PR #601, opened 2020-10-25
Update Metro configuration docs to add hasteMapCacheDirectory

qihaomah:
**Summary**
- hasteMapCacheDirectory is added in https://github.com/facebook/metro/pull/424 but the docs have not been added
- Updated `CONTRIBUTING.md` to use `yarn` instead of `npm install`. (The repo is having `yarn.lock` instead of `package-lock.json`).

**Test plan**
- Only docs updated, no test is done.


cpojer (approved):
Thank you.

PR #602, opened 2020-10-27
feat: update lazy import

Naturalclar:
**Summary**

Updated the lazy-import config to match the `index.js` of currently released react-native version
https://github.com/facebook/react-native/blob/0.63-stable/index.js

**Test plan**
pass unit test
be able to import added modules


cpojer (approved):
Thank you!

PR #603, opened 2020-10-27
update istanbul to remove core-js

Naturalclar:
**Summary**
Updated istanbul to latest version to remove core-js from node_modules.

Also updated node version of CI to use node v12, since node v10 does not have Array.prototype.flatmap implementation which is required in the test cases. (It was polyfilled by core-js before)

**Test plan**
CI passes

Reduced node_modules size from 228MB down to 217MB


cpojer (approved):
Very nice! 👍

PR #604, opened 2020-10-28
Fix regular expression compatibility issue with node v12

### LINKED ISSUE #578: Broken regular expression in Windows exclusionList.js path separator replacement

Bug

On Windows, I get an error when running `react-native start`:

```
Invalid regular expression: /(.*\\__fixtures__\\.*|node_modules[\\\]react[\\\]dist[\\\].*|website\\node_modules\\.*|heapCapture\\bundle\.js|.*\\__tests__\\.*)$/: Unterminated character class. Run CLI with --verbose flag for more details.
SyntaxError: Invalid regular expression: /(.*\\__fixtures__\\.*|node_modules[\\\]react[\\\]dist[\\\].*|website\\node_modules\\.*|heapCapture\\bundle\.js|.*\\__tests__\\.*)$/: Unterminated character class
```

**The Fix**

```patch
 function escapeRegExp(pattern) {
   if (Object.prototype.toString.call(pattern) === '[object RegExp]') {
-    return pattern.source.replace(/\//g, path.sep);
+    return pattern.source.replace(/\//g, '\\' + path.sep);
   } else if (typeof pattern === 'string') {
     var escaped = pattern.replace(/[\-\[\]\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
     // convert the '/' into an escaped local file separator
     return escaped.replace(/\//g, '\\' + path.sep);
#-----------------------------------^ The replace method above should also escape `path.sep`
 ```

If you look at [line 22 of exclusionList.js](https://github.com/facebook/metro/blob/94c0b541b4bfa17aee4efa0f1969565522ce830d/packages/metro-config/src/defaults/exclusionList.js#L22) you'll see it replaces `/` with `"\\" + path.sep` but on [line 18](https://github.com/facebook/metro/blob/94c0b541b4bfa17aee4efa0f1969565522ce830d/packages/metro-config/src/defaults/exclusionList.js#L18) it replaces the source with only `path.sep`. Instead it should also replace it with `"\\" + path.sep`.

### PR DESCRIPTION
**Pull request by ryancat**
**Summary**
This pull request fixes #578 we have in regular expression treatment for converting OS specific path separators. When running in node js version 12, forward slashes in brackets (eg. `/[/\\]foo/` won't be escaped automatically (Also see another issue I filed in nodejs project https://github.com/nodejs/help/issues/3039). That is causing backward compatibility issue in our use case, where we assume they will always be escaped and replaced the string accordingly.

**Test plan**
Added new tests to cover this edge case. Those new tests should pass in both nodejs 10 and nodejs 12.

Before:
```
xch at xch-mbp in ~/oss/metro/packages/metro-configfix_regexp_metro_config
$ nvm use 10
Now using node v10.19.0 (npm v6.13.4)

xch at xch-mbp in ~/oss/metro/packages/metro-configfix_regexp_metro_config [!]
$ jest src/defaults/
jest-haste-map: Watchman crawl failed. Retrying once with node crawler.
  Usually this happens when watchman isn't running. Create an empty `.watchmanconfig` file in your project's root folder or initialize a git or hg repository in your project.
  Error: Watchman error: N8watchman22CommandValidationErrorE: failed to validate command: resolve_projpath:  None of the files listed in global config root_files are present in path `/Users/xch/oss/metro/packages/metro-config` or any of its parent directories.  root_files is defined by the `/etc/watchman.json` config file and includes `.watchmanconfig`.  One or more of these files must be present in order to allow a watch. Try pulling and checking out a newer version of the project?. Make sure watchman is running for this project. See https://facebook.github.io/watchman/docs/troubleshooting.
jest-haste-map: duplicate manual mock found: cosmiconfig
  The following files share their name; please delete one of them:
    * <rootDir>/build/__mocks__/cosmiconfig.js
    * <rootDir>/src/__mocks__/cosmiconfig.js

 PASS  src/defaults/__tests__/exclusionList-test.js
  exclusionList
    ✓ proves we can write to path.sep for setting up the tests (2 ms)
    ✓ converts forward slashes in the RegExp to the OS specific path separator in macOS/linux (2 ms)
    ✓ converts forward slashes in the RegExp to the OS specific path separator in windows
    ✓ converts forward slashes in the string to the OS specific path separator in macOS/linux (1 ms)
    ✓ converts forward slashes in the string to the OS specific path separator in windows
    ✓ converts forward slashes in the RegExp to the OS specific path separator for macOS/linux in nodejs 10 or below (1 ms)
    ✓ converts forward slashes in the RegExp to the OS specific path separator for windows in nodejs 10 or below

Test Suites: 1 passed, 1 total
Tests:       7 passed, 7 total
Snapshots:   0 total
Time:        0.824 s, estimated 1 s
Ran all test suites matching /src\/defaults\//i.

xch at xch-mbp in ~/oss/metro/packages/metro-configfix_regexp_metro_config [!]
$ nvm use 12
Now using node v12.18.4 (npm v6.14.6)

xch at xch-mbp in ~/oss/metro/packages/metro-configfix_regexp_metro_config [!]
$ jest src/defaults/
jest-haste-map: Watchman crawl failed. Retrying once with node crawler.
  Usually this happens when watchman isn't running. Create an empty `.watchmanconfig` file in your project's root folder or initialize a git or hg repository in your project.
  Error: Watchman error: N8watchman22CommandValidationErrorE: failed to validate command: resolve_projpath:  None of the files listed in global config root_files are present in path `/Users/xch/oss/metro/packages/metro-config` or any of its parent directories.  root_files is defined by the `/etc/watchman.json` config file and includes `.watchmanconfig`.  One or more of these files must be present in order to allow a watch. Try pulling and checking out a newer version of the project?. Make sure watchman is running for this project. See https://facebook.github.io/watchman/docs/troubleshooting.
jest-haste-map: duplicate manual mock found: cosmiconfig
  The following files share their name; please delete one of them:
    * <rootDir>/build/__mocks__/cosmiconfig.js
    * <rootDir>/src/__mocks__/cosmiconfig.js

 FAIL  src/defaults/__tests__/exclusionList-test.js
  exclusionList
    ✓ proves we can write to path.sep for setting up the tests (2 ms)
    ✓ converts forward slashes in the RegExp to the OS specific path separator in macOS/linux
    ✕ converts forward slashes in the RegExp to the OS specific path separator in windows (3 ms)
    ✓ converts forward slashes in the string to the OS specific path separator in macOS/linux (1 ms)
    ✓ converts forward slashes in the string to the OS specific path separator in windows
    ✓ converts forward slashes in the RegExp to the OS specific path separator for macOS/linux in nodejs 10 or below (1 ms)
    ✓ converts forward slashes in the RegExp to the OS specific path separator for windows in nodejs 10 or below

  ● exclusionList › converts forward slashes in the RegExp to the OS specific path separator in windows

    SyntaxError: Invalid regular expression: /(.*[\\\]foo[\\\]bar|website\\node_modules\\.*|.*\\__tests__\\.*)$/: Unterminated character class
        at new RegExp (<anonymous>)

      31 | 
      32 | function exclusionList(additionalExclusions) {
    > 33 |   return new RegExp(
         |          ^
      34 |     '(' +
      35 |       (additionalExclusions || [])
      36 |         .concat(list)

      at exclusionList (src/defaults/exclusionList.js:33:10)
      at Object.<anonymous> (src/defaults/__tests__/exclusionList-test.js:59:34)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 6 passed, 7 total
Snapshots:   0 total
Time:        0.757 s, estimated 1 s
Ran all test suites matching /src\/defaults\//i.
```

After:
```
xch at xch-mbp in ~/oss/metro/packages/metro-configfix_regexp_metro_config
$ nvm use 10
Now using node v10.19.0 (npm v6.13.4)

xch at xch-mbp in ~/oss/metro/packages/metro-configfix_regexp_metro_config
$ jest src/defaults/
jest-haste-map: Watchman crawl failed. Retrying once with node crawler.
  Usually this happens when watchman isn't running. Create an empty `.watchmanconfig` file in your project's root folder or initialize a git or hg repository in your project.
  Error: Watchman error: N8watchman22CommandValidationErrorE: failed to validate command: resolve_projpath:  None of the files listed in global config root_files are present in path `/Users/xch/oss/metro/packages/metro-config` or any of its parent directories.  root_files is defined by the `/etc/watchman.json` config file and includes `.watchmanconfig`.  One or more of these files must be present in order to allow a watch. Try pulling and checking out a newer version of the project?. Make sure watchman is running for this project. See https://facebook.github.io/watchman/docs/troubleshooting.
jest-haste-map: duplicate manual mock found: cosmiconfig
  The following files share their name; please delete one of them:
    * <rootDir>/build/__mocks__/cosmiconfig.js
    * <rootDir>/src/__mocks__/cosmiconfig.js

 PASS  src/defaults/__tests__/exclusionList-test.js
  exclusionList
    ✓ proves we can write to path.sep for setting up the tests (2 ms)
    ✓ converts forward slashes in the RegExp to the OS specific path separator in macOS/linux (2 ms)
    ✓ converts forward slashes in the RegExp to the OS specific path separator in windows
    ✓ converts forward slashes in the string to the OS specific path separator in macOS/linux (1 ms)
    ✓ converts forward slashes in the string to the OS specific path separator in windows
    ✓ converts forward slashes in the RegExp to the OS specific path separator for macOS/linux in nodejs 10 or below
    ✓ converts forward slashes in the RegExp to the OS specific path separator for windows in nodejs 10 or below (1 ms)

Test Suites: 1 passed, 1 total
Tests:       7 passed, 7 total
Snapshots:   0 total
Time:        0.998 s, estimated 1 s
Ran all test suites matching /src\/defaults\//i.

xch at xch-mbp in ~/oss/metro/packages/metro-configfix_regexp_metro_config
$ nvm use 12
Now using node v12.18.4 (npm v6.14.6)

xch at xch-mbp in ~/oss/metro/packages/metro-configfix_regexp_metro_config
$ jest src/defaults/
jest-haste-map: Watchman crawl failed. Retrying once with node crawler.
  Usually this happens when watchman isn't running. Create an empty `.watchmanconfig` file in your project's root folder or initialize a git or hg repository in your project.
  Error: Watchman error: N8watchman22CommandValidationErrorE: failed to validate command: resolve_projpath:  None of the files listed in global config root_files are present in path `/Users/xch/oss/metro/packages/metro-config` or any of its parent directories.  root_files is defined by the `/etc/watchman.json` config file and includes `.watchmanconfig`.  One or more of these files must be present in order to allow a watch. Try pulling and checking out a newer version of the project?. Make sure watchman is running for this project. See https://facebook.github.io/watchman/docs/troubleshooting.
jest-haste-map: duplicate manual mock found: cosmiconfig
  The following files share their name; please delete one of them:
    * <rootDir>/build/__mocks__/cosmiconfig.js
    * <rootDir>/src/__mocks__/cosmiconfig.js

 PASS  src/defaults/__tests__/exclusionList-test.js
  exclusionList
    ✓ proves we can write to path.sep for setting up the tests (2 ms)
    ✓ converts forward slashes in the RegExp to the OS specific path separator in macOS/linux (1 ms)
    ✓ converts forward slashes in the RegExp to the OS specific path separator in windows
    ✓ converts forward slashes in the string to the OS specific path separator in macOS/linux (1 ms)
    ✓ converts forward slashes in the string to the OS specific path separator in windows
    ✓ converts forward slashes in the RegExp to the OS specific path separator for macOS/linux in nodejs 10 or below
    ✓ converts forward slashes in the RegExp to the OS specific path separator for windows in nodejs 10 or below (1 ms)

Test Suites: 1 passed, 1 total
Tests:       7 passed, 7 total
Snapshots:   0 total
Time:        0.741 s, estimated 1 s
Ran all test suites matching /src\/defaults\//i.
```

### DIFF
--- a/packages/metro-config/src/defaults/__tests__/exclusionList-test.js
+++ b/packages/metro-config/src/defaults/__tests__/exclusionList-test.js
@@ -25,14 +25,84 @@ describe('exclusionList', () => {
     path.sep = originalSeparator;
   });
 
-  it('converts forward slashes in the RegExp to the OS specific path separator', () => {
+  it('proves we can write to path.sep for setting up the tests', () => {
     // $FlowFixMe: property sep is not writable.
     path.sep = '/';
-    expect('a/b/c').toMatch(exclusionList([new RegExp('a/b/c')]));
-
+    expect(require('path').sep).toBe('/');
     // $FlowFixMe: property sep is not writable.
     path.sep = '\\';
     expect(require('path').sep).toBe('\\');
+  });
+
+  it('converts forward slashes in the RegExp to the OS specific path separator in macOS/linux', () => {
+    // $FlowFixMe: property sep is not writable.
+    path.sep = '/';
+    // Simple case
+    expect('a/b/c').toMatch(exclusionList([new RegExp('a/b/c')]));
+    expect('a/b/c').toMatch(exclusionList([/a\/b\/c/]));
+    // Regular expression that already considered OS specific path separator.
+    // When explictly construct RegExp instance, the string needs to be escaped first.
+    // eg. /.*[/\\]foo[/\\]bar/ => '.*[/\\\\]foo[/\\\\]bar'
+    expect('/foo/bar').toMatch(exclusionList([new RegExp('.*[/\\\\]foo[/\\\\]bar')]));
+    expect('/foo/bar').toMatch(exclusionList([/.*[/\\]foo[/\\]bar/]));
+  });
+
+  it('converts forward slashes in the RegExp to the OS specific path separator in windows', () => {
+    // $FlowFixMe: property sep is not writable.
+    path.sep = '\\';
+    // Simple case
     expect('a\\b\\c').toMatch(exclusionList([new RegExp('a/b/c')]));
+    expect('a\\b\\c').toMatch(exclusionList([/a\/b\/c/]));
+    // Regular expression that already considered OS specific path separator.
+    // When explictly construct RegExp instance, the string needs to be escaped first.
+    // eg. /.*[/\\]foo[/\\]bar/ => '.*[/\\\\]foo[/\\\\]bar'
+    expect('\\foo\\bar').toMatch(exclusionList([new RegExp('.*[/\\\\]foo[/\\\\]bar')]));
+    expect('\\foo\\bar').toMatch(exclusionList([/.*[/\\]foo[/\\]bar/]));
+  });
+
+  it('converts forward slashes in the string to the OS specific path separator in macOS/linux', () => {
+    // $FlowFixMe: property sep is not writable.
+    path.sep = '/';
+    // Simple case
+    expect('a/b/c').toMatch(exclusionList(['a/b/c']));
+    // Strings that already considered OS specific path separator.
+    expect('/foo/bar').toMatch(exclusionList(['.*[/\\]foo[/\\]bar']));
+  });
+
+  it('converts forward slashes in the string to the OS specific path separator in windows', () => {
+    // $FlowFixMe: property sep is not writable.
+    path.sep = '\\';
+    // Simple case
+    expect('a\\b\\c').toMatch(exclusionList(['a/b/c']));
+    // Strings that already considered OS specific path separator.
+    expect('\\foo\\bar').toMatch(exclusionList(['.*[/\\]foo[/\\]bar']));
+  });
+
+  it('converts forward slashes in the RegExp to the OS specific path separator for macOS/linux in nodejs 10 or below', () => {
+    // In node version 10 or below, the forward slash in brackets are escaped automatically.
+    // eg. /[/\\]/ => /[\/\\]/
+    // Ideally this test case should be removed and instead the whole test should run in
+    // multiple node versions.
+    // $FlowFixMe: property sep is not writable.
+    path.sep = '/';
+    // Regular expression that already considered OS specific path separator.
+    // When explictly construct RegExp instance, the string needs to be escaped first.
+    // eg. /.*[\/\\]foo[\/\\]bar/ => '.*[\\/\\\\]foo[\\/\\\\]bar'
+    expect('/foo/bar').toMatch(exclusionList([new RegExp('.*[\\/\\\\]foo[\\/\\\\]bar')]));
+    expect('/foo/bar').toMatch(exclusionList([/.*[\/\\]foo[\/\\]bar/]));
+  });
+
+  it('converts forward slashes in the RegExp to the OS specific path separator for windows in nodejs 10 or below', () => {
+    // In node version 10 or below, the forward slash in brackets are escaped automatically.
+    // eg. /[/\\]/ => /[\/\\]/
+    // Ideally this test case should be removed and instead the whole test should run in
+    // multiple node versions.
+    // $FlowFixMe: property sep is not writable.
+    path.sep = '\\';
+    // Regular expression that already considered OS specific path separator.
+    // When explictly construct RegExp instance, the string needs to be escaped first.
+    // eg. /.*[\/\\]foo[\/\\]bar/ => '.*[\\/\\\\]foo[\\/\\\\]bar'
+    expect('\\foo\\bar').toMatch(exclusionList([new RegExp('.*[\\/\\\\]foo[\\/\\\\]bar')]));
+    expect('\\foo\\bar').toMatch(exclusionList([/.*[\/\\]foo[\/\\]bar/]));
   });
 });
--- a/packages/metro-config/src/defaults/__tests__/exclusionList-test.js
+++ b/packages/metro-config/src/defaults/__tests__/exclusionList-test.js
@@ -39,22 +39,20 @@ describe('exclusionList', () => {
   });
 
   describe('simulate macOS/linux enviornment', () => {
-    it('converts forward slashes in the RegExp to the OS specific path separator in macOS/linux', () => {
-      setPathSeperator('/');
+    beforeEach(() => setPathSeperator('/'));
+
+    it('converts forward slashes in the RegExp to the OS specific path separator', () => {
       // Simple case
       expect('a/b/c').toMatch(exclusionList([new RegExp('a/b/c')]));
       expect('a/b/c').toMatch(exclusionList([/a\/b\/c/]));
       // Regular expression that already considered OS specific path separator.
-      // When explictly construct RegExp instance, the string needs to be escaped first.
-      // eg. /.*[/\\]foo[/\\]bar/ => '.*[/\\\\]foo[/\\\\]bar'
       expect('/foo/bar').toMatch(
         exclusionList([new RegExp('.*[/\\\\]foo[/\\\\]bar')]),
       );
       expect('/foo/bar').toMatch(exclusionList([/.*[/\\]foo[/\\]bar/]));
     });
 
-    it('converts forward slashes in the string to the OS specific path separator in macOS/linux', () => {
-      setPathSeperator('/');
+    it('converts forward slashes in the string to the OS specific path separator', () => {
       // Simple case
       expect('a/b/c').toMatch(exclusionList(['a/b/c']));
       // Make sure the special characters are escaped properly
@@ -63,15 +61,12 @@ describe('exclusionList', () => {
       );
     });
 
-    it('converts forward slashes in the RegExp to the OS specific path separator for macOS/linux in nodejs 10 or below', () => {
+    it('converts forward slashes in the RegExp to the OS specific path separator in nodejs 10 or below', () => {
       // In node version 10 or below, the forward slash in brackets are escaped automatically.
       // eg. /[/\\]/ => /[\/\\]/
       // Ideally this test case should be removed and instead the whole test should run in
       // multiple node versions.
-      setPathSeperator('/');
       // Regular expression that already considered OS specific path separator.
-      // When explictly construct RegExp instance, the string needs to be escaped first.
-      // eg. /.*[\/\\]foo[\/\\]bar/ => '.*[\\/\\\\]foo[\\/\\\\]bar'
       expect('/foo/bar').toMatch(
         exclusionList([new RegExp('.*[\\/\\\\]foo[\\/\\\\]bar')]),
       );
@@ -80,22 +75,20 @@ describe('exclusionList', () => {
   });
 
   describe('simulate windows enviornment', () => {
-    it('converts forward slashes in the RegExp to the OS specific path separator in windows', () => {
-      setPathSeperator('\\');
+    beforeEach(() => setPathSeperator('\\'));
+    
+    it('converts forward slashes in the RegExp to the OS specific path separator', () => {
       // Simple case
       expect('a\\b\\c').toMatch(exclusionList([new RegExp('a/b/c')]));
       expect('a\\b\\c').toMatch(exclusionList([/a\/b\/c/]));
       // Regular expression that already considered OS specific path separator.
-      // When explictly construct RegExp instance, the string needs to be escaped first.
-      // eg. /.*[/\\]foo[/\\]bar/ => '.*[/\\\\]foo[/\\\\]bar'
       expect('\\foo\\bar').toMatch(
         exclusionList([new RegExp('.*[/\\\\]foo[/\\\\]bar')]),
       );
       expect('\\foo\\bar').toMatch(exclusionList([/.*[/\\]foo[/\\]bar/]));
     });
 
-    it('converts forward slashes in the string to the OS specific path separator in windows', () => {
-      setPathSeperator('\\');
+    it('converts forward slashes in the string to the OS specific path separator', () => {
       // Simple case
       expect('a\\b\\c').toMatch(exclusionList(['a/b/c']));
       // Make sure the special characters are escaped properly
@@ -104,15 +97,12 @@ describe('exclusionList', () => {
       );
     });
 
-    it('converts forward slashes in the RegExp to the OS specific path separator for windows in nodejs 10 or below', () => {
+    it('converts forward slashes in the RegExp to the OS specific path separator in nodejs 10 or below', () => {
       // In node version 10 or below, the forward slash in brackets are escaped automatically.
       // eg. /[/\\]/ => /[\/\\]/
       // Ideally this test case should be removed and instead the whole test should run in
       // multiple node versions.
-      setPathSeperator('\\');
       // Regular expression that already considered OS specific path separator.
-      // When explictly construct RegExp instance, the string needs to be escaped first.
-      // eg. /.*[\/\\]foo[\/\\]bar/ => '.*[\\/\\\\]foo[\\/\\\\]bar'
       expect('\\foo\\bar').toMatch(
         exclusionList([new RegExp('.*[\\/\\\\]foo[\\/\\\\]bar')]),
       );
--- a/packages/metro-config/src/defaults/__tests__/exclusionList-test.js
+++ b/packages/metro-config/src/defaults/__tests__/exclusionList-test.js
@@ -76,7 +76,7 @@ describe('exclusionList', () => {
 
   describe('simulate windows enviornment', () => {
     beforeEach(() => setPathSeperator('\\'));
-    
+
     it('converts forward slashes in the RegExp to the OS specific path separator', () => {
       // Simple case
       expect('a\\b\\c').toMatch(exclusionList([new RegExp('a/b/c')]));
--- a/packages/metro-config/src/defaults/__tests__/exclusionList-test.js
+++ b/packages/metro-config/src/defaults/__tests__/exclusionList-test.js
@@ -16,6 +16,12 @@ const path = require('path');
 
 describe('exclusionList', () => {
   let originalSeparator;
+
+  function setPathSeperator(sep: string) {
+    // $FlowFixMe: property sep is not writable.
+    path.sep = sep;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/metro-config/src/defaults/exclusionList.js [hunk 1]
```diff
@@ -15,10 +15,19 @@ var list = [/website\/node_modules\/.*/, /.*\/__tests__\/.*/];
 
 function escapeRegExp(pattern) {
   if (Object.prototype.toString.call(pattern) === '[object RegExp]') {
-    return pattern.source.replace(/\//g, path.sep);
+    // the forward slash may or may not be escaped in regular expression depends
+    // on if it's in brackets. eg. /foo\/bar/ and /[/\\]foo/ are both valid.
```
MichaReiser, line 19: This example is somewhat confusing because there are multiple things that changes (bar no longer present, using `\\`). I think what you meant is that

`/foo\/bar/` and `/foo[/]bar/` are equal

packages/metro-config/src/defaults/__tests__/exclusionList-test.js
```diff
@@ -25,14 +25,84 @@ describe('exclusionList', () => {
...
+    path.sep = '\\';
+    expect(require('path').sep).toBe('\\');
+  });
+
+  it('converts forward slashes in the RegExp to the OS specific path separator in macOS/linux', () => {
+    // $FlowFixMe: property sep is not writable.
+    path.sep = '/';
+    // Simple case
     expect('a/b/c').toMatch(exclusionList([new RegExp('a/b/c')]));
+    expect('a/b/c').toMatch(exclusionList([/a\/b\/c/]));
```
MichaReiser, line 42: Would you mind adding a test for the case when the passed in path is a 'string'. The line 40 & 41 test the same path since both create a `RegExp`.

packages/metro-config/src/defaults/__tests__/exclusionList-test.js
```diff
@@ -25,14 +25,84 @@ describe('exclusionList', () => {
...
+  });
+
+  it('converts forward slashes in the string to the OS specific path separator in macOS/linux', () => {
+    // $FlowFixMe: property sep is not writable.
+    path.sep = '/';
+    // Simple case
+    expect('a/b/c').toMatch(exclusionList(['a/b/c']));
+    // Strings that already considered OS specific path separator.
+    expect('/foo/bar').toMatch(exclusionList(['.*[/\\]foo[/\\]bar']));
+  });
```
MichaReiser, line 70: This already seems to be tested in `converts forward slashes in the RegExp to the OS specific path separator in macOS/linux`. I personally prefer the split but we should then remove it from the other test case.

packages/metro-config/src/defaults/__tests__/exclusionList-test.js
```diff
@@ -25,14 +25,84 @@ describe('exclusionList', () => {
...
+    // Strings that already considered OS specific path separator.
+    expect('\\foo\\bar').toMatch(exclusionList(['.*[/\\]foo[/\\]bar']));
+  });
+
+  it('converts forward slashes in the RegExp to the OS specific path separator for macOS/linux in nodejs 10 or below', () => {
+    // In node version 10 or below, the forward slash in brackets are escaped automatically.
+    // eg. /[/\\]/ => /[\/\\]/
+    // Ideally this test case should be removed and instead the whole test should run in
+    // multiple node versions.
+    // $FlowFixMe: property sep is not writable.
```
MichaReiser, line 86: I believe we can emulate the node 10 behavior by writing a single test that asserts that two RegExp are the same regardless if the slash is automatically escaped.

```
const escaped = exclusionList(/[\/]/);
const nonEscaped = exclusionList(/[/]/); 

expect(nonEscaped.source).toEqual(escaped.source);

**Review by MichaReiser — COMMENTED** [review]
Thanks @ryancat for creating this PR and helping to significantly increase the test coverage of the `exclusionList`. 

I've made a couple of comments. My main question is why the path where the user passes a string has to change as well. My understanding is that the main issue is the missing escaping of `path.sep` in the case where a `RegExp` is passed.

File: packages/metro-config/src/defaults/__tests__/exclusionList-test.js [hunk 5]
```diff
@@ -25,14 +25,84 @@ describe('exclusionList', () => {
     path.sep = originalSeparator;
   });
 
-  it('converts forward slashes in the RegExp to the OS specific path separator', () => {
+  it('proves we can write to path.sep for setting up the tests', () => {
     // $FlowFixMe: property sep is not writable.
     path.sep = '/';
+    expect(require('path').sep).toBe('/');
```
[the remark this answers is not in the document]
ryancat: The idea here is we assume `require('path')` should give the same path reference here or in the source code. This test is to make sure we can set the `path.sep` to some mock value like the line above, and when the source code requires it, it's using the mock value, not the value from the hosting OS.

ryancat:
> Thanks @ryancat for creating this PR and helping to significantly increase the test coverage of the `exclusionList`.
> 
> I've made a couple of comments. My main question is why the path where the user passes a string has to change as well. My understanding is that the main issue is the missing escaping of `path.sep` in the case where a `RegExp` is passed.

The string use case doesn't need to be changed for the regexp fix. That by itself is a new bug to me. My new test cases on strings will fail without that change.

I don't think we have test cases on strings before, and may not have use cases for string input. But IMO it's been broken as is.

packages/metro-config/src/defaults/__tests__/exclusionList-test.js
```diff
@@ -25,14 +25,84 @@ describe('exclusionList', () => {
...
+    // $FlowFixMe: property sep is not writable.
+    path.sep = '\\';
```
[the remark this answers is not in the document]
ryancat: I like that. Will update.

MichaReiser:
Thank you @ryancat  for elaborating. Regarding passing `strings`. My understanding is that the `string` case will only match the exact string and does not allow you to use a regex. E.g. `'.*[/\\]foo[/\\]bar']` from your test should exactly match the path `.*[/\\]foo[/\\]bar']`. The escaping makes sure that any characters requiring escaping are correctly escaped. 
That's why I believe that the old behavior was correct.

packages/metro-config/src/defaults/exclusionList.js
```diff
@@ -15,10 +15,19 @@ var list = [/website\/node_modules\/.*/, /.*\/__tests__\/.*/];
...
+    // We should replace all forward slashes to proper OS specific separators.
+    // The separator needs to be escaped in the regular expression source string,
+    // hence the '\\' prefix.
+    return pattern.source.replace(/\/|\\\//g, '\\' + path.sep);
   } else if (typeof pattern === 'string') {
-    var escaped = pattern.replace(/[\-\[\]\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
-    // convert the '/' into an escaped local file separator
+    // escape back slashes in regular expression so that when
+    // it's used in RegExp constructor, the back slashes are preserved.
+    // var escaped = pattern.replace(/[\-\[\]\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&');
```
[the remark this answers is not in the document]
ryancat: Yes, I can add a test case for this. The special characters are fine when creating regular expressions, the only thing that needs to be escaped is the escape character (`\`) itself.

For example, the original logic escaped all special characters in the string, so that the input string
```
"[/\\]foo"
```
 which is effectively "[/\]foo" after string escape, will become
```
"\[\/\\\]foo"
```
It then replaced all forward slashes to the OS specific separators, let's say is `\\` for windows:
```
"\[\\\\\\]foo"
```
Now, we are going to use this to construct a regular expression instance. This is clearly a wrong string to use, as `new RegExp("\[\\\\\\]foo")` is the same to `/[\\\]foo/`, which is giving `Invalid regular expression` error.

My change here will result in the correct result with this steps:
`"[/\\]foo"` which is `"[/\]foo"` after string escape --> `"[/\\]foo"` --> `"[\\\\]foo"` --> `new RegExp("[\\\\]foo")` --> `/[\\]foo/`

Which is expected.

packages/metro-config/src/defaults/exclusionList.js
ryancat: My comments weren't clear enough. Regular expression is confusing!

What I want to say here is the forward slash is what causing the trouble for our original code. If it's in brackets like `/[/\\]foo/`, in node v10, it will automatically be converted to `/[\/\\]foo/`. However,  in node v12, it will not be coverted at all (you can try that in https://www.jdoodle.com/ia/3tF).

If the forward slash is outside of brackets, as it will mess up with the regular expression start/end character, it will always need to be escaped. Hence, node v10 won't escape it again. That is to say, `/foo\/bar/` will not be converted in both node versions.

Hence, our code should properly handle both cases by the `|` operator. Since regular expression is greedy in matching patterns, it will automatically cover both cases. Assuming `const a = /foo\/bar/;` and `const b = /[/\\]foo/;`.

node 12:
```
a.source foo\/bar
a in windows foo\\bar
a in linux foo\/bar
b.source [/\\]foo
b in windows [\\\\]foo
b in linux [\/\\]foo
```

node 10
```
a.source foo\/bar
a in windows foo\\bar
a in linux foo\/bar
b.source [\/\\]foo
b in windows [\\\\]foo
b in linux [\/\\]foo
```

packages/metro-config/src/defaults/__tests__/exclusionList-test.js
ryancat: I like this! I will add to the test list.

packages/metro-config/src/defaults/__tests__/exclusionList-test.js
ryancat: I already have test cases cover that. See line 67-69

packages/metro-config/src/defaults/__tests__/exclusionList-test.js
ryancat: This test is for string input.

packages/metro-config/src/defaults/__tests__/exclusionList-test.js
ryancat: Give it another thought. They are actually not the same. 
```
const escaped = exclusionList(/[\/]/); // [\\\\]
const nonEscaped = exclusionList(/[/]/); // [\\]
```

However, for regular expression, `/[\\]/` is the same to `/[\\\\]/` as we are just repeating the options in the square brackets.

ryancat (commented):
> Thank you @ryancat for elaborating. Regarding passing `strings`. My understanding is that the `string` case will only match the exact string and does not allow you to use a regex. E.g. `'.*[/\\]foo[/\\]bar']` from your test should exactly match the path `.*[/\\]foo[/\\]bar']`. The escaping makes sure that any characters requiring escaping are correctly escaped.
> That's why I believe that the old behavior was correct.

Thanks for the details! That makes sense. I can update this and put some comments for future reference.

packages/metro-config/src/defaults/__tests__/exclusionList-test.js
```diff
@@ -25,14 +31,92 @@ describe('exclusionList', () => {
...
+    setPathSeperator('/');
+    expect(require('path').sep).toBe('/');
+    setPathSeperator('\\');
     expect(require('path').sep).toBe('\\');
-    expect('a\\b\\c').toMatch(exclusionList([new RegExp('a/b/c')]));
+  });
+
+  describe('simulate macOS/linux enviornment', () => {
+    it('converts forward slashes in the RegExp to the OS specific path separator in macOS/linux', () => {
+      setPathSeperator('/');
```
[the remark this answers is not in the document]
ryancat: Will update!

packages/metro-config/src/defaults/__tests__/exclusionList-test.js
```diff
@@ -25,14 +31,92 @@ describe('exclusionList', () => {
...
+      // Regular expression that already considered OS specific path separator.
+      // When explictly construct RegExp instance, the string needs to be escaped first.
+      // eg. /.*[/\\]foo[/\\]bar/ => '.*[/\\\\]foo[/\\\\]bar'
+      expect('/foo/bar').toMatch(
+        exclusionList([new RegExp('.*[/\\\\]foo[/\\\\]bar')]),
+      );
+      expect('/foo/bar').toMatch(exclusionList([/.*[/\\]foo[/\\]bar/]));
+    });
+
+    it('converts forward slashes in the string to the OS specific path separator in macOS/linux', () => {
```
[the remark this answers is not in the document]
ryancat: Will update!

packages/metro-config/src/defaults/__tests__/exclusionList-test.js
```diff
@@ -25,14 +31,92 @@ describe('exclusionList', () => {
...
+      setPathSeperator('/');
+      // Simple case
+      expect('a/b/c').toMatch(exclusionList([new RegExp('a/b/c')]));
+      expect('a/b/c').toMatch(exclusionList([/a\/b\/c/]));
+      // Regular expression that already considered OS specific path separator.
+      // When explictly construct RegExp instance, the string needs to be escaped first.
+      // eg. /.*[/\\]foo[/\\]bar/ => '.*[/\\\\]foo[/\\\\]bar'
+      expect('/foo/bar').toMatch(
+        exclusionList([new RegExp('.*[/\\\\]foo[/\\\\]bar')]),
+      );
```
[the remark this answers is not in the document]
ryancat: Will remove the comments for the details on adding escaped character. I put them here for future references on why I put `'.*[/\\\\]foo[/\\\\]bar'` in the `RegExp` constructor.

PR #606, opened 2020-11-07
Update terser to v5, support async minifiers

janicduplessis:
**Summary**

Update terser to the latest version that supports more modern js syntax like optional chaining. The main breaking change in v5 is that minify is now async. To support this I also needed to add async minifier support in metro-transform-worker.

**Test plan**

- Run tests
- Test building my app bundle with minifier terser and untransformed optional chaining

### DIFF
--- a/packages/metro-minify-terser/src/minifier.js
+++ b/packages/metro-minify-terser/src/minifier.js
@@ -15,8 +15,8 @@ const terser = require('terser');
 import type {BasicSourceMap} from 'metro-source-map';
 import type {MinifierResult, MinifierOptions} from 'metro-transform-worker';
 
-function minifier(options: MinifierOptions): MinifierResult {
-  const result = minify(options);
+async function minifier(options: MinifierOptions): Promise<MinifierResult> {
+  const result = await minify(options);
 
   if (!options.map || result.map == null) {
     return {code: result.code};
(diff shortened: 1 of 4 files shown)


packages/metro-minify-terser/src/minifier.js
```diff
@@ -50,11 +50,7 @@ function minify({
   /* $FlowFixMe(>=0.111.0 site=react_native_fb) This comment suppresses an
    * error found when Flow v0.111 was deployed. To see the error, delete this
    * comment and run Flow. */
-  const result = terser.minify(code, options);
-
-  if (result.error) {
```
janicduplessis, line 55: Terser will now throw an error so there is no need for additional handling anymore.

cpojer (approved):
Thank you 👍

cpojer:
We have a big integration with Metro and Buck internally that relies on this code being sync and changing every Flow to be async is a ton of work I unfortunately won't be able to get to.

The good news is that with Hermes we shouldn't need a minifier, so we may be able to clean up our internal integration soon.

janicduplessis:
I see, so you are using the metro-minify-terser package directly? What about landing the changes in metro-transform-worker only so it supports async minifiers? This would make it easier to just publish a terser-v5 minify plugin.

cpojer:
No, we don't actually use terser but we use the API synchronously in `metro-transform-worker` pretty heavily deep down a stack where changing everything to be async would be quite a bit of work. As said, with Hermes we may not need to support minifiers for much longer so maybe we can clean that up internally and then merge this.

janicduplessis:
Ah ok, the function in transform-worker is already async, but I guess it works since there are no await in it currently. Alright in this case let's just keep this open and we can revisit later when the internal infra changes.

rh389:
FYI, I’ve rebased this internally now that the groundwork @cpojer mentioned has been done behind the scenes of #754, hope to land it shortly.

Issue #400, opened 2019-05-02
postMinifyProcess and postProcessBundleSourcemap not working


devattendant:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
The documentation at [https://facebook.github.io/metro/docs/en/configuration](url) lists two parameters `postMinifyProcess` and `postProcessBundleSourcemap` which I both tried to use to hook into bundle creation. While both are documented, they are actually not implemented.

I have added calling `postProcessBundleSourcemap` in metro/src/server.js to get it working at least temporarily but I failed to get `postMinifyProcess` working as it is deleted from the config after a short time and not passed to the transformer, even if I remove the deletion at metro/src/DeltaBundler/Transformer.js.

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**
I've modified metro.config.js like the following, but there is no output:
`module.exports = {
	resetCache: true,
	transformer: {
		getTransformOptions: async () => ({
			transform: {
				experimentalImportSupport: false,
				inlineRequires: false,
			},
		}),
		minifierPath: 'metro-minify-terser',
		postMinifyProcess: ({ code, map }) => {
			console.log("postMinifyProcess")
			return { code, map };
		}
	}
};`

**What is the expected behavior?**
Both `postMinifyProcess` and `postProcessBundleSourcemap` should be called or removed in documentation.

**Please provide your exact Metro configuration and mention your Metro, node, yarn/npm version and operating system.**
Metro 0.51.1, RN 0.59.5, metro.config.js see above.

TheHighriser:
Same problem here. Still not working. Is there anything working in the Serializer config?

dominikfoldi:
We are building a JavaScript code optimizer and we want to create a Metro plugin which would be a good fit as the `postMinifyProcess`. It is really not implemented why is it the case?

vividh:
👍 
Please implement this!

EmilijusS:
Still not working.

robhogan:
Sorry folks - this was indeed broken at the time and the docs were wrong.

To close the loop, this particular extension point was then deleted completely in Metro 0.66.1. As @motiz88  said in 5d778de73:

> We can separately discuss redesigning and reintroducing this extension point, but it's likely not necessary: Users of Metro can already replace the minifier with their own implementation - e.g. one that calls the regular minifier and does any additional processing that would go in `postMinifyProcess`.

[...]

PR #608, opened 2020-11-19
Implement postProcessBundleSourcemap option

inokawa:
**Summary**

#400 

I implemented `PostProcessBundleSourcemap` option, which is on document but not in current code.
With this, we can modify processed code, which makes making plugins easier.

I took a look to the original code https://github.com/facebook/metro/commit/cd67cd47941a942509a360bac3d8a11c5ce070ff#diff-660c08c7ccb569e12d8a268bd8fa2011?w=1

I fixed `PostProcessBundleSourcemap` type to fit current code.

**Test plan**

I added `bundle-test.js`, to test postProcessBundleSourcemap works.

### DIFF
--- a/packages/metro-config/src/configTypes.flow.js
+++ b/packages/metro-config/src/configTypes.flow.js
@@ -38,13 +38,12 @@ export type PostMinifyProcess = ({
 };
 
 export type PostProcessBundleSourcemap = ({
-  code: Buffer | string,
+  code: string,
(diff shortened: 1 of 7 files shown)


robhogan:
Very belated thanks for the work here and apologies it didn't get a response. This extension point was eventually removed in favour of others that should serve the same use cases - see https://github.com/react/metro/issues/400#issuecomment-5602351768.

PR #624, opened 2021-01-13
Fix undefined assetDataPlugins issue in metro/Assets.js

stamp:
**Summary**

After upgrading from Expo SDK39 to SDK40 metro stopped loading my custom otf fonts with the error message "Cannot read property "length" of undefined. The fonts are loaded with the `useFonts` hook from [expo-font](https://github.com/expo/expo/tree/master/packages/expo-font) and `require` like this:
```
const [fontsLoaded] = useFonts({
  'SharpGrotesk18': require('./assets/fonts/SharpGroteskBook18.otf'),
  'SharpGrotesk19': require('./assets/fonts/SharpGroteskBold19.otf'),
  'TuskerGrotesk-4500': require('./assets/fonts/TuskerGrotesk4500Medium.otf'),
  'TuskerGrotesk-5700': require('./assets/fonts/TuskerGrotesk5700Bold.otf'),
  'TuskerGrotesk-7700': require('./assets/fonts/TuskerGrotesk7700Bold.otf'),
});
```

I traced the issue to https://github.com/facebook/metro/blob/master/packages/metro/src/Assets.js#L234 and by changing the line to 
```
if (!assetDataPlugins || !assetDataPlugins.length) {
  return assetData; 
}
```
everything worked as expected again. 

The changed line is 3 years old so I suppose the reason for seeing this  cause is probably not metro but some other part of the expo build process

### DIFF
--- a/packages/metro/src/Assets.js
+++ b/packages/metro/src/Assets.js
@@ -231,7 +231,7 @@ async function applyAssetDataPlugins(
   assetDataPlugins: $ReadOnlyArray<string>,
   assetData: AssetData,
 ): Promise<AssetData> {
-  if (!assetDataPlugins.length) {
+  if (!assetDataPlugins || !assetDataPlugins.length) {
     return assetData;
   }
 


motiz88 (changes requested):
Hey @stamp, thanks for this PR! I don't think we should add a null check on an parameter that is annotated with a non-nullable type (in this case `$ReadOnlyArray<string>`). I haven't looked at the details of this function yet, but if it makes sense to make the parameter nullable, let's make sure it's reflected in the type. Otherwise we may need to find the incorrect call site and fix the problem there.

stamp:
Yea, you are probably right! I'll do a try to find the call site and see how it goes.

PR #597, opened 2020-10-20
Add transformProfile `jsc-ios10.3`

newyankeecodeshop:
**Summary**

React Native now requires iOS 10, which includes a JavaScriptCore runtime that implements much of ES2015. It also bundles a modern JavaScriptCore on Android, which also implements ES2015 (I think it is roughly at the level of JSC in iOS 12).

This PR adds a new transform profile `jsc-ios10.3`, which removes all Babel transforms that are no longer needed on devices that are iOS 10.3+ or Android. This creates more optimal JS when running in these environments.

This PR is intended to replace the work done in PR #515 in a way that will not destabilize React Native users that use the default transform profile, or other options such as `unstable_disableES6Transforms`.

Fixes #359

**Test plan**

I added a unit test for the `getPreset()` function that tests that the appropriate plugins are removed when the transform profile is set to `jsc-ios10.3`.

I modified the `babel.config.js` in our project as follows:
```
module.exports = {
  presets: [
    [
      "module:metro-react-native-babel-preset",
      {unstable_transformProfile: "jsc-ios10.3"},
    ],
  ],
};
```
The bundle was generated with modern JS syntax such as `const` and `() => {}`.

PR #636, opened 2021-02-10
Remove preset plugins no longer required by JSC and Hermes

newyankeecodeshop:
**Summary**

React Native 0.64 now supports the following JavaScript engines:
1. JavaScriptCore >= iOS 11.x
2. JavaScriptCore `245459.0.0` on Android
3. Hermes 0.7+
4. Hermes "canary"

These runtimes all support ES2016 features. Therefore, there is a reduced need for Babel transforms to be applied which convert the generated JS to ES5.

Existing Hermes users are already filtering out the plugins I've removed, so this change will affects only JavaScriptCore users on iOS and Android.

**Test plan**

1. All unit tests pass.
2. Applied this change to my large React Native application, which includes several common 3rd-party libraries including `react-navigation`, `react-native-webview`, `react-native-gesture-handler`, etc. Started up the application and ran our automated test suite without issue.
3. Used `react-native bundle` to build a minified bundle without error.


packages/metro-react-native-babel-preset/src/configs/main.js
```diff
@@ -108,14 +102,6 @@ const getPreset = (src, options) => {
   if (isHermesCanary && (isNull || src.indexOf('async') !== -1)) {
```
newyankeecodeshop, line 102: Seems like this should be `if (! isHermesCanary && ...)`?

motiz88:
Hey @newyankeecodeshop, sorry for the delay in reviewing this. I suspect there will now be merge conflicts with @jsx's recent work here, b28047795b37da79867bd4b4914ede24e0238885, which moves generators to `hermes-stable`.

In general I'm not opposed to this PR, which seems correct, but please note that **`unstable_transformProfile` is an experimental Metro feature that might go away**. I think it needs to be redesigned to support specifying engine versions (a la `@babel/preset-env`) before we can support it as a stable feature in open source.

motiz88 (changes requested):
Requesting changes because of merge conflicts.

newyankeecodeshop:
Hi @motiz88 thanks for the feedback. Were you referring to my other PR #597, above? If so, I've closed that PR because, as you point out, it has issues with the new way the babel plugins are lazy loaded. It also looked like `unstable_transformProfile` wasn't going to be the best path forward.
That's why this PR is just doing a very small change, removing babel plugins that (1) Hermes doesn't use and (2) JavaScriptCore doesn't need either. 

What are the merge conflicts in this PR?

newyankeecodeshop:
Hi @motiz88, I updated this PR based on the recent commits that moved several more plugins to Hermes stable. Now 10 plugins are unnecessary for both Hermes and JavaScriptCore users. Yay!

motiz88 (approved):
Looks good, thanks @newyankeecodeshop!

packages/metro-react-native-babel-preset/src/configs/main.js
```diff
@@ -115,20 +102,6 @@ const getPreset = (src, options) => {
...
-    extraPlugins.push([
-      require('@babel/plugin-transform-exponentiation-operator'),
-    ]);
-  }
-  if (!isHermes && (isNull || src.indexOf('Object.assign')) !== -1) {
-    extraPlugins.push([require('@babel/plugin-transform-object-assign')]);
-  }
-  if (!isHermes && hasForOf) {
-    extraPlugins.push([
-      require('@babel/plugin-transform-for-of'),
```
motiz88, line 128: We have some internal tests that are now failing because the `for-of` transform is a prerequisite for `regenerator` (https://github.com/facebook/regenerator/issues/229). Have you looked into whether we can remove `regenerator` altogether for JSC? (We are already skipping it in `hermes-stable` and `hermes-canary`)

motiz88 (changes requested):

packages/metro-react-native-babel-preset/src/configs/main.js
ls-andrew-goodale: I believe we can - we should be able to use `async-to-generator` at least. I will try it out.

newyankeecodeshop: @motiz88 I was able to replace regenerator with `async-to-generator` and didn't have any issues with async code in our app. Code and tests updated. Thanks!

newyankeecodeshop:
@motiz88 Are the Facebook Internal tests still running? It seems like that build check is stuck.

newyankeecodeshop:
@motiz88 Any idea what's holding up the merge?

newyankeecodeshop:
@motiz88 any update on this merge?

aulneau:
@motiz88 seems like this has been in a good place for months, can this be approved and merged?

newyankeecodeshop:
@motiz88 checking in to see if there’s anything I can do for this PR. I’d love to know why the Facebook Internal tests have never completed or failed.

newyankeecodeshop:
@motiz88 should I close this PR? I'm guessing it's breaking some internal FB process?

newyankeecodeshop:
This work appears to be blocked so I'm closing this PR.

ariccio:
*mimes jokingly shaking fist at clouds*

Honestly, as a novice RN developer, it's surprising to me that in 2022 we still need to use regenerator runtime, and that async/await isn't some kind of natively supported first class feature. *Even C++ has coroutines now!* 

Why should we be fancypants javascript devs if we don't have first class support for async/await? ;)

chenyulun:
@newyankeecodeshop We are using RN version 0.64.4, copy your code and test it with our company intranet release, do you have any suggestions from your side? Thanks

newyankeecodeshop:
@chenyulun I think the latest versions of metro have removed many of the transformations so this may be less of an issue now. I am not doing React Native development currently, so I don't have any immediate suggestions for what you can do. Sorry, and good luck!

PR #642, opened 2021-03-03
Update troubleshooting docs for metro cache tmpdir

junoatscroll:
**Summary**

Use os default temp directory by checking POSIX $TMPDIR envvar for portability, with current `/tmpdir` as fallback.

For instance, on macos, this resolves to some long name like `/var/folders/yn/t9rvkpyj4b13s_gmtwgh1lw80000gn/T//metro-cache`

Reference:
- Metro default config uses node.js `os.tmpdir()` https://github.com/facebook/metro/blob/b2528d40c46085a32adbae370d0e47b6a24e3c8c/packages/metro-config/src/defaults/index.js#L122
- node.js `os.tmpdir()` returns system default https://nodejs.org/api/os.html#os_os_tmpdir
- system default is usually made available in $TMPDIR https://en.wikipedia.org/wiki/TMPDIR

**Test plan**

None


motiz88 (approved):
Makes sense, thanks @junoatscroll!

PR #643, opened 2021-03-04
Fix scope of function declaration binding in constant folding

leops:
**Summary**

This is a possibly incomplete fix for #291, it corrects at least one issue with name shadowing but there might be other problems present in the very large code payload presented in the original issue. The exact bug fixed here happens when a function declares an unused variable shadowing its own name:

```js
export function foo() {
    let foo;
}
```

The current code checks whether that the binding corresponding to the function is referenced in the scope of the FunctionDeclaration node, however in Babel that scope is situated inside the body of the function and not in the outer program. Since the binding is retrieved by name, Babel returns informations related to the inner, unreferenced variable and thus the entire function is removed. This fix changes the biding retrieval to read from the parent scope of the FunctionDefinition, which returns the correct binding.

**Test plan**

An additional test case has been added to verify the function remains in the final output.

### DIFF
--- a/packages/metro-transform-plugins/src/__tests__/constant-folding-plugin-test.js
+++ b/packages/metro-transform-plugins/src/__tests__/constant-folding-plugin-test.js
@@ -346,4 +346,20 @@ describe('constant expressions', () => {
 
     compare([constantFoldingPlugin], nonChanged, nonChanged);
   });
+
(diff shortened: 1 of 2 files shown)


motiz88 (approved):
Thank you @leops! This seems correct AFAICT. I appreciate the test, too.

PR #649, opened 2021-03-22
Use absolute path to project root as a component of cache key

maxkomarychev:
Problem in question was initially observed with `metro@0.58.0` and `react-native@0.63.4`

`getTransformCacheKey` produces same cache key for projects located
in different paths (while having all other configs equal).

When running several builds on CI computer I noticed strange failures
happening randomly which looked like this (every time with different
files):

```
error Unable to resolve module `/something/FOLDER_1/packages/app/src/reactotron-config` from `packages/app/src/app/init/loggers.ts`:
None of these files exist:
   * /something/FOLDER_1/packages/app/src/reactotron-config(.native|.ios.js|.native.js|.js|.ios.json|.native.json|.json|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx|.ios.svg|.native.svg|.svg)
   * /something/FOLDER_1/packages/app/src/reactotron-config/index(.native|.ios.js|.native.js|.js|.ios.json|.native.json|.json|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx|.ios.svg|.native.svg|.svg). Run CLI with --verbose flag for more details.
```

The problem here is that the process which caught this error was
launched from `/something/FOLDER_2/...` - two independent bundle
commands shared the cache.

This could also be reproduced locally when running two
`react-native bundle --reset-cache ...` instances from different folder.

I was able to fix this on our side by doing
`cacheVersion: "1.0-" + __dirname`  in `metro.config.js`

But I believe that cache components should account for an absolute path
of the project.

Previous commits introducing and then removing path to a project:

1. https://github.com/facebook/metro/commit/2322db4f819fde85bcb8867e1ef6ceb860b15ef4
2. https://github.com/facebook/metro/commit/c2dfe747b50457f870fe2afb79d0df6a0499f209

**Summary**

**Test plan**

### DIFF
--- a/packages/metro/src/DeltaBundler/getTransformCacheKey.js
+++ b/packages/metro/src/DeltaBundler/getTransformCacheKey.js
@@ -10,6 +10,7 @@
 
 'use strict';
 
+const path = require('path');
 const crypto = require('crypto');
 const getCacheKey = require('metro-cache-key');
 
@@ -42,6 +43,7 @@ function getTransformCacheKey(opts: {|
         'metro-cache',
         VERSION,
         opts.cacheVersion,
+        path.resolve(path.join(__dirname, '../../..'), opts.projectRoot),
         getCacheKey([require.resolve(transformerPath)]),
         transformerKey,
         transformerConfig.globalPrefix,


packages/metro/src/DeltaBundler/getTransformCacheKey.js
```diff
@@ -42,6 +43,7 @@ function getTransformCacheKey(opts: {|
         'metro-cache',
         VERSION,
         opts.cacheVersion,
+        path.resolve(path.join(__dirname, '../../..'), opts.projectRoot),
```
motiz88, line 46: I don't think we can assume that `../../..` is meaningful here, as the `metro` can be installed and used in different ways/locations. Even if we wanted to resolve `projectRoot` to an absolute path, we would have to do it in a way that is always correct.

motiz88 (changes requested):
Hey @maxkomarychev - I agree that having this cache key collision behaviour out of the box is a problem. Baking absolute paths into the cache keys is a bit of an extreme solution, as we do also support sharing the cache across machines using `NetworkStore`.

I think a more sustainable solutions would be to keep the relative paths in the keys, but automatically include some sort of configurable "project ID" as well. That project ID can default to the absolute path of the project root but would then be overridable for shared cache use cases.

packages/metro/src/DeltaBundler/getTransformCacheKey.js
maxkomarychev: can we use `process.cwd()` instead?

robhogan:
Closing as this stalled on changes requested from the author.

PR #659, opened 2021-04-10
docs/addDefaultFieldForResolverMainFields

Naturalclar:
**Summary**

added default value for resolverMainField config so user won't need to look in the code to find out the default value

**Test plan**
doc change.
verify that it matches the actual default value

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -115,7 +115,7 @@ An array of source extensions to include in the bundle. For example, if you woul
 
 #### `resolverMainFields`
 
-Type: `Array<string>` (default: ['react-native', 'browser', 'main'])
+Type: `Array<string>` (default: ['browser', 'main'])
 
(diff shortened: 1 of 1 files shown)


docs/Configuration.md
```diff
@@ -115,7 +115,7 @@ An array of source extensions to include in the bundle. For example, if you woul
 
 #### `resolverMainFields`
 
-Type: `Array<string>`
+Type: `Array<string>` (default: ['react-native', 'browser', 'main'])
```
Naturalclar, line 118:
```suggestion
Type: `Array<string>` (default: ['browser', 'main'])
```

docs/Configuration.md
```diff
@@ -115,7 +115,7 @@ An array of source extensions to include in the bundle. For example, if you woul
 
 #### `resolverMainFields`
 
-Type: `Array<string>`
+Type: `Array<string>` (default: ['browser', 'main'])
```
motiz88, line 118: I don't think we use this style of documenting defaults elsewhere in the docs. Would you mind mentioning this default in the descriptive text below?

motiz88 (changes requested):

PR #661, opened 2021-04-13
replace uglify-es with uglify-js

dulmandakh:
**Summary**

Replace uglify-es with uglify-js because of deprecation message from uglify-es

> support for ECMAScript is superseded by `uglify-js` as of v3.13.0

**Test plan**


Ashoat (approved):
Looks good to me, just a simple rename

radiosilence:
Should we be using uglify-js or terser?

PR #662, opened 2021-04-15
Update inspectorProxy with IPV6 localhost address

patrickgalbraith:
**Summary**

For IPV4 the webSocketDebuggerUrl defaults to `ws://localhost:8081` but for IPV6 this defaults to `ws://[::]:8081`. This causes issues since `ws://[::]` is not a valid websocket URL, I am guessing it should be `ws:[::1]` which is the IPV6 equivalent. 

**Test plan**

The webSocketDebuggerUrl should be `ws:[::1]` rather than `ws://[::]`.

This is related to this issue in Flipper https://github.com/facebook/flipper/issues/1520.


motiz88 (approved):
Looks good @patrickgalbraith, thanks for this!

PR #677, opened 2021-06-14
Add experimental transformer setting for disabling require() renaming

tido64:
**Summary**

Adds a flag, `unstable_disableModuleWrapping`, that disables wrapping of JS modules and transforming `require(...)` calls to `$$_REQUIRE(dependencyMap[n], ...)`. This is required for plugging in third-party bundlers such as [esbuild](https://esbuild.github.io/).

**Test plan**

A test has been added to ensure that `require()` calls are untouched.


tido64:
cc @motiz88

motiz88:
Looks good ! Can you share how you tested the `getGraphId` and `transformHelpers` changes? They do not seem to be covered by the `metro-transform-worker` unit test.

motiz88:
I'm unclear on why the new option needs to live in `getTransformOptions()` / `ExtraTransformOptions` which is used for "dynamic" config (i.e. options that can be set *during* bundling based on querying the dependency graph). It's a degree of freedom that we're trying to remove from Metro and which I don't believe is needed here.

Seems to me that this option should either 

1. be considered part of the bundle request, in which case you can add it to `parseOptionsFromUrl` and specify it as a URL param; or
2. be considered part of the "static" config i.e. moved from `JsTransformerOptions` to `JsTransformerConfig` and specified directly under the `transformer` key in `metro.config.js`.

I'm leaning towards (2) since it matches how you would configure the esbuild "serializer" (I assume) - you need to pass this flag *iff* you're using esbuild as your serializer, i.e. any extra degree of freedom to set this flag (either in the bundle request or `getTransformOptions`) could only lead to bugs.

tido64:
> I'm leaning towards (2) since it matches how you would configure the esbuild "serializer" (I assume) - you need to pass this flag _iff_ you're using esbuild as your serializer, i.e. any extra degree of freedom to set this flag (either in the bundle request or `getTransformOptions`) could only lead to bugs.

Thanks for explaining this to me. It makes a lot of more sense now. I wasn't sure what the difference was, but we definitely want (2). I've made the appropriate changes. Please have another look.

motiz88 (approved):

PR #679, opened 2021-06-16
Add transformOptions to each graph. Update tests.

afoxman:
**Summary**

Each graph is built using a specific set of transformOptions. Bind these to the graph, making them available to extensibility APIs such as `customSerializer` and `experimentalSerializerHook`. This data -- specifically, `platform` -- makes it possible to perform TypeScript validation during bundle/serve scenarios. `platform` is used to resolve types for platform-specific files.

**Test plan**

I updated existing Metro tests to validate the transformOptions in a graph.

I also wrote a sample application which bundles using Metro, and uses `experimentalSerializerHook`. I had my app print `platform` to the console, verifying that it was the correct value.

### DIFF
--- a/packages/metro/src/DeltaBundler.js
+++ b/packages/metro/src/DeltaBundler.js
@@ -21,6 +21,7 @@ import type {
   MixedOutput,
   Options,
 } from './DeltaBundler/types.flow';
+import type {TransformInputOptions} from './DeltaBundler/types.flow';
 
 export type {
   DeltaResult,
@@ -57,11 +58,17 @@ class DeltaBundler<T = MixedOutput> {
 
   async getDependencies(
     entryPoints: $ReadOnlyArray<string>,
+    transformOptions: TransformInputOptions,
     options: Options<T>,
   ): Promise<Dependencies<T>> {
     const depGraph = await this._bundler.getDependencyGraph();
 
-    const deltaCalculator = new DeltaCalculator(entryPoints, depGraph, options);
+    const deltaCalculator = new DeltaCalculator(
+      entryPoints,
+      depGraph,
+      transformOptions,
+      options,
+    );
 
     await deltaCalculator.getDelta({reset: true, shallow: options.shallow});
     const graph = deltaCalculator.getGraph();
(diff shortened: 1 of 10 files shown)


afoxman:
@motiz88

packages/metro/src/DeltaBundler.js
```diff
@@ -21,6 +21,7 @@ import type {
   MixedOutput,
   Options,
 } from './DeltaBundler/types.flow';
+import type {TransformInputOptions} from './DeltaBundler/types.flow';
```
acoates-ms, line 24: Nit: Add this to the import on the line above.

acoates-ms (approved):

motiz88:
Instead of explicitly threading `transformOptions` as a new argument through `getDependencies`, `buildGraph`, `DeltaCalculator`, etc - can we add it as a property to the `Options` type? Seems like we already pass that around to most of the call sites that care about `transformOptions`, and it makes sense semantically IMO. That would make for a smaller change that I'd be a bit more comfortable maintaining, especially since we don't have a test that captures what all this passing-`transformOptions`-around is _for_.

On that note, it's not ideal that we don't have an existing test for `experimentalSerializerHook` for you to update here. If we go with the less-invasive approach above, I think we can live with that for now (given that it's an experimental feature, and that this isn't a breaking change). But definitely feel free to write a test if you want to :)

afoxman:
Plumbing work is done. I took at look at adding tests, but I couldn't find a place to easily add one. Looks like we'd need a new test suite under /metro/src/__tests__/IncrementalBundler-test.js, which means it would need to cover the full IncrementalBundler class. I'm not that much of a Metro expert yet. :)

If you have advice on another place I can add a few experimentalSerializerHook tests, let me know.

packages/metro/src/DeltaBundler/types.flow.js
```diff
@@ -63,10 +65,20 @@ export type Module<T = MixedOutput> = {|
...
+    inlineRequires: boolean,
+    ...
+  },
+>;
+
 export type Graph<T = MixedOutput> = {|
   dependencies: Dependencies<T>,
   importBundleNames: Set<string>,
   +entryPoints: $ReadOnlyArray<string>,
+  +transformOptions: TransformInputOptions,
```
motiz88, line 81: Let's add a comment explaining why this is here.

```suggestion
  // Unused in core but useful for custom serializers / experimentalSerializerHook
  +transformOptions: TransformInputOptions,
```

motiz88 (approved):
lgtm with one minor comment.

Issue #645, opened 2021-03-12
ES Numeric Separators broken by babel (Unexpected token: name (_000))


westphalen:
**Do you want to request a *feature* or report a *bug*?**
Bug / improvement for `metro-react-native-babel-preset`

**What is the current behavior?**
```
Error: Unexpected token: name (_000) in file App.tsx at 16:25
```

A build error that makes little sense in the context of a larger code base. The line:col isn't true to the original code, I had to `console.log` from within `/metro/src/JSTransformer/worker.js` to find the problematic line.

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**
- `react-native init MyApp --template react-native-template-typescript`
- Use _ES Numeric Separators_ syntax in your code, e.g. `const thousand = 1_000;`
- Try to bundle the project for release

Minimal repository reproducing the issue: 
- `git clone https://github.com/westphalen/metro-babel-preset-numeric-separators`
- `yarn install`
- `yarn build`

**What is the expected behavior?**
Since the boilerplate project is set up to support TypeScript and es2017/esnext, one would expect that ES Numeric Separators were supported out of the box.

**Solution**
`yarn install @babel/plugin-proposal-numeric-separator`

Update `babel.config.js` with this line:
```
  plugins: ['@babel/plugin-proposal-numeric-separator'],
```

Would suggest this plugin is included by default in `metro-react-native-babel-preset`

**Please provide your exact Metro configuration and mention your Metro, node, yarn/npm version and operating system.**
- macOS 10.15.7
- `node -v` v10.22.0
- `"metro-react-native-babel-preset": "0.63.0"`
- `metro "0.58.0"`

motiz88:
Hey @westphalen, would you be interested in sending a PR that adds `@babel/plugin-proposal-numeric-separator` to the preset?

SConaway:
@motiz88 after having a weird issue where numeric separators were only an issue when building for iOS without Hermes (it works fine in `dev` on iOS and on Android in `dev` and `prod`), i'd be willing to make a super-quick PR to do that. Would it be accepted? 

Should it go in https://github.com/facebook/metro/blob/master/packages/metro-react-native-babel-preset/src/passthrough-syntax-plugins.js or https://github.com/facebook/metro/blob/master/packages/metro-react-native-babel-preset/src/configs/main.js?

motiz88:
[...]

PR #681, opened 2021-07-05
Add `@babel/plugin-proposal-numeric-separator` to the babel plugins when not in Hermes

SConaway:
**Summary**

This PR solves issue #645. It adds `@babel/plugin-proposal-numeric-separator` to the list of babel plugins when not using Hermes to more closely match ES2021. Now that ES2021 has been finalized, people will probably (?—I know I did) expect `1_000_000` to be treated as a valid number.  

As I noted in a [comment](https://github.com/facebook/metro/issues/645#issuecomment-873670023) in #645, should `@babel/preset-env` be included instead of including many different plugins manually? Even for engines like Hermes which don't need all the same transforms as JSC, I don't think the unneeded transforms would drastically increase the bundle size, right? 

**Test plan**
Tested it using a sample app on `"react-native": "0.65.0-rc.2"`. Solves the `react-native bundle` issue on iOS (only tested on device as I encountered issues compiling for the simulator) and Android production / release builds when not using Hermes.

### DIFF
--- a/packages/metro-react-native-babel-preset/package.json
+++ b/packages/metro-react-native-babel-preset/package.json
@@ -22,6 +22,7 @@
     "@babel/plugin-proposal-class-properties": "^7.0.0",
     "@babel/plugin-proposal-export-default-from": "^7.0.0",
     "@babel/plugin-proposal-nullish-coalescing-operator": "^7.0.0",
+    "@babel/plugin-proposal-numeric-separator": "^7.0.0",
     "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
     "@babel/plugin-proposal-optional-catch-binding": "^7.0.0",
     "@babel/plugin-proposal-optional-chaining": "^7.0.0",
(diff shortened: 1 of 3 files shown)


motiz88 (approved):
Looks good to me.

kelset:
@motiz88 any updates on the state of this one?

chriswiggins:
Until this is merged, the current workaround is outlined here: https://github.com/facebook/metro/issues/645#issuecomment-1144385337

aamagda:
Hi, do you plan to merge this request?

LinusU:
@motiz88 is there any updates on this? Would love it if this could land ☺️

huntie:
@kelset @aamagda @LinusU This PR is all good to go, however at this point we'll need to wait until after Meta's internal holiday code freeze. Changes should land in very early 2023, apologies! 🙏🏻

robhogan:
I'm not sure this is actually necessary any more - the error came from `uglify-es` minifier, which is no longer our default and is due for removal. Since the switch to `terser`, I believe we support numeric separators all the way through the stack. 

A bit more on how you can use Terser with an older Metro version here: https://github.com/facebook/metro/issues/645#issuecomment-1359307444

PR #696, opened 2021-08-05
Add a `metro` script so Metro can be used in development.

cpojer:
**Summary**
There is no way to run Metro in development outside of Facebook unless somebody is prepared to go through `yarn build; yarn lerna prepare-release` and `yarn cleanup-release` cycles manually every time when making a code change. This PR adds a script `metro` that can be invoked like `./metro serve` to start a development server of Metro running from source. This is helpful for anyone not working at Facebook who wants to use the master version of Metro to do verify changes.

**Test plan**
* Run `./metro serve`, it works!


package.json
```diff
@@ -31,6 +31,9 @@
     "jest": "^26.6.3",
     "lerna": "2.4.0",
     "micromatch": "^4.0.2",
+    "metro": "*",
+    "metro-config": "*",
+    "metro-babel-register": "*",
```
cpojer, line 36: This is necessary because otherwise eslint will complain. This change has no other effect since those modules are already linked as part of Yarn's workspaces.

metro
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env node
+'use strict';
+
+const path = require('path');
+
+require('metro-babel-register')([path.join(__dirname, 'packages/')]);
+require('metro-config/src/defaults').setDefaultWorkerPath(
+  path.join(__dirname, 'devTransformWorker.js'),
+);
```
cpojer, line 9: This is necessary because `metro-babel-register` needs to be set up for the worker processes too. You will find the same somewhere at Facebook in `xplat/js/tools` (?) if you need to double check.

motiz88: Hmm, I wish we didn't need to add an ad hoc setter to `metro-config/src/defaults` like this. Any chance we can get away with creating a dedicated config file for use in this dev scenario?

cpojer: This change is not for running Metro on itself, it's for running Metro from source for other projects, like for example React Native, which have their own configuration. There is no way to specify two configs that are merged into each other without extending the CLI, and manually detecting whether Metro is run from source and embedding this config there doesn't seem right.

The way this was done in the past is by having a cli around Metro's cli that knows when it is being run from source, but that's not really an option here as I don't want to bring in a custom CLI just for running from source and overwriting `config.transformer.transformWorkerPath` after loading the configuration. Also, if we used the approach of setting the worker after loading the config, it would overwrite custom transform worker implementations that people might use.

I'm open to better solutions if you have other ideas.

motiz88:
Closing in favour of #823. Thanks @cpojer for the initial idea and implementation!

PR #700, opened 2021-08-26
Add stubs for experimental dev-mode bundle splitting to asyncRequire

motiz88:
Summary: Adds stubs that allow Metro bundles to error gracefully if `experimentalImportBundleSupport: true` is used without swapping out `asyncRequire` for a compatible implementation.

Differential Revision: D30579830


robhogan:
Superseded by ac3adced4 and 4ef14f9a2 - the experiment shipped as lazy bundling

PR #701, opened 2021-08-31
Separate package roots from watch folders

afoxman:
**Summary**

When Metro runs, it gets the list of package root paths from the config property `watchFolders`. This means all packages, in-repo and external, must be watched for changes. In large mono-repos, this causes problems when thousands of external packages are used. File-watcher limits are hit, and a lot of CPU, memory and disk resources are used.

This PR adds a new `roots` Metro config property to explicitly separate package roots from watch folders. When `roots` is missing or empty, it is initialized using `watchFolders` for backward compatibility.

**Test plan**

Update tests to include validation for `roots`.

Test a build of this PR locally with a separate project that does bundling and serving using Metro. Verify that setting `roots` and `watchFolders` to different values works. Additionally, verify that when `roots` is not set, it defaults to using `watchFolders`.

### DIFF
--- a/packages/metro/src/node-haste/DependencyGraph.js
+++ b/packages/metro/src/node-haste/DependencyGraph.js
@@ -133,7 +133,7 @@ class DependencyGraph extends EventEmitter {
       retainAllFiles: true,
       resetCache: config.resetCache,
       rootDir: config.projectRoot,
-      roots: config.watchFolders,
+      roots: config.roots,
       throwOnModuleCollision: true,
       useWatchman: config.resolver.useWatchman,
       watch: watch == null ? !ci.isCI : watch,
--- a/packages/metro-config/src/__tests__/__snapshots__/loadConfig-test.js.snap
+++ b/packages/metro-config/src/__tests__/__snapshots__/loadConfig-test.js.snap
@@ -69,6 +69,9 @@ Object {
     ],
     "useWatchman": true,
   },
+  "roots": Array [
+    "/",
+  ],
   "serializer": Object {
     "createModuleIdFactory": [Function],
     "customSerializer": null,
@@ -212,6 +215,9 @@ Object {
     ],
     "useWatchman": true,
   },
+  "roots": Array [
+    "/",
+  ],
   "serializer": Object {
     "createModuleIdFactory": [Function],
     "customSerializer": null,
@@ -355,6 +361,9 @@ Object {
     ],
     "useWatchman": true,
   },
+  "roots": Array [
+    "/",
+  ],
   "serializer": Object {
     "createModuleIdFactory": [Function],
     "customSerializer": null,
@@ -498,6 +507,9 @@ Object {
     ],
     "useWatchman": true,
   },
+  "roots": Array [
+    "/",
+  ],
   "serializer": Object {
     "createModuleIdFactory": [Function],
     "customSerializer": null,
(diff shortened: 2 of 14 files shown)


tido64:
cc @motiz88

motiz88:
I'd like to better understand the use case, and make sure we've considered alternative solutions, before introducing a new option/concept for this.

My current stance: I believe Metro's defaults _should_ be to watch everything that could go into a build, since otherwise it would provide a weaker guarantee that you're actually running the latest code for your project. That means I believe watching should fundamentally be opt-out and not opt-in - so even if we go with introducing an option to directly control it, we'll probably want to tweak the design.

With regards to scaling file watching to large repos, have you tried Watchman? That's probably the first alternative to discuss.

packages/metro/src/node-haste/DependencyGraph.js
```diff
@@ -133,7 +133,7 @@ class DependencyGraph extends EventEmitter {
       retainAllFiles: true,
       resetCache: config.resetCache,
       rootDir: config.projectRoot,
-      roots: config.watchFolders,
+      roots: config.roots,
```
motiz88, line 136: If we specify both `roots` and `watchFolders`, `roots` will be watched and `watchFolders` won't be. Is this what you intended?

tido64:
> I'd like to better understand the use case, and make sure we've considered alternative solutions, before introducing a new option/concept for this.

Essentially, we want to separate what are considered roots and what needs to be watched for changes. In big monorepos, we can very quickly hit file node limitations if we simply watch the repo root because `<repo>/node_modules` is so big. But if we don't include it, Metro will be unable to find `react-native` and other installed packages. By separating the two, we can set root to be `<repo>/` and tell Metro to watch the individual packages under `<repo>/packages/*`. This way we avoid watching all of `<repo>/node_modules`.

@afoxman correct me if I miss anything.

packages/metro/src/node-haste/DependencyGraph.js
afoxman: My intent was to build the haste map around the roots so they'd be found by the resolver. I thought watching happened elsewhere. I didn't know the haste map also handled watching.

It sounds like this needs to go deeper, then, into the haste map code. Do you have pointers/advice on how I can expand the haste map library to support this? I haven't looked at it before.

afoxman:
> > I'd like to better understand the use case, and make sure we've considered alternative solutions, before introducing a new option/concept for this.
> 
> Essentially, we want to separate what are considered roots and what needs to be watched for changes. In big monorepos, we can very quickly hit file node limitations if we simply watch the repo root because `<repo>/node_modules` is so big. But if we don't include it, Metro will be unable to find `react-native` and other installed packages. By separating the two, we can set root to be `<repo>/` and tell Metro to watch the individual packages under `<repo>/packages/*`. This way we avoid watching all of `<repo>/node_modules`.
> 
> @afoxman correct me if I miss anything.

This is all correct. I'll add that with pnpm monorepos, you end up needing to add thousands of package roots due to the way packages are organized. This overwhelms the OS file watching systems, and runs into limitations. It's also a huge drain on OS resources to watch external packages that aren't likely to change -- at least not without restarting the dev server.

Regarding opt-in, I was careful to have `roots` be set to `watchFolders` when it wasn't explicitly specified. That way all existing users of Metro would be unaffected, and only devs that set `roots` would get the separation.

afoxman:
Let's pause this PR while Tommy and I look into jest-haste-map and see if separating roots from watchFolders is even possible. Not sure if it is with watchman.

motiz88:
> with pnpm monorepos, you end up needing to add thousands of package roots due to the way packages are organized

I'm not familiar with pnpm - could you elaborate on why this is the case? Is there a way to add a single watch root instead and use exclusion patterns to control what subfolders get pulled in?

Again, my first recommendation when it comes to scaling file watching to large repos is to try out Watchman, which _should_ perform better than `jest-haste-map`'s built-in watcher out of the box.

PR #702, opened 2021-09-10
fix(metro-config): Pass absolute default worker path to metro

asmundg:
**Summary**

If we pass an unresolved path to metro, metro's worker pool,
jest-worker, will need to resolve the worker path itself. This relies on
jest-worker being able to resolve metro, even though it doesn't
explicitly depend on metro (doing so would result in a nice cyclic
dependency).

In package managers that eliminate phantom dependencies, like Yarn PnP
or pnpm, this resolution just fails.

To avoid relying on a phantom dependency, we need to resolve the
absolute path of the worker before handing it off to jest-worker. Since
_we_ depend on metro, this is fine, and jest-worker can consume the
absolute path.

**Test plan**

This is based on the assumption that jest-worker's BaseWorkerPool accepts a resolved worker path: https://github.com/facebook/jest/blob/e0b33b74b5afd738edc183858b5c34053cfc26dd/packages/jest-worker/src/base/BaseWorkerPool.ts#L35.


motiz88 (approved):
Looks good, thanks @asmundg! Can you fix the merge conflicts?

asmundg:
@motiz88, Done!

robhogan:
This was implemented (a little differently, but fixing the same issue) in https://github.com/facebook/metro/commit/6d46078e74ae9a43aa90bed46dbd6610e2696cd0

PR #714, opened 2021-10-18
Require babel helpers instead of inlining

oblador:
**Summary**

It's unclear exactly why or when this regression was introduced, but in my project somewhere when going from babel 7.12.x to 7.15.x due to an upgrade of the `metro-react-native-babel-preset` package our bundle size ballooned with ~8%. Upon inspecting the generated code I found that `@babel/runtime` helper functions were littered all over the place. This also happens in a freshly created `react-native` project, I've tried with 0.64.2, 0.65.1 and 066.1

To avoid the inlining/duplication this PR simply passes the [`version`](https://babeljs.io/docs/en/babel-plugin-transform-runtime#version) option of whatever version of `@babel/runtime` is installed. 

For a `react-native init` project, this reduces the JS bundle size from 756K to 675K = 10,7% reduction. 

**Test plan**

```bash
npx react-native init Example
cd Example
yarn react-native bundle --platform ios --dev false --entry-file ./index.js --bundle-output bundle.js --minify false --reset-cache

grep -o 'function _getRequireWildcardCache' bundle.js | wc -l
-> before=216; after=2;

du -h bundle.js
-> before=1.7M; after=1.5M;
```

### DIFF
--- a/packages/metro-react-native-babel-preset/package.json
+++ b/packages/metro-react-native-babel-preset/package.json
@@ -56,6 +56,7 @@
     "@babel/plugin-transform-template-literals": "^7.0.0",
     "@babel/plugin-transform-typescript": "^7.5.0",
     "@babel/plugin-transform-unicode-regex": "^7.0.0",
+    "@babel/runtime": "^7.0.0",
     "@babel/template": "^7.0.0",
     "react-refresh": "^0.4.0"
   },
--- a/packages/metro-react-native-babel-preset/src/configs/main.js
+++ b/packages/metro-react-native-babel-preset/src/configs/main.js
@@ -162,6 +162,7 @@ const getPreset = (src, options) => {
       {
         helpers: true,
         regenerator: !isHermes,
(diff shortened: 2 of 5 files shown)


motiz88:
Hi @oblador, thanks for the PR! I completely agree that this is something we should improve, but I'm not sure the approach here is entirely correct.

The `require` statements injected by `@babel/transform-helpers` are resolved relative to the *project* being built with Metro, while the `require('@babel/runtime/package.json')` statement you added is resolved relative to `metro-react-native-babel-preset`. In the worst case, the project might have an older version of the helpers installed, but Metro's transformer will now assume a newer one, and emit code that only breaks at runtime.

To avoid breaking apps, we should guarantee that the version of `@babel/transform-helpers` we *actually* resolve to is semver-compatible with the version we assumed it would be in the transformer. Without getting too deeply into solution design here - I think solving this might require us to retain more metadata about injected helpers (and their expected versions) in the transformer's output.

PR #727, opened 2021-11-03
Fix SyntaxError bundling TypeScript class with `declare` property

newyankeecodeshop:
**Summary**

After upgrading Babel to 7.14.5 in my TypeScript-based React Native project, I got the following error when building the bundle:
```
SyntaxError: /Users/.../myproject/src/components/contexts.tsx: The 'declare' modifier is only allowed when the 'allowDeclareFields' option of @babel/plugin-transform-flow-strip-types or @babel/preset-flow is enabled.
  36 |     public static contextType = ReactAppContext;
  37 |
> 38 |     declare context: React.ContextType<typeof ReactAppContext>;
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  39 |
  40 |     public render() {
  41 |       /* @ts-ignore-next */
    at File.buildCodeFrameError (/Users/.../myproject/node_modules/@babel/core/lib/transformation/file/file.js:250:12)
    at NodePath.buildCodeFrameError (/Users/.../myproject/node_modules/@babel/traverse/lib/path/index.js:139:21)
    at /Users/.../myproject/node_modules/@babel/plugin-transform-flow-strip-types/lib/index.js:100:29
    at Array.forEach (<anonymous>)
    at PluginPass.Class (/Users/.../myproject/node_modules/@babel/plugin-transform-flow-strip-types/lib/index.js:93:31)
    at newFn (/Users/.../myproject/node_modules/@babel/traverse/lib/visitors.js:177:21)
    at NodePath._call (/Users/.../myproject/node_modules/@babel/traverse/lib/path/context.js:53:20)
    at NodePath.call (/Users/.../myproject/node_modules/@babel/traverse/lib/path/context.js:40:17)
    at NodePath.visit (/Users/.../myproject/node_modules/@babel/traverse/lib/path/context.js:100:31)
```

This occurs because of a change to the `plugin-transform-flow-strip-types` plugin introduced in version 7.9.0. Since the file being bundled is TypeScript, I thought a solution would be to exclude the flow plugin from TypeScript files, and instead rely on `plugin-transform-typescript` to strip the TypeScript type declarations. This will help prevent future issues when flow and Typescript may diverge in their syntax and features.

**Test plan**

I updated the TypeScript test file in the unit test to have a `declare` property. 

I also had to change the root babel config for metro to exclude TypeScript files because it's configuration of `plugin-transform-flow-strip-types` was overriding the configuration from the `babel-preset-react-native`. I expect this is because the unit tests are inheriting the project babel config, but I am not sure how to disable that. Excluding TypeScript files seemed like a reasonable, low-impact way to fix the issue.

### DIFF
--- a/packages/metro-react-native-babel-preset/src/configs/main.js
+++ b/packages/metro-react-native-babel-preset/src/configs/main.js
@@ -20,6 +20,15 @@ function isTSXSource(fileName) {
   return !!fileName && fileName.endsWith('.tsx');
 }
 
+function notTypeScriptSource(fileName) {
+  return !isTypeScriptSource(fileName) && !isTSXSource(fileName);
+}
+
+function optionsForTransformTS(opts) {
+  // `allowDeclareFields` will be enabled by default in Babel 8
+  return {...opts, allowDeclareFields: true, allowNamespaces: true};
+}
+
 const defaultPluginsBeforeRegenerator = [
   [require('@babel/plugin-syntax-flow')],
   [require('@babel/plugin-transform-block-scoping')],
@@ -173,24 +182,15 @@ const getPreset = (src, options) => {
       // the flow strip types plugin must go BEFORE class properties!
       // there'll be a test case that fails if you don't.
       {
+        test: notTypeScriptSource,
         plugins: [require('@babel/plugin-transform-flow-strip-types')],
-      },
-      {
-        plugins: [
-          ...defaultPluginsBeforeRegenerator,
-          isHermes ? null : require('@babel/plugin-transform-regenerator'),
-          ...defaultPluginsAfterRegenerator,
-        ].filter(Boolean),
       },
       {
         test: isTypeScriptSource,
         plugins: [
           [
             require('@babel/plugin-transform-typescript'),
-            {
-              isTSX: false,
-              allowNamespaces: true,
-            },
+            optionsForTransformTS({isTSX: false}),
           ],
         ],
       },
@@ -199,12 +199,16 @@ const getPreset = (src, options) => {
         plugins: [
           [
             require('@babel/plugin-transform-typescript'),
-            {
-              isTSX: true,
-              allowNamespaces: true,
-            },
+            optionsForTransformTS({isTSX: true}),
           ],
         ],
+      },
+      {
+        plugins: [
+          ...defaultPluginsBeforeRegenerator,
+          isHermes ? null : require('@babel/plugin-transform-regenerator'),
+          ...defaultPluginsAfterRegenerator,
+        ].filter(Boolean),
       },
       {
         plugins: extraPlugins,
--- a/packages/metro-react-native-babel-preset/src/configs/main.js
+++ b/packages/metro-react-native-babel-preset/src/configs/main.js
@@ -20,17 +20,13 @@ function isTSXSource(fileName) {
   return !!fileName && fileName.endsWith('.tsx');
 }
 
-function notTypeScriptSource(fileName) {
-  return !isTypeScriptSource(fileName) && !isTSXSource(fileName);
-}
-
-function optionsForTransformTS(opts) {
+const COMMON_TS_OPTIONS = {
   // `allowDeclareFields` will be enabled by default in Babel 8
-  return {...opts, allowDeclareFields: true, allowNamespaces: true};
-}
+  allowDeclareFields: true,
+  allowNamespaces: true,
+};
 
 const defaultPluginsBeforeRegenerator = [
-  [require('@babel/plugin-syntax-flow')],
   [require('@babel/plugin-transform-block-scoping')],
   [
     require('@babel/plugin-proposal-class-properties'),
@@ -182,7 +178,7 @@ const getPreset = (src, options) => {
       // the flow strip types plugin must go BEFORE class properties!
       // there'll be a test case that fails if you don't.
       {
-        test: notTypeScriptSource,
+        test: file => !isTypeScriptSource(file) && !isTSXSource(file),
         plugins: [require('@babel/plugin-transform-flow-strip-types')],
       },
       {
@@ -190,7 +186,7 @@ const getPreset = (src, options) => {
         plugins: [
           [
             require('@babel/plugin-transform-typescript'),
-            optionsForTransformTS({isTSX: false}),
+            {isTSX: false, ...COMMON_TS_OPTIONS},
           ],
         ],
       },
@@ -199,7 +195,7 @@ const getPreset = (src, options) => {
         plugins: [
           [
             require('@babel/plugin-transform-typescript'),
-            optionsForTransformTS({isTSX: true}),
+            {isTSX: true, ...COMMON_TS_OPTIONS},
           ],
         ],
       },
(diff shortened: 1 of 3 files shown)


packages/metro-react-native-babel-preset/src/configs/main.js
```diff
@@ -20,6 +20,15 @@ function isTSXSource(fileName) {
   return !!fileName && fileName.endsWith('.tsx');
 }
 
+function notTypeScriptSource(fileName) {
+  return !isTypeScriptSource(fileName) && !isTSXSource(fileName);
+}
```
motiz88, line 25: Nit: I would probably inline this as an arrow function below.

packages/metro-react-native-babel-preset/src/configs/main.js
```diff
@@ -20,6 +20,15 @@ function isTSXSource(fileName) {
...
+
+function optionsForTransformTS(opts) {
+  // `allowDeclareFields` will be enabled by default in Babel 8
+  return {...opts, allowDeclareFields: true, allowNamespaces: true};
+}
```
motiz88, line 30: Nit: I would instead create a constant e.g.
```
const COMMON_TS_OPTIONS = {
  // `allowDeclareFields` will be enabled by default in Babel 8
  allowDeclareFields: true,
  allowNamespaces: true,
};
```

and spread it into the usages below.

Also, do you have a reference for why Babel is treating this as a breaking change? Could this affect Metro users who for some reason need to not use this syntax yet?

packages/metro-react-native-babel-preset/src/configs/main.js
```diff
@@ -199,13 +199,17 @@ const getPreset = (src, options) => {
...
           ],
         ],
       },
+      {
+        plugins: [
+          ...defaultPluginsBeforeRegenerator,
+          isHermes ? null : require('@babel/plugin-transform-regenerator'),
+          ...defaultPluginsAfterRegenerator,
+        ].filter(Boolean),
+      },
```
motiz88, line 212: Can you elaborate on why this needed to move below the TypeScript passes?

packages/metro-react-native-babel-preset/src/configs/main.js
```diff
@@ -20,6 +20,15 @@ function isTSXSource(fileName) {
...
+  return !isTypeScriptSource(fileName) && !isTSXSource(fileName);
+}
+
+function optionsForTransformTS(opts) {
+  // `allowDeclareFields` will be enabled by default in Babel 8
+  return {...opts, allowDeclareFields: true, allowNamespaces: true};
+}
+
 const defaultPluginsBeforeRegenerator = [
   [require('@babel/plugin-syntax-flow')],
```
motiz88, line 33: Should we also make this conditional on `notTypeScriptSource`? (e.g. move it into the same `plugins` block as `flow-strip-type`)

packages/metro-react-native-babel-preset/src/configs/main.js
newyankeecodeshop: "why Babel is treating this as a breaking change?" - I'm not sure what you're asking. Here is the PR that adds the change to the Flow plugin: https://github.com/babel/babel/pull/11178
I think the reason it's breaking is that the default for `allowDeclareFields` is false. I'm not sure why it defaulted to that.

"could this affect Metro users...?" - I don't think it will. I tried to keep this low impact by (1) keeping the flow plugin in use for all files that are not `.ts[x]`, but switching TypeScript files to use the TypeScript transform instead of the Flow transform will cause some behavior changes. I hope that these changes are more desirable since using a Flow parser on TypeScript files could cause other unforeseen problems.

packages/metro-react-native-babel-preset/src/configs/main.js
newyankeecodeshop: Yes, I think it's for a similar reason that the comment above the flow plugin mentions. If the TS transforms are not first, the tests fail with the following error:
```
SyntaxError: TypeScript.ts: /Users/andrew/Development/Libraries/metro/TypeScript.ts: TypeScript 'declare' fields must first be transformed by @babel/plugin-transform-typescript.
    If you have already enabled that plugin (or '@babel/preset-typescript'), make sure that it runs before any plugin related to additional class features:
     - @babel/plugin-proposal-class-properties
     - @babel/plugin-proposal-private-methods
     - @babel/plugin-proposal-decorators
```
It seemed like a good change to have the metro preset do the type removal properties first, for both Flow and TypeScript, so that the overall behavior is consistent between .js and .ts files.

packages/metro-react-native-babel-preset/src/configs/main.js
newyankeecodeshop: Yes, that's probably a good idea.

newyankeecodeshop: Actually I wonder if the plugin is even necessary? Since the `transform-flow-strip-types` already includes it for parsing flow syntax, and that plugin is earlier in the configuration, this plugin is probably not actually doing anything.

packages/metro-react-native-babel-preset/src/configs/main.js
motiz88: > I think the reason it's breaking is that the default for `allowDeclareFields` is false

I guess my question is the other way around: presumably, the default for `allowDeclareFields` is false because the Babel team must have judged it to be a breaking change. So is there a reason for us to move separately from Babel on this? IMO, we should make it possible for people to configure Babel explicitly [per the Babel recommentation](https://babeljs.io/blog/2019/11/05/7.7.0#:~:text=it%20is%20recommended%20that%20you%20migrate%20your%20config%20to%20use%20it), and only switch to Babel 8's defaults when Babel 8 actually ships.

I fully agree that we should not be trying to run a Flow plugin on TypeScript code - that part is purely a bug.

newyankeecodeshop: I believe I've found the breaking change thanks to your link to that PR. I found this PR for the original TypeScript feature: https://github.com/babel/babel/pull/10546
It would appear that `allowDeclareFields` affects the JS emitted for uninitialized fields. I was able to verify using the metro unit tests that given TypeScript:
```
class B {
  field1: string;
}
```
The code emitted in the class constructor with `allowDeclareFields: true` is
`this.field1=void 0`
whereas with `allowDeclareFields: false`, no code is generated for that statement.

If we want to enable Metro users to enable `allowDeclareFields` for TypeScript and/or Flow, will that require adding a new metro-specific option, or can a user simply add a reference to the plugin in their own `babel.config.js`?

newyankeecodeshop:
@motiz88 I updated the PR to only change the file filters so that the flow plugin will be excluded from TypeScript files. This will enable RN users to add the following to their babel config:
```
overrides: [{
  test: /\.tsx/,
  plugins: [
    [
      require('@babel/plugin-transform-typescript'),
      {isTSX: true, allowDeclareFields: true, allowNamespaces: true},
    ],
  ]
```

Issue #737, opened 2021-11-09
Metro always resolves packages at the root of the directory when using the `nodeModulesPaths` option


sharmilajesupaul:
**Do you want to request a *feature* or report a *bug*?**
report a bug

**What is the current behavior?**

*I believe this is a potential bug, but I'm unsure if it is due to the configuration I'm using.* It seems like Metro doesn't respect duplicated transitive dependencies of packages when using the `nodeModulesPaths` option.

Meaning, when the same package exists at the root of an additional node_modules directory & inside of another package. Metro will attempt to resolve packages at the root of the additional node_modules directory first.

This differs from the way NodeJS handles `NODE_PATH` resolution where it will attempt to resolve the transitive dependency first.

The way that I came across this was a mismatch in the `ts-invariant` package which was required by our required version of `@apollo/client@3.4` at `v0.9.3`. But the root `node_modules` had `ts-invariant@0.4.4`.

The `@apollo/client` package goes on to call `setVerbosity` which fails and throws an error, since `setVerbosity` was introduced in the newer version of `ts-invariant` that `@apollo/client@3.4` had an explicit dependency on.

i.e.

```sh
.
├── some_cached_node_modules
│   ├── @apollo/client
│   │   ├── node_modules
│   │   │   └── ts-invariant (0.9.3) # <-- this is the package that should be used
│   │   └── package.json
│   └── ts-invariant (0.4.4) # <-- this is the package included in the bundle
└── project-root
    ├── index.js # imports @apollo/client & does things
    ├── package.json
    └── yarn.lock
```

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**

See [this repo](https://github.com/sharmilajesupaul/metro-node-modules-test-case ).  To see reproduce this bug run: `cd project-root; yarn install; yarn run metroBuildExample`. Also, see [the README.md](https://github.com/sharmilajesupaul/metro-node-modules-test-case#readme) for more information.

**What is the expected behavior?**

I expect the behavior to match how NODE_PATH resolution is done, which is also how the resolver's `nodeModulesPaths` field is described in the metro configuration.

From the configuration docs:

> This option can be used to add additional node_modules folders for Metro to locate third-party dependencies when resolving modules. This is useful if third-party dependencies are installed in a different location outside of the direct path of source files.
>
> This option works similarly to how $NODE_PATHS works for Node.js based tooling.

[...]

PR #738, opened 2021-11-10
[metro-resolver] Respect transitive deps when using `nodeModulesPaths`

sharmilajesupaul:
**Summary**
This fixes https://github.com/facebook/metro/issues/737 take a look at the issue for more info. I've also added a test case to hopefully catch this in the future. The order of the paths is important in this change.

This change changes the resolution order slightly, we can see this by observing `allDirPaths` https://github.com/facebook/metro/blob/02fade69cad95488e1afba0c3dba8e25d8453bc5/packages/metro-resolver/src/resolve.js#L126-L128

Without this change, the `allDirPaths` looks like this:
```
    [
      '/other-root/node_modules/banana-module',
      '/other-root/node_modules/banana/node_modules/banana-module',
      '/other-root/node_modules/node_modules/banana-module',
      '/other-root/node_modules/banana-module',
      '/node_modules/banana-module'
    ]
```

See how the `banana-module` nested inside `banana` is second & not first.

With this change `allDirPaths` is update to this:
```
    [
      '/other-root/node_modules/banana/node_modules/banana-module',
      '/other-root/node_modules/node_modules/banana-module',
      '/other-root/node_modules/banana-module',
      '/node_modules/banana-module',
      '/other-root/node_modules/banana-module'
    ]
```

We correctly attempt to resolve the transitive `banana-module` dependency first.

**Test plan**
Added a unit test

### DIFF
--- a/packages/metro-resolver/src/__tests__/index-test.js
+++ b/packages/metro-resolver/src/__tests__/index-test.js
@@ -56,9 +56,19 @@ const CONTEXT: ResolutionContext = (() => {
       },
       'other-root': {
         node_modules: {
+          'banana-module': {
+            'package.json': true,
+            'main.js': true,
+          },
           banana: {
             'package.json': true,
             'main.js': true,
+            node_modules: {
+              'banana-module': {
+                'package.json': true,
+                'main.js': true,
+              },
+            },
           },
         },
       },
@@ -216,12 +226,33 @@ it('uses `nodeModulesPaths` to find additional node_modules not in the direct pa
   expect(() => Resolver.resolve(context, 'kiwi', null))
     .toThrowErrorMatchingInlineSnapshot(`
     "Module does not exist in the Haste module map or in these directories:
-      /other-root/node_modules
       /root/project/node_modules
       /root/node_modules
       /node_modules
+      /other-root/node_modules
     "
   `);
+});
+
+it('resolves transitive dependencies when using `nodeModulesPaths`', () => {
+  const context = Object.assign(
+    {},
+    {...CONTEXT, originModulePath: '/other-root/node_modules/banana/main.js'},
+    {
+      nodeModulesPaths: ['/other-root/node_modules'],
+    },
+  );
+
+  expect(Resolver.resolve(context, 'banana-module', null)).toEqual({
+    type: 'sourceFile',
+    filePath:
+      '/other-root/node_modules/banana/node_modules/banana-module/main.js',
+  });
+
+  expect(Resolver.resolve(context, 'banana-module', null)).not.toEqual({
+    type: 'sourceFile',
+    filePath: '/other-root/node_modules/banana-module/main.js',
+  });
 });
 
 describe('disableHierarchicalLookup', () => {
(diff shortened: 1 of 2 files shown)


packages/metro-resolver/src/__tests__/index-test.js
```diff
@@ -56,9 +56,19 @@ const CONTEXT: ResolutionContext = (() => {
...
+            'main.js': true,
+          },
           banana: {
             'package.json': true,
             'main.js': true,
+            node_modules: {
+              'banana-module': {
+                'package.json': true,
+                'main.js': true,
+              },
```
raejin, line 70: I like the example when you suggested a regular module under `banana/node_modules/other-module/main.js` that can also be resolved from `banana/main.js`.

raejin (commented):
Looks great! Thanks for fixing this!

raejin:
cc @motiz88

motiz88:
I'm working on moving an internal call site away from `nodeModulesPaths` so we can land this. This has been on my radar in recent months - thanks so much for the fix, clear summary and updated tests.

I'll cut a new major release of Metro afterwards, as this is strictly speaking a breaking change.

sharmilajesupaul:
Thanks @motiz88 & @raejin! Let me know if you'd liked to see more changes made here as well.

packages/metro-resolver/src/resolve.js
```diff
@@ -102,6 +103,8 @@ function resolve(
     } while (candidate !== next);
   }
 
+  nodeModulesPaths.push(...context.nodeModulesPaths);
```
raejin, line 106: Let's add helpful comments here explaining why we only push the custom node paths until now

sharmilajesupaul: good suggestion! what do you think about something like this?
```suggestion
  /**
   * It's important that we push nodeModulesPaths to the list of resolution
   * candidates after the do while loop above.
   *
   * This ensures that, when we resolve a transitive dependencies inside
   * nodeModulesPaths, the hierarchical lookup is still respected and takes
   * precedence over the root of the nodeModulesPaths locations.
   *
   * E.g.
   * When resolving an import in /extra-node-modules/foo/main.js with
   * nodeModulesPaths=["/extra-node-modules"].
   *
   * We want to resolve to modules in /extra-node-modules/foo/node_modules
   * first, before /extra-node-modules.
   */
  nodeModulesPaths.push(...context.nodeModulesPaths);
```

motiz88: FWIW I'd be OK without a super verbose comment here. Something like `// Fall back to `nodeModulesPaths` after hierarchical lookup, similar to $NODE_PATH` would be fine. The commit history will have the rest of the context.

Not totally opposed to the longer comment, but it's the kind of thing that can be hard to keep up to date and correctly anchored as the code evolves.

(OTOH we should absolutely have a page in the docs that goes into detail about the resolution algorithm - I intend to write one soon - and I can see why in the absence of one we'd want to annotate the code more.)

sharmilajesupaul: Sounds good! I'm always trying to weigh what might confusing to people, but your suggestion works! & the issue commit should help explain more in lieu of that docs page 😅 
```suggestion
  // Fall back to `nodeModulesPaths` after hierarchical lookup, similar to $NODE_PATH
  nodeModulesPaths.push(...context.nodeModulesPaths);
```

PR #745, opened 2021-11-29
Use `Object.assign` instead of Babel `_extends`

newyankeecodeshop:
**Summary**

All React Native JavaScript runtimes, as well as Node, implement `Object.assign()`. This PR removes the transform and configures plugins to use `Object.assign` instead of injecting the Babel helper `_extends()`. This simplifies the generated code, and it aligns JavaScriptCore users more closely with Hermes, which doesn't use several of the plugins.

The following plugins generate code using `Object.assign` directly:

- `transform-destructuring`
- `proposal-object-rest-spread`
- `transform-react-jsx`

**Test plan**

I added a unit test to verify the new behavior.
The Babel `_extends` helper already uses Object.assign if it exists, so the runtime behavior will be identical.
```
_extends = Object.assign || function (target) {
  ...
}
```


ericanderson (commented):
thanks!

rh389:
Thanks again @newyankeecodeshop 👍

PR #750, opened 2021-12-21
Fix edit url

softmarshmallow:
**Summary**

This PR is a very simple update that alters metro's docs page edit url by changing docusurus.config.js' edit url property
The previous one was outdated.


charlesbdudley:
Hi @softmarshmallow, looks like this url is still using the old `master` branch. Can we update this to `main` and write a Summary in the PR description?

softmarshmallow:
@charlesbdudley Done!

not much to describe anyway :)

PR #754, opened 2022-01-07
Allow for async minifiers

rockwotj:
**Summary**

Terser v5 is async, so if we want to upgrade to support the latest syntax we need this.

**Test plan**

Existing unit tests - this only adds an await clause, which is safe to do to non-promises, so it shouldn't break anything.

### DIFF
--- a/packages/metro-transform-worker/src/index.js
+++ b/packages/metro-transform-worker/src/index.js
@@ -72,7 +72,9 @@ export type MinifierResult = {
   ...
 };
 
-export type Minifier = MinifierOptions => MinifierResult;
+export type Minifier = MinifierOptions =>
+  | MinifierResult
+  | Promise<MinifierResult>;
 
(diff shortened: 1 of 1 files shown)


rockwotj:
https://github.com/facebook/metro/pull/606 is related - but this doesn't change the terser package - I'll send another PR for that. That PR was never merged because of a need for using the transform worker sync, but I believe that is obsolete as there is already a bunch of await clauses in that file

rockwotj:
@sota000 is there any hope of merging this?

rh389:
@rockwotj this requires a few modifications to the way we use metro internally, but it’s in the pipeline

rockwotj:
Thanks so much!

rh389:
FYI @rockwotj - no need to send another PR for updating to terser v5 - I’ve picked up #606, updated internally and it’s in review now.

PR #761, opened 2022-01-24
Use node devtools url

janicduplessis:
**Summary**

This uses the new chrome devtools ui, this is the one used by the nodejs inspector. It is more focused on js debugging / profiling instead of general browser devtools.

**Test plan**

Tested that it works in Flipper and also when opened directly in chrome.

### Before

image

### After

image


janicduplessis:
Note that the Profiler tab is currently broken in flipper, see https://github.com/facebook/flipper/pull/3355

However this functionality is not working currently anyway since it's not integrated with hermes so it should not block this.

rh389:
Thanks @janicduplessis !

PR #766, opened 2022-01-30
Bump react-refresh to 0.11.x

stianjensen:
It has received a lot of small fixes since 0.4.x that would be nice to
use in react-native.


robhogan:
Closing this as Metro no longer has a dependency on `react-refresh`, and RN's has been updated.

PR #768, opened 2022-02-01
Remove `mkdirp@^0.5.1` due to CVE warnings, use native recursive `fs.mkdir`

ericanderson:
**Summary**

mkdirp was triggering github security warnings. Upgrading version will resolve that.

**Details**

Github security warnings. The ones where you are the owner of a project and you visit the project's home page and get a giant warning from GitHub telling you there are CVE's in libraries within your repo.

I don't know for sure if metro was or was not vulnerable. I for sure know this wouldn't affect someone's production code. What I do know is that humans build tolerance and blindness to things over time. Having a security warning linger on a repo/project is a long term risk to other, real security vulnerabilities being seen and fixed.

This version bump fixes the issue because it is the recommended version to resolve the CVE.

The code delta is to deal with the vulnerable library having a major semver change and thus the code needed to be adapted.

**Test plan**

`yarn test`


packages/metro-memory-fs/src/index.js
```diff
@@ -675,19 +675,22 @@ class MemoryFs {
...
+    const realMode = (typeof mode == 'number' ? mode : mode?.mode) ?? 0o777;
+
     dirPath = pathStr(dirPath);
     const {dirNode, node, basename} = this._resolve(dirPath);
-    if (node != null) {
+    if (node != null && !recursive) {
       throw makeError('EEXIST', dirPath, 'directory or file already exists');
     }
-    dirNode.entries.set(basename, this._makeDir(mode));
+    dirNode.entries.set(basename, this._makeDir(realMode));
```
philIip, line 693: can we move the initialization of `realMode` here? we potentially may not even use the realMode we calculated if the condition in L690 passes.

philIip (changes requested):
thanks @ericanderson for working on this! in the diff summary, can you give some examples of the security warnings and why this version bump and implementation change fixes them? thanks!

ericanderson:
@philIip Github security warnings. The ones where you are the owner of a project and you visit the project's home page and get a giant warning from GitHub telling you there are CVE's in libraries within your repo. 

I don't know for sure if metro was or was not vulnerable. I for sure know this wouldn't affect someone's production code. What I do know is that humans build tolerance and blindness to things over time. Having a security warning linger on a repo/project is a long term risk to other, real security vulnerabilities being seen and fixed.

This version bump fixes the issue because it is the recommended version to resolve the CVE.

The code delta is to deal with the vulnerable library having a major semver change and thus the code needed to be adapted.

packages/metro-memory-fs/src/index.js
ericanderson: I could. I think keeping logic like this based on the input of a function together at the start helps the readability of the code. Just as the original code had the default value of `mode` provided at the top of the file and not down near `this._makeDir(...)` and just as the original code did not make use of the updated `mode` value unless the if check were violated.

Given that this would optimize at best a degenerate case of the node already existing and at worst 2 more comparisons compared to the prior version, I don't know that its worth the legibility change. But if you insist I will make that change.

(Side note, I don't know flow. In TS, I could have reused the `mode` variable in my default processing assignment instead of introducing `realMode` and the compiler would infer that from that point on the code is always a number and not a number or undefined or object. Is this possible with flow? See here for TS example: https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAWwNYBMYCcDKBPSACgChEzFMsAFAQygAsAuRAZyixjAHMAaU85HHQBTAPzMwIZACNhWRAB9EAbyzCIILCxgA3Mc2lw4AG2E0wPFEP2JJMuQF8+ZAJQr+ZCAjaI1NYwCy1ogAvIiEUHgADsJwwFYioWEA5HayWMmIognCiMyCIqIAdAXCbqLZAAxwAOx1ANweOaHhkTFxzSEpaXKZ2aV5OcWl5VW1DcRNXmAsJsJFxnBchCON5IgA9BvrO4gAeqL8DkA )

packages/metro-memory-fs/src/index.js
```diff
@@ -675,19 +675,22 @@ class MemoryFs {
     }
   };
 
-  mkdirSync: (dirPath: string | Buffer, mode?: number) => void = (
+  mkdirSync: (
     dirPath: string | Buffer,
-    mode?: number,
+    mode?: number | {recursive?: boolean, mode?: number},
```
rh389, line 680: This API change seems unused and unrelated to the dependency bump. What's the thinking here?

ericanderson: Great question. This one caught me by surprise when I made the change too.

metro-memory-fs is a shim for the nodejs `fs` package. It is used a lot in the tests at the very least. 

mpdirp used to do the recursive directory creation manually but in node 10, the `require('fs').mkdir` api changed to add the recursive flag. So the mkdirp package uses that now instead. The (very good and thorough) tests were then failing because this library is shim'ing nodejs and thus the code is no longer working.

This change to mkdirp mirrors the API from node 10's mkdir, as far as is needed for this repo's use of mkdirp.

ericanderson:
As an aside which shouldn't block this change, it seems like maintaining `metro-memory-fs` is an unnecessary use of resources. This package is only used for tests (at least within this repo). 

The [memfs](https://github.com/streamich/memfs) library seems complete and well maintained. It is what webpack switched to after abandoning its own [memory-fs](https://github.com/webpack/memory-fs) (top right corner see deprecation notice).

packages/metro-memory-fs/src/index.js
rh389: Ah, got it, thanks - I didn't realise this was a shim for `fs`. In that case I'd just rename `mode` -> `options` and `realMode` -> `mode`.

rh389:
Hi @ericanderson - thanks for working on this. Unfortunately, it's proving problematic to merge internally due to the way we manage direct dependencies across the containing workspace (upshot being we'd have to update `mkdirp` in *a lot* of other places in order to merge this). Sorry that has an affect on a perfectly good PR.

To be honest, I'm thinking it'd be simpler to eliminate `mkdirp` from metro entirely - we support node >= 12 so the native `recursive` option is available to us and should be all we need. If you're interested in pivoting this PR in that direction it'd be very welcome, completely understand if not - apologies again for the wasted time.

ericanderson:
@rh389 fair enough, will pivot as I'd prefer to get these warnings out of my repos.

packages/metro/src/shared/output/RamBundle/as-assets.js
```diff
@@ -86,9 +87,8 @@ function saveAsAssets(
 }
 
 function createDir(dirName: string): Promise<empty> {
-  return new Promise((resolve: void => void, reject: mixed => mixed) =>
-    mkdirp(dirName, error => (error ? reject(error) : resolve())),
-  );
+  const mkdirPromised = util.promisify(fs.mkdir);
+  return mkdirPromised(dirName, {recursive: true}).then(() => undefined);
```
rh389, line 91:
```suggestion
  return fsPromises.mkdir(dirName, {recursive: true});
```

packages/metro/src/shared/output/RamBundle/as-assets.js
```diff
@@ -86,9 +87,8 @@ function saveAsAssets(
 }
 
 function createDir(dirName: string): Promise<empty> {
```
rh389, line 89:
```suggestion
function createDir(dirName: string): Promise<void> {
```

packages/metro/src/shared/output/RamBundle/as-assets.js
```diff
@@ -19,8 +19,9 @@ const writeFile = require('../writeFile');
 const buildSourcemapWithMetadata = require('./buildSourcemapWithMetadata');
 const MAGIC_RAM_BUNDLE_NUMBER = require('./magic-number');
 const {joinModules} = require('./util');
+const fs = require('fs');
+const util = require('util');
```
rh389, line 23:
```suggestion
const fsPromises = require('fs').promises;
```

rh389 (commented):
I think we could just tidy up the async call with the fs promises API, otherwise perfect (and thanks again)

ericanderson:
Thanks for the help getting it through @rh389. Sorry about the false start earlier. I was in and out of meetings and pushed before running tests.

rh389:
Not at all - I was just keen to get it pulled in before the end of my day yesterday. It's waiting for internal review now but everything's green so I'd expect it to be in the next fortnightly release, this PR will auto-close when it lands. Thanks again for the contribution (and patience!)

PR #771, opened 2022-02-10
Allow customizing the babel transform runtime version

newyankeecodeshop:
**Summary**

As described in PR #714 , when the bundle is built using a recent version of babel, some helpers are defined repeatedly instead of being referenced via the runtime. In our case, `_getRequireWildcardCache()` and `_interopRequireWildcard()`.

Since the react native preset configures `@babel/plugin-transform-runtime` when `enableBabelRuntime` is enabled, I thought it would be straightforward to overload this metro configuration option to allow the runtime version to be specified. Therefore, changing `babel.config.js` to be:
```javascript
module.exports = {
  presets: [
    [
      "module:metro-react-native-babel-preset",
      { enableBabelRuntime: "7.14.5" }
    ]
  ]
};
```
The above would both enable the runtime and configure it to use version `7.14.5` behaviors. An app developer can do `yarn why @babel/runtime` to determine the runtime version and specify it in this configuration.

Adding this option doesn't change the behavior for existing users who are using `true | false | undefined` for the babel runtime configuration.

**Test plan**

All unit tests pass.

I modified our application as specified above while referencing my modified build of metro. The bundle showed a 6% size reduction and `_interopRequireWildcard` was required via the runtime.

### DIFF
--- a/packages/metro-react-native-babel-preset/src/configs/main.js
+++ b/packages/metro-react-native-babel-preset/src/configs/main.js
@@ -165,7 +165,7 @@ const getPreset = (src, options) => {
 
   if (!options || options.enableBabelRuntime !== false) {
     // Allows configuring a specific runtime version to optimize output
-    const isVersion = typeof options.enableBabelRuntime === 'string';
+    const isVersion = typeof options?.enableBabelRuntime === 'string';
 
     extraPlugins.push([
       require('@babel/plugin-transform-runtime'),
--- a/packages/metro-react-native-babel-preset/src/configs/main.js
+++ b/packages/metro-react-native-babel-preset/src/configs/main.js
@@ -172,7 +172,7 @@ const getPreset = (src, options) => {
       {
         helpers: true,
         regenerator: !isHermes,
-        version: isVersion ? options.enableBabelRuntime : '7.0.0-beta.0',
+        ...(isVersion && {version: options.enableBabelRuntime}),
       },
     ]);
   }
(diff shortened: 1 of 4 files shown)


packages/metro-react-native-babel-preset/src/configs/main.js
```diff
@@ -164,11 +164,15 @@ const getPreset = (src, options) => {
...
   if (!options || options.enableBabelRuntime !== false) {
+    // Allows configuring a specific runtime version to optimize output
+    const isVersion = typeof options?.enableBabelRuntime === 'string';
+
     extraPlugins.push([
       require('@babel/plugin-transform-runtime'),
       {
         helpers: true,
         regenerator: !isHermes,
+        version: isVersion ? options.enableBabelRuntime : '7.0.0-beta.0',
```
rh389, line 175: Can we avoid baking in the plugin's default fallback version?
```suggestion
        ...(isVersion && {version: options.enableBabelRuntime}),
```

rh389:
Thanks again @newyankeecodeshop - this looks like a useful optimisation

packages/metro-react-native-babel-preset/src/configs/main.js
newyankeecodeshop: Yes definitely - I like this mechanism for initializing a specific property only under certain circumstances.

leotm:
Just confirming, does this mean we need to update the overload each time to benefit from latest `@babel/runtime`
- https://github.com/leotm/react-native-template-typescript/pull/452

With the -6% benefit? Just thinking how to automate this easily, might look into a `+` option

newyankeecodeshop:
@leotm You don't have to update the version each time you upgrade your babel necessarily. Setting the version to a reasonably newer version, such as `7.14.5`, will provide that bundle size reduction (via optimizing `interopRequireWildcard`).
You can see what helpers the version affects in this file:
https://github.com/babel/babel/blob/main/packages/babel-helpers/src/helpers.ts

PR #787, opened 2022-03-10
Enable `loose` on `plugin-transform-classes` for smaller bundles

tido64:
## Summary

`plugin-transform-classes` currently injects lots of helper functions for every class it encounters. If a codebase makes heavy use of classes, there is a lot to be saved by enabling loose mode.

## Test plan

The output should be smaller, but it's functionally the same.

### DIFF
--- a/packages/metro-react-native-babel-preset/src/configs/main.js
+++ b/packages/metro-react-native-babel-preset/src/configs/main.js
@@ -77,7 +77,12 @@ const getPreset = (src, options) => {
   }
 
   if (hasClass) {
-    extraPlugins.push([require('@babel/plugin-transform-classes')]);
+    // If the code makes heavy use of classes, we can save a lot of bytes by
+    // avoiding injecting `@babel/runtime` helpers.
+    extraPlugins.push([
+      require('@babel/plugin-transform-classes'),
+      {loose: !options.dev},
+    ]);
   }
 
   // TODO(gaearon): put this back into '=>' indexOf bailout


tido64:
cc @motiz88

motiz88:
I haven't looked into the class transform in particular yet, but in general we shouldn't turn on any new loose semantics by default - the risk of breaking some correct code or silently allowing incorrect code is too large (when multiplied by all React Native users). I'd much prefer alternatives that focus on lowering the cost of using Babel helpers - I'll link some relevant PRs.

motiz88:
See #714, #771.

packages/metro-react-native-babel-preset/src/configs/main.js
```diff
@@ -77,7 +77,12 @@ const getPreset = (src, options) => {
   }
 
   if (hasClass) {
-    extraPlugins.push([require('@babel/plugin-transform-classes')]);
+    // If the code makes heavy use of classes, we can save a lot of bytes by
+    // avoiding injecting `@babel/runtime` helpers.
+    extraPlugins.push([
+      require('@babel/plugin-transform-classes'),
+      {loose: !options.dev},
```
motiz88, line 84: Further to my other comments: we should also strongly avoid compiling with different semantics in dev vs prod unless absolutely necessary. Diverging creates the risk of prod-only bugs and a JS environment that is harder to reason about.

motiz88:
This is pretty much a wontfix - React Native users wishing to deviate from the JS spec can always use a custom Babel preset of their choosing. Happy to discuss other approaches to reducing the footprint of Babel helpers.

tido64:
> I haven't looked into the class transform in particular yet, but in general we shouldn't turn on any new loose semantics by default - the risk of breaking some correct code or silently allowing incorrect code is too large (when multiplied by all React Native users). I'd much prefer alternatives that focus on lowering the cost of using Babel helpers - I'll link some relevant PRs.
> 
> tl;dr the best practice is to ensure helpers aren't duplicated in a bundle - injecting them in each module is a bad default that we should fix, and is the root cause of the bloat you're trying to fix here AFAICT.

Thanks for the explanation. I wasn't sure myself whether this makes sense, but I did see significant savings turning this on. The reason for the bloat is due to all the extra checks that Babel injects, e.g.:

```js
// 12 bytes
class Foo {}
```

Gets transformed into:

```js
// 406 bytes
"use strict";

var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefault");

var _createClass2 = _interopRequireDefault(require("@babel/runtime/helpers/createClass"));

var _classCallCheck2 = _interopRequireDefault(require("@babel/runtime/helpers/classCallCheck"));

var Foo = /*#__PURE__*/(0, _createClass2.default)(function Foo() {
  (0, _classCallCheck2.default)(this, Foo);
});
```

Compared to loose mode (and what TypeScript outputs):

```js
// 43 bytes
"use strict";

var Foo = function Foo() {};
```

That's 12 -> 406 bytes for an empty class. Without the runtime helper, it's 806 bytes. Loose mode is closer to what TypeScript outputs. Since we use TypeScript everywhere, the checks have already been done so it's wasteful to have all these helpers in the bundle.

motiz88:
I think the `require`s should minify well, and we can try to squeeze out the `interopRequireDefault`s and `.default`s too with some additional effort. How much larger is the minified version of Babel's output? (And the Hermes bytecode version?)

ls-andrew-goodale:
@tido64 It looks like Babel has a new feature, "assumptions", that is designed to address what you're talking about.
https://babeljs.io/docs/en/assumptions
I haven't tried it but some options like `noClassCalls` that looks like will remove that extra logic.

PR #789, opened 2022-03-11
Remove `@babel/plugin-transform-regenerator`

tido64:
## Summary

Hermes added support for generators in 0.2.0, and JSC is ES6 feature complete since [r202125](https://webkit.org/blog/6756/es6-feature-complete/).

## Test plan

n/a

### DIFF
--- a/packages/metro-react-native-babel-preset/package.json
+++ b/packages/metro-react-native-babel-preset/package.json
@@ -47,7 +47,6 @@
     "@babel/plugin-transform-react-jsx": "^7.0.0",
     "@babel/plugin-transform-react-jsx-self": "^7.0.0",
     "@babel/plugin-transform-react-jsx-source": "^7.0.0",
-    "@babel/plugin-transform-regenerator": "^7.0.0",
     "@babel/plugin-transform-async-to-generator": "^7.0.0",
     "@babel/plugin-transform-runtime": "^7.0.0",
     "@babel/plugin-transform-shorthand-properties": "^7.0.0",
--- a/packages/metro-react-native-babel-preset/src/configs/main.js
+++ b/packages/metro-react-native-babel-preset/src/configs/main.js
@@ -120,9 +120,7 @@ const getPreset = (src, options) => {
     extraPlugins.push([
       require('@babel/plugin-proposal-async-generator-functions'),
     ]);
-    extraPlugins.push([
-      require('@babel/plugin-transform-async-to-generator'),
-    ]);
+    extraPlugins.push([require('@babel/plugin-transform-async-to-generator')]);
   }
   if (!isHermes && (isNull || src.indexOf('**') !== -1)) {
     extraPlugins.push([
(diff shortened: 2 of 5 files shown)


tido64:
@motiz88: It'd be great if you could make sure this doesn't break Meta internally.

motiz88:
Awesome - can you please run Prettier, remove `@babel/plugin-transform-regenerator` from `package.json` and update `yarn.lock` accordingly?

motiz88:
Looking good so far - I'm fixing up some minor Meta-internal things and planning to land this on Monday. Thanks for this, @tido64!

newyankeecodeshop:
This is great. I believe you can also remove `babel-plugin-transform-for-of` since iteration is also supported in both runtimes. Hopefully the minifier supports the syntax too.

rh389:
`uglify-es` and `terser` both fully support ES6 so I wouldn’t expect any issue for generators or for/of. `uglify-es` starts to get ugly for ES7+, which is part of the reason I’m looking to shift the default to `terser`.

tido64:
image

@motiz88: Something's failing internally. Is it because of this change?

> This is great. I believe you can also remove `babel-plugin-transform-for-of` since iteration is also supported in both runtimes.

That's a great idea, but I think it should be a separate PR.

> > Hopefully the minifier supports the syntax too.
> 
> That is a good point - I'm not actually sure we have a test that would fail if it didn't. @tido64 would you mind checking to see what happens if you run this PR with `minify=true`? (cc @rh389 re: minifier stuff)

How do I turn this on in tests? I tried pasting some of the snapshots in [Terser's REPL](https://try.terser.org/) and [Uglify's demo](https://lisperator.net/uglifyjs/). Terser is fine with the generator function, but UglifyJS doesn't support it. I know uglify-es is supposed to support ES6, but it's deprecated and I see that people are trying to migrate to the former: https://github.com/facebook/metro/pull/661

Is this going to be a problem?

rh389:
> I know uglify-es is supposed to support ES6, but it's deprecated and I see that people are trying to migrate to the former: https://github.com/facebook/metro/pull/661

There's a bit of history here - `uglify-js` originally didn't support ES6. The author created `uglify-es` (originally a `harmony` branch of `mishoo/UglifyJS2`) to address that, but that was itself [deprecated](https://www.npmjs.com/package/uglify-es) some time ago in favour of `uglify-js`, as `uglify-js` v3 *does* now support ES6 out of the box. I suspect that demo is just out of date. They all have CLIs:

```sh
➜ echo "function* myGenerator(myVar) { yield myVar + 1; }" | npx uglify-js -m
function*myGenerator(e){yield e+1}

➜  echo "function* myGenerator(myVar) { yield myVar + 1; }" | npx terser -m
function*myGenerator(e){yield e+1}

➜  echo "function* myGenerator(myVar) { yield myVar + 1; }" | npx uglify-es -m
function*myGenerator(e){yield e+1}
```

Terser was originally a fork of `mishoo/UglifyJS2` so it has a shared ancestry with `uglify-es`, and internally we've found it to be closer to `uglify-es` in output than `uglify-js@3` is (#661 turned out to be problematic internally), so we're working to make `terser` the metro default and we'll probably deprecate `metro-minifier-uglify`.

**TL;DR**: Nothing minifier-related to worry about for this PR.

rh389:
> Something's failing internally. Is it because of this change?

Unrelated - React Native iOS tests are broken at the moment.

PR #791, opened 2022-03-14
Added named groups babel transform to rn babel preset

danilobuerger:
## Summary

Hermes currently does not support regex named groups. Whenever a library is included that uses named groups the rn build with hermes will fail (as it doesn't compile). This is already the case when using jest 27 as this includes a dependency (ansi-styles) that uses named-groups which is transitively included via pretty-format in HMRClient.

## Test plan

no test plan

### DIFF
--- a/packages/metro-react-native-babel-preset/src/configs/main.js
+++ b/packages/metro-react-native-babel-preset/src/configs/main.js
@@ -34,6 +34,7 @@ const defaultPluginsBeforeRegenerator = [
 ];
 
 const defaultPluginsAfterRegenerator = [
+  [require('@babel/plugin-transform-named-capturing-groups-regex')],
   [require('@babel/plugin-transform-unicode-regex')],
 ];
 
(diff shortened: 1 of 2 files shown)


packages/metro-react-native-babel-preset/src/configs/main.js
```diff
@@ -31,6 +31,7 @@ const defaultPlugins = [
   [require('@babel/plugin-syntax-dynamic-import')],
   [require('@babel/plugin-syntax-export-default-from')],
   ...passthroughSyntaxPlugins,
+  [require('@babel/plugin-transform-named-capturing-groups-regex')],
```
newyankeecodeshop, line 34: I think JavaScriptCore for iOS 11+ has this feature - maybe this plugin should be added in the "extraPlugins" section only `if (isHermes)`?
The Android JavaScriptCore has functionality of about the iOS 12 release, so I expect it also has this feature.

danilobuerger: Thanks for the review @newyankeecodeshop . I updated the PR with your recommendation.

danilobuerger:
@robhogan anything I can do about the internal test fails?

robhogan:
Thanks but I've taken care of it - just an internal lockfile needed syncing.

PR #793, opened 2022-03-19
Replace deprecated `uglify-es` with `uglify-js`

Simek:
## Summary

This PR replaces the deprecated `uglify-es` package (https://www.npmjs.com/package/uglify-es) with the `uglify-js` as suggested in the deprecation notice. 

Screenshot 2022-03-19 at 18 18 49

Affected packages are `metro-minify-uglify` and `metro-source-map`.

It looks like the switch should be quite trouble free, since there are almost no difference between packages APIs.

## Test plan

I have updated the dependencies and code, then I run the `test` script from the workspace root.

### DIFF
--- a/packages/metro-minify-uglify/package.json
+++ b/packages/metro-minify-uglify/package.json
@@ -13,6 +13,6 @@
   },
   "license": "MIT",
   "dependencies": {
-    "uglify-es": "^3.1.9"
+    "uglify-js": "^3.15.3"
   }
 }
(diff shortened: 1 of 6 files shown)


rh389:
Hi @Simek - thanks for the PR. It's a duplicate of #661 though, and the reason that hasn't been merged is that the minified output of `uglify-js` differs from `uglify-es` in a couple of ways that turn out to be problematic internally.

We're planning to switch Metro's default to `terser` instead (originally a fork of `uglify-es`), as soon as we can overcome the issue in https://github.com/terser/terser/pull/1164.

PR #796, opened 2022-03-23
Auto-build `emhermesc.js` with GitHub Action

robhogan:
## Motivation

 - The [current build instructions](https://github.com/facebook/metro/blob/bdcf52de737f20088e18a926005f822e82017cbf/packages/metro-hermes-compiler/README.md) are out of date since Hermes required a two-stage build, and the suggested Docker base is deprecated.
 - It's currently very difficult to replicate/verify this build without Buck
 - It's a laborious process to update or review any updates to this file
 - There's no good way to know what version of Hermes the current `emhermesc.js` was built from.
 - We're overdue an update now, not least to fix the temporary patch in https://github.com/facebook/metro/commit/4be4099184c98ffc0b7687c2322f7dd863d3fbb6
 
## Change
See the diff on the Readme (TL;DR - this uses a GitHub action to perform the build based on `hermes-github-ref`, and auto-pushes to a PR / uploads the result / fails on a mismatch).

## Test plan

I iterated this on my fork: https://github.com/rh389/metro 

### Job fails because the file is out of sync on main
https://github.com/rh389/metro/runs/5662424803?check_suite_focus=true
image
 
### Job succeeds by pushing to a PR
https://github.com/rh389/metro/pull/3
image

### Artifacts are uploaded
image


robhogan:
Closing this for now as there's some more work to do to find a nice workflow - it seems that we can't commit to forks from an action even when the PR author allows commits from maintainers.

PR #800, opened 2022-04-08
Fix platform inline transform for babel esmodules commonjs transform output

janicduplessis:
## Summary

I noticed that when using esmodule named import of platform dead code elimination of Platform.OS did not work.

#### Does not work

Source:
```js
import { Platform } from 'react-native';

if (Platform.OS === 'android') {
  // code
}
```
Transformed:
```js
var _reactNative = _$$_REQUIRE(_dependencyMap[4]);

if (_reactNative.Platform.OS !== 'android') {
  // code
}

```

#### Works

Source:
```js
const { Platform } = require('react-native');

if (Platform.OS === 'android') {
  // code
}
```
Transformed:
```js
var _require = _$$_REQUIRE(_dependencyMap[4]),
Platform = _require.Platform;

if ('ios' !== 'android') {
  // code
}
```

The inline plugin works on transformed and bundled code. For the working `require` case it works since `Platform` is assigned to a top level variable, which the plugin considers as valid for inlining. In the case of `import` babel transforms it to a single top level require call (`_reactNative`), but the name doesn't match any of the allowed ones (`ReactNative`, `React`). This means it is not considered valid for inlining.

The name of the variable seems based on the name of the module imported, so `react-native` becomes `_reactNative`. I also noticed some cases in an app bundle where babel will add a number suffix to the variable like `_reactNative2`. This seems to be cause by other imports containing the name `react-native` like `@sentry/react-native`.

To support both cases I added support for passing a regex to the valid top level binding names. We can then use this to match `_reactNative` + optional number.

## Test plan

Added a test for the case similar to how code looks in the transformed bundle before being passed to the inline plugin. Also tested this in an app bundle and looked for instances of `Platform.OS` than should be inlined and made sure there was none.

### DIFF
--- a/packages/metro-transform-plugins/src/__tests__/inline-plugin-test.js
+++ b/packages/metro-transform-plugins/src/__tests__/inline-plugin-test.js
@@ -795,6 +795,60 @@ describe('inline constants', () => {
     );
   });
 
+  it('can work with transformed import calls', () => {
+    const code = `
+      __arbitrary(require, function(arbitraryMapName) {
+        var _reactNative = require(arbitraryMapName[123]);
+
+        function a() {
+          if (_reactNative.Platform.OS === 'android') {
+            a = function() {};
+          }
+
+          var b = a.Platform.OS;
+        }
+      });
+    `;
+
+    compare(
+      [inlinePlugin],
+      code,
+      code.replace(/_reactNative\.Platform\.OS/, '"ios"'),
+      {
+        inlinePlatform: true,
+        platform: 'ios',
+        isWrapped: true,
+      },
+    );
+  });
+
+  it('can work with transformed import calls numbered', () => {
+    const code = `
+      __arbitrary($$require, function(arbitraryMapName) {
+        var _reactNative2 = $$require(arbitraryMapName[123]);
+
+        function a() {
+          if (_reactNative2.Platform.OS === 'android') {
+            a = function() {};
+          }
+
+          var b = a.Platform.OS;
+        }
+      });
+    `;
+
+    compare(
+      [inlinePlugin],
+      code,
+      code.replace(/_reactNative2\.Platform\.OS/, '"ios"'),
+      {
+        inlinePlatform: true,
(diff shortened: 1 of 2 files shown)


motiz88:
I agree that this needs to be fixed, but I don't think we should use a regex to do this as it's too brittle and unsound. We need to find a way to properly track/detect imported bindings in Metro's intermediate representation.

Notably I think Platform inlining does work with `experimentalImportExportSupport: true`, which we should probably document and turn on by default.

motiz88 (changes requested):
See above comment.

motiz88:
One practical and safe way we can do this is by passing the dependency map to the plugin and using it to understand that the `_$$_REQUIRE(_dependencyMap[4])` call evaluates to the `react-native` module.

janicduplessis:
Yea, the current implementation is brittle since it doesn't have any info about modules and just relies on variable names. Passing the dependency map seems like the proper solution. I'll see when I can find some time to work on this again.

janicduplessis:
@motiz88 Any change we can merge this as is to fix the issue? Sadly I don't think I will have time to implement a better solution.

If we don't want to merge this then how stable is `experimentalImportExportSupport`? Could it be turned on by default and considered not experimental anymore at the moment?

motiz88:
At Meta, we've been using `experimentalImportSupport` in production for a few years. I think the main thing to consider at this point is how spec-compliant it is, and more to the point, whether it has any observable differences from the stock Babel transform in that regard. If we find that it is a drop-in replacement it would be an easy choice to default it to `true`.

(cc @jacdebug for the spec compliance question, maybe?)

robhogan:
Thanks for your efforts here - this issue has (just recently) been fixed by moving the inline plugin ahead of the CommonJS transform, and into `@react-native/babel-preset`: https://github.com/facebook/react-native/issues/57848. Metro's own plugin will be removed in future.

That will be released in RN 0.88, so closing as superseded.

PR #806, opened 2022-04-19
Add allowDeclareFields to flow babel plugin

janicduplessis:
## Summary

The recommended way to type context in class components when using typescript is to use the declare field syntax (`declare context: React.ContextType<typeof MyContext>`). This syntax triggers an error in `@babel/plugin-transform-flow-strip-types` since it needs the option `allowDeclareFields` (https://babeljs.io/docs/en/babel-plugin-transform-flow-strip-types#allowdeclarefields) to parse it.

## Test plan

Tested in an app using the following code:

```ts
const AppContext = React.createContext<{}>(null);

class App extends React.Component<any> {
  declare context: React.ContextType<typeof AppContext>;
}
```

### DIFF
--- a/packages/metro-react-native-babel-preset/src/configs/main.js
+++ b/packages/metro-react-native-babel-preset/src/configs/main.js
@@ -173,7 +173,13 @@ const getPreset = (src, options) => {
       // the flow strip types plugin must go BEFORE class properties!
       // there'll be a test case that fails if you don't.
       {
-        plugins: [require('@babel/plugin-transform-flow-strip-types')],
+        plugins: [
+          [
+            require('@babel/plugin-transform-flow-strip-types'),
+            // Can be removed when updating to babel 8 since this will be the default.
+            {allowDeclareFields: true},
+          ],
+        ],
       },
       {
         plugins: defaultPlugins,


janicduplessis:
This actually seems to cause some issues since it will now cause property that are defined to be initialized to undefined, which means that if it overrides a property from the base class it will now break.

newyankeecodeshop:
@janicduplessis Yes I tried to fix this issue in #727. It also seemed wrong that the Flow plugin was parsing TypeScript files, so I changed that as well. I had #727 in a good place but perhaps it was still too high an impact PR to import so I closed it a couple weeks ago.

janicduplessis:
This PR seems like a good solution, maybe @makovkastar can have a look.

makovkastar:
cc @motiz88 who has some context on https://github.com/facebook/metro/pull/727.

newyankeecodeshop:
This comment describes why enabling `allowDeclareFields` is a breaking change:
https://github.com/facebook/metro/pull/727#discussion_r748725857
I think the only way this issue can be fixed safely is to provide a new metro config option to enable it?

PR #808, opened 2022-04-22
Add `"react-native"` to `resolverMainFields` defaults

### LINKED ISSUE #807: Default `resolverMainFields` should include "react-native"

It depends. I'd consider this a bug as there is an inconsistency between the Metro defaults and the actual values used when invoked from the React Native CLI.

The current defaults for `resolver.resolverMainFields` doesn't include `"react-native"`:

```js
const {getDefaultConfig} = require('metro-config');
const defaults = getDefaultConfig.getDefaultValues();
console.log(defaults.resolver.resolverMainFields); // Logs [ 'browser', 'main' ]
```

I had a user of our library trying to extend the default metro config ending up in a situation where `"react-native"` was missing from the `resolverMainFields` because of this. See https://github.com/realm/realm-js/issues/4517#issuecomment-1106330586 for context.

See the reproduction above.

I'd expect this default to be the same as when the metro server is started through the React Native CLI: `['react-native', 'browser', 'main']`. See [cli-plugin-metro/src/tools/loadMetroConfig.ts#L85](https://github.com/react-native-community/cli/blob/master/packages/cli-plugin-metro/src/tools/loadMetroConfig.ts#L85) for the implementation.

N/A IMHO - ask and I'll provide this if you feel it's important for the description.

kraenhansen:
## Summary

This fixes #807 by adding `'react-native'` to the list of default values for the `resolver.resolverMainFields`.

## Test plan

This is a very small change. I verified the snapshot tests failed and updated them accordingly.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -117,7 +117,7 @@ An array of source extensions to include in the bundle. For example, if you woul
 
 Type: `Array<string>`
 
-Specify the fields in package.json files that will be used by the module resolver to do redirections when requiring certain packages. The default is `['browser', 'main']`, so the resolver will use the `browser` field if it exists and `main` otherwise.
+Specify the fields in package.json files that will be used by the module resolver to do redirections when requiring certain packages. The default is `['react-native', 'browser', 'main']`, so the resolver will use the `react-native` field if it exists, alternatively the `browser` and `main` otherwise.
 
 #### `disableHierarchicalLookup`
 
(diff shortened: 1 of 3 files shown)


motiz88:
Thanks @kraenhansen - I appreciate the high-quality PR including updates to tests and docs! However, I am closing it at this point, as explained in https://github.com/facebook/metro/issues/807#issuecomment-1106403161.

PR #809, opened 2022-04-22
Added documentation about React Native CLI defaults

kraenhansen:
## Summary

As suggested in https://github.com/facebook/metro/issues/807#issuecomment-1106403161 this adds a general note on the fact that React Native CLI is setting its own defaults. I'm using the same block-quote style note as used in the "Merging Configurations" section of the document.

## Test plan

Only documentation is changed. No need for testing.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -11,6 +11,8 @@ A Metro config can be created in these three ways (ordered by priority):
 
 You can also give a custom file to the configuration by specifying `--config <path/to/config>` when calling the CLI.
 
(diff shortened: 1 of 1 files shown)


motiz88:
Awesome, thanks!

Nit: I would use [`:::note` blocks](https://docusaurus.io/docs/markdown-features/admonitions) since those are styled a bit more nicely, but this is great too.

kraenhansen:
@motiz88 thanks 👍  I've updated this to use the `:::note` instead - including the existing note in the "Merging Configurations" section.

PR #816, opened 2022-05-05
Add explicit dependency on `@babel/runtime` to `metro-runtime`

robhogan:
Summary:
Previously, `babel/runtime` was a transitive dependency of RN projects via `metro-react-native-babel-preset` -> `babel/plugin-transform-regenerator`.

`babel/runtime` is a dependency of injected babel helpers, so we need it to be a transitive dependency of an RN project under `react-native`.

Removing the previous implicit dependency broke RN builds, this adds the dependency to `metro-runtime` to make it explicit.

I've created a 0.70.x branch for Metro from 0.70.2 and will cherry pick this up as a hotfix 0.70.3 release.

Reviewed By: motiz88

Differential Revision: D36165366

### DIFF
--- a/packages/metro-runtime/package.json
+++ b/packages/metro-runtime/package.json
@@ -12,6 +12,9 @@
     "cleanup-release": "test ! -e build && mv src build && mv src.real src"
   },
   "license": "MIT",
+  "dependencies": {
+    "@babel/runtime": "^7.0.0"
+  },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


schmod:
Would it make sense to specify a more recent version and include it in the default value of `enableBabelRuntime`? This would allow projects to be opted into some (fairly substantial) bundle-size optimizations by default.

PR #821, opened 2022-05-24
feat: add support for capturing require.context in dependency collector

EvanBacon:
## Summary

> First part of my [`require.context`](https://webpack.js.org/api/module-methods/#requirecontext) feature request. [Learn more](https://github.com/microsoft/rnx-kit/discussions/1515).

The key here is to add support for importing the contents of a folder recursively and providing to the user as some sort of map, the regular expression is an added layer for refining the results. This means in areas where we'd register a filename, we'd use a directory name instead.

This PR implements the ability to capture `require.context` invocations in `collectDependencies`. One could think of this like "language support" for the `require.context` feature, in the sense that Metro can now make sense of the new syntax.

Because multiple dependencies can use the same module identifier (`require.context('[module-id]')`) but also provide extra properties like `recursive`, or a RegExp matcher, we need to change the registry key format to include extra properties. The aforementioned mentioned functionality is primarily implemented in `getKeyForDependency`.

The entire feature is implemented behind a feature flag `unstable_allowRequireContext`, e.g. 

`metro.config.js`
```js
const metroConfig = {
  // ...
  transformer: { unstable_allowRequireContext: true }
};
```

This PR does not implement the entire `require.context` feature, we'll need a follow up PR implementing changes to the `DependencyGraph`, `DeltaBundler`, and `metro-runtime`.

## Test plan

- Added unit tests.


packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -45,6 +45,11 @@ type DependencyData<TSplitCondition> = $ReadOnly<{
   // If left unspecified, then the dependency is unconditionally split.
   splitCondition?: TSplitCondition,
   locs: Array<BabelSourceLocation>,
+
+  /** Should search recursively through folders. */
+  recursive?: boolean,
+  /** A filter to run against files that match the directory */
+  filter?: string,
```
motiz88, line 52: I wonder if this should be a RegExp instance (but we probably `JSON.stringify` this data in a few places, which would be problematic)

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -70,6 +77,8 @@ export type State<TSplitCondition> = {
 
 export type Options<TSplitCondition = void> = $ReadOnly<{
   asyncRequireModulePath: string,
+  /** Path to file for handling `require.context` statements */
+  requireContextModulePath: string,
```
motiz88, line 81: I don't think we're going to need a separate helper module a la `asyncRequire` - rather, any shared runtime logic for this would live in `require.js` directly.

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -199,6 +209,16 @@ function collectDependencies<TSplitCondition = void>(
         return;
       }
 
+      if (
+        callee.type === 'MemberExpression' &&
+        (callee.object || {}).name === 'require' &&
+        (callee.property || {}).name === 'context'
```
motiz88, line 215: Please also check:
1. That `object` and `property` are identifiers
2. That `computed` is false
3. That `require` refers to the global and not something else. (`!path.scope.getBinding('require')`)

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -199,6 +209,16 @@ function collectDependencies<TSplitCondition = void>(
         return;
       }
 
+      if (
```
motiz88, line 212: Let's gate this on a new `unstable_allowRequireContext` option that defaults to false.

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
```diff
@@ -42,6 +42,22 @@ const opts = {
...
+  const a = require.context('./', true, /foobar/);
+`);
+  const {dependencies, dependencyMapName} = collectDependencies(ast, opts);
+
+  expect(dependencies).toEqual([
+    {name: './', data: objectContaining({asyncType: null})},
+  ]);
+  expect(codeFromAst(ast)).toEqual(
+    comparableCode(`
+      const a = require.context(${dependencyMapName}[0], { recursive: true, filter: /foobar/ }, "./");
```
motiz88, line 56: I think the code basically needs to be

```suggestion
      const a = require(${dependencyMapName}[0], "./");
```

The way I see it, the values of `recursive` and `filter` will have no effect at runtime, so they only need to be recorded in `dependencies` and not in the transformed code. (You'll want to update the assertion above to check that they are in `dependencies` accordingly.)

This assumes that `${dependencyMapName}[0]` will evaluate at runtime to the ID of an module injected by the bundler, which will have the matching files (after all expansion+filtering) as *its* dependencies. `require(${dependencyMapName][0])` will therefore evaluate to the result of evaluating that module factory - i.e. some object that gives access to those dependencies, as dictated by the `require.context` API.

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -45,6 +45,11 @@ type DependencyData<TSplitCondition> = $ReadOnly<{
   // If left unspecified, then the dependency is unconditionally split.
   splitCondition?: TSplitCondition,
   locs: Array<BabelSourceLocation>,
+
```
motiz88, line 48: We need a way to distinguish `require.context` from all other dependency types, so we know not to try to resolve its name as a file. The problem is this data structure is a bit of a mess - it should probably be a disjoint union with a `type` discriminant.

To avoid a painful refactor (this early in the process) we can add something like `contextParams: $ReadOnly<{recursive: boolean, filter: string}> | null`, along the same lines as `asyncType` and `splitCondition`, such that the dependency is `require.context` iff `contextParams` is nonnull.

motiz88: By the way, looks like Webpack's `require.context` supports a `mode` parameter - let's add at least basic support for that too. (`sync`, `lazy`, and `lazy-once` seem pretty straightforward and valuable; we *might* also be able to implement `weak` but I'm less sure about that one.) https://webpack.js.org/api/module-methods/#requirecontext

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
```diff
@@ -42,6 +42,22 @@ const opts = {
   dependencyMapName: null,
 };
 
+it(`collects require context arguments`, () => {
+  const ast = astFromCode(`
+  const a = require.context('./', true, /foobar/);
```
motiz88, line 47: Let's also test the cases where the optional parameters are omitted.

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -251,6 +271,70 @@ function collectDependencies<TSplitCondition = void>(
   };
 }
 
+/** Extract args passed to `require.context` method. Exposed for testing. */
```
motiz88, line 274: > Exposed for testing

We don't seem to be using that here (and I don't think we should - better to write full test cases against `collectDependencies`)

motiz88 (changes requested):
This is a good start! I have some comments, hope I didn't miss anything major in this first pass.

We also need `registerDependency` to distinguish between `require.context` and other types of dependencies, and between `require.context` calls with different parameters.

Each of these needs to be registered as a separate dependency (even if they all occur in the same file):

```js
require('./foo');
require.context('./foo');
require.context('./foo', true, /something/);
require.context('./foo', false, /something/);
require.context('./foo', false, /something/, 'lazy');
```

packages/metro/src/ModuleGraph/worker/collectDependencies.js
EvanBacon: oops, this is from an earlier pass -- removing.

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
EvanBacon: I figured there could be some sort of debugging benefit

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -251,6 +289,98 @@ function collectDependencies<TSplitCondition = void>(
...
+}
+
+function getContextMode(mode: any): ContextMode {
+  if (
+    typeof mode === 'string' &&
+    ['sync', 'eager', 'lazy', 'lazy-once'].includes(mode)
+  ) {
+    return ((mode: any): ContextMode);
+  }
+  throw new Error(`require.context "${mode}" mode is not supported.`);
```
motiz88, line 357: This is a user-facing error - let's throw it with an AST path attached (e.g. reuse `InvalidRequireCallError`).

Side note: The error handling in this file could be better overall - i.e. it would be best to attach the path of the `mode` argument as well as a specific message, which `InvalidRequireCallError` doesn't allow; feel free to make that type of change here if you want to, but at minimum let's ensure we are propagating a path like we do with other errors.

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -251,6 +289,98 @@ function collectDependencies<TSplitCondition = void>(
...
+      name: directory,
+      // Capture the matching context
+      contextParams,
+      asyncType: null,
+      optional: isOptionalDependency(directory, path, state),
+    },
+    path,
+  );
+
+  transformer.transformSyncRequire(path, dep, state);
```
motiz88, line 381:
```suggestion
  // require() the generated module representing this context
  transformer.transformSyncRequire(path, dep, state);
```

motiz88 (changes requested):
See comments; we also still need make some [changes to `DefaultModuleDependencyRegistry.registerDependency`](https://github.com/facebook/metro/pull/821#pullrequestreview-983807629) + add tests for those.

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -251,6 +289,98 @@ function collectDependencies<TSplitCondition = void>(
...
+  if (args.length > 3) {
+    const argNode = args[3].node;
+    if (argNode.type === 'StringLiteral') {
+      mode = getContextMode(argNode.value);
+    } else {
+      // TODO: Handle `new RegExp(...)` -- `argNode.type === 'NewExpression'`
+      throw new InvalidRequireCallError(path);
+    }
+  }
+
```
motiz88, line 339: Let's also be strict about unknown arguments.

```suggestion

  if (args.length > 4) {
    throw new InvalidRequireCallError(path);
  }

```

packages/metro-transform-worker/src/index.js
```diff
@@ -99,6 +99,8 @@ export type JsTransformerConfig = $ReadOnly<{
   unstable_disableModuleWrapping: boolean,
   unstable_disableNormalizePseudoGlobals: boolean,
   unstable_compactOutput: boolean,
+  /** Enable `require.context` statements which can be used to import multiple files in a directory. */
+  unstable_allowRequireContext?: boolean,
```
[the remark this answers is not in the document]
EvanBacon: Should I add `unstable_allowRequireContext` to the key creation array?

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -199,6 +216,27 @@ function collectDependencies<TSplitCondition = void>(
...
+      if (
+        // Feature gate, defaults to `false`.
+        state.unstable_allowRequireContext &&
+        callee.type === 'MemberExpression' &&
+        callee.object &&
+        // `require`
+        callee.object.type === 'Identifier' &&
+        callee.object.name === 'require' &&
+        // `context`
+        callee.property &&
```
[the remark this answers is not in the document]
EvanBacon: Sorry, getting used to flow again

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -251,6 +289,98 @@ function collectDependencies<TSplitCondition = void>(
...
+    directory,
+    {
+      recursive,
+      filter,
+      mode,
+    },
+  ];
+}
+
+function getContextMode(mode: any): ContextMode {
```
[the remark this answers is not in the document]
EvanBacon: I'm not entirely sure if this exists in Flow, but in TypeScript I'd do something like:
```ts
function isContextMode(mode: any): mode is ContextMode {
  return ...
}
```
Given the strict check, it's common to just pass `any`.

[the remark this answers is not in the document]
motiz88: 1. Flow has [predicate functions](https://flow.org/en/docs/types/functions/#toc-predicate-functions) which are a limited version of this, but [enums](https://flow.org/en/docs/enums/) are generally superior anyway.
2. In idiomatic Flow code `any` is a last resort. Replacing `string` with `any` here loses information and allows Flow to catch fewer bugs - consider that you could accidentally call `getContextMode` with the wrong property or (sometimes) a property that doesn't even exist, and you'd only get an error at runtime.

packages/metro-transform-worker/src/index.js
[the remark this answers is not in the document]
motiz88: No, it gets hashed along with the other properties [here](https://github.com/facebook/metro/blob/f26088df99ce3cb0a45c111cee9c86ed3e1532df/packages/metro-transform-worker/src/index.js#L752).

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -350,14 +365,23 @@ function getRequireContextArgs(
   ];
 }
 
-function getContextMode(mode: any): ContextMode {
+function getContextMode(
+  path: NodePath<CallExpression>,
+  mode: string,
+): ContextMode {
   if (
     typeof mode === 'string' &&
```
motiz88, line 373: No longer necessary now that `mode` is typed as `string`.

```suggestion
```

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -251,6 +288,127 @@ function collectDependencies<TSplitCondition = void>(
...
+
+  let directory: string;
+  if (!Array.isArray(args) || args.length < 1) {
+    throw new InvalidRequireCallError(path);
+  } else {
+    const argNode = args[0].node;
+    if (argNode.type !== 'StringLiteral') {
+      throw new InvalidRequireCallError(
+        argNode,
+        `First argument of \`require.context\` should be a string denoting the directory to require, instead found node of type: ${argNode.type}.`,
```
motiz88, line 305: `InvalidRequireCallError`'s [constructor](https://github.com/facebook/metro/blob/57108f84338017b0c7a2d4bc38a3a3cc814df807/packages/metro/src/ModuleGraph/worker/collectDependencies.js#L582-L590) doesn't take a second argument, I assume you're planning to change it? (Comment mostly for my benefit so I don't forget in the final review pass)

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
```diff
@@ -42,45 +43,206 @@ const opts = {
...
 
-it(`collects require context arguments`, () => {
-  const ast = astFromCode(`
-  const a = require.context('./', false, /foobar/m, 'eager');
-  const b = require.context('./baz');
-`);
-  const {dependencies, dependencyMapName} = collectDependencies(ast, {
-    ...opts,
-    unstable_allowRequireContext: true,
+describe(DefaultModuleDependencyRegistry, () => {
```
motiz88, line 46:
```suggestion
describe('DefaultModuleDependencyRegistry', () => {
```

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -103,9 +104,26 @@ export type CollectedDependencies<+TSplitCondition> = $ReadOnly<{
...
+   * require('./foo');
+   * require.context('./foo');
+   * require.context('./foo', true, /something/);
+   * require.context('./foo', false, /something/);
+   * require.context('./foo', false, /something/, 'lazy');
+   * ```
+   *
+   * This method should be utilized by `registerDependency`.
+   */
+  getKeyForDependency(qualifier: ImportQualifier): string;
```
motiz88, line 126: I would not expose `getKeyForDependency` in the registry interface; it's already part of the `registerDependency` method's responsibilities so it should just happen internally. If anything does need to know the key `collectDependencies` derived for a dependency, we will likely bake it into the `InternalDependency` data structure rather than expose a separate method.

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
motiz88: I see the appeal of testing the dependency registry in isolation, but I would prefer that we write the tests in terms of `collectDependencies`. I'd go so far as to say that the entire concept of a pluggable dependency registry may be redundant, and I'll look into killing it after this PR lands. Selfishly, I'd appreciate having tests that I don't have to rewrite when I do that :D

packages/metro/src/ModuleGraph/worker/collectDependencies.js
motiz88: (I get that this is exposed at least in part for testing, but see my comment on the dependency registry tests)

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -709,13 +732,36 @@ function createModuleNameLiteral(dependency: InternalDependency<mixed>) {
...
+  static getKeyForDependency(qualifier: ImportQualifier): string {
+    let key = qualifier.name;
+
+    // Add extra qualifiers when using `require.context` to prevent collisions.
+    if (qualifier.contextParams) {
+      // NOTE(EvanBacon): Keep this synchronized with `RequireContextParams`, if any other properties are added
+      // then this key algorithm should be updated to account for those properties.
+      // Example: `./directory__true__/foobar/m__lazy`
+      key += [
+        '',
```
motiz88, line 745: Let's future-proof slightly to allow us to distinguish between `require.context` and other things that may need extra qualifiers.

```suggestion
        '',
        'context',
```

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -750,6 +796,9 @@ class DefaultModuleDependencyRegistry<TSplitCondition = void>
   }
 }
 
+collectDependencies.DefaultModuleDependencyRegistry =
+  DefaultModuleDependencyRegistry;
+
```
motiz88, line 801: See my comment on the dependency registry tests - I don't think we need this.

motiz88 (changes requested):

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
EvanBacon: I can write tests at both levels. Feel free to delete the lowers level ones later. 

I exposed/tested this function in particular since I can see people extending the collector and wanting to reuse the key resolver.

packages/metro/src/ModuleGraph/worker/collectDependencies.js
motiz88: I would still prefer it if you turned this into a local function in the file instead of a method on the interface, and tested its behaviour via `collectDependencies`. Having `getKeyForDependency` on the interface will require us to update all the implementations of `ModuleDependencyRegistry`, including the mock (which is failing CI) and the one we have internally at Meta (which will delay landing this). This makes the abstraction unnecessarily leaky (and harder to delete, which as I've mentioned I want to do soon).

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
```diff
@@ -42,6 +43,219 @@ const opts = {
   dependencyMapName: null,
 };
 
+describe('DefaultModuleDependencyRegistry', () => {
```
motiz88, line 46: Can you rewrite these tests in terms of `collectDependencies`?

motiz88 (changes requested):
Can you fix the Flow errors and failing tests, plus update the PR summary? That plus the last comment on the tests, and I think we're ready to land this.

packages/metro/src/DeltaBundler/types.flow.js
```diff
@@ -18,6 +18,19 @@ export type MixedOutput = {
...
 
+// TODO: Convert to a Flow enum
+export type ContextMode = 'sync' | 'eager' | 'lazy' | 'lazy-once';
+
+/** Context for requiring a collection of modules. */
+export type RequireContextParams = $ReadOnly<{
+  /* Should search for files recursively. Optional, default `true` when `require.context` is used */
+  recursive: boolean,
+  /* Filename filter pattern for use in `require.context`. Optional, default `/^\.\/.*$/` (any file) when `require.context` is used */
+  filter: {pattern: string, flags?: string},
```
motiz88, line 29:
```suggestion
  filter: $ReadOnly<{pattern: string, flags: string}>,
```

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
```diff
@@ -32,94 +32,24 @@ const nullthrows = require('nullthrows');
...
-  });
-});
-
 describe(`require.context`, () => {
   it(`collects require context arguments`, () => {
     const ast = astFromCode(`
   const a = require.context('./', false, /foobar/m, 'eager');
   const b = require.context('./baz');
+  const c = require.context('./dir', true);
+  const d = require.context('./dir');
```
motiz88, line 52: Needs a few more test cases, e.g. what happens when we have the same directory etc but different values for `recursive`, plus everything else that can cause the dependency key to be different. (This will be a lot, so ideally these would be separate test cases with clear descriptions as opposed to just additional lines in this one test case.)

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
```diff
@@ -32,94 +32,24 @@ const nullthrows = require('nullthrows');
...
-  collectDependencies;
+const {InvalidRequireCallError} = collectDependencies;
 const opts = {
   asyncRequireModulePath: 'asyncRequire',
   dynamicRequires: 'reject',
   inlineableCalls: [],
   keepRequireNames: true,
   allowOptionalDependencies: false,
   dependencyMapName: null,
+  unstable_allowRequireContext: true,
```
motiz88, line 43: 1. Let's set this to `false` for the test and then override it for the `require.context` tests. (I think we're actually doing that in the test below.)
2. We could use a test of what happens when `require.context` is found in the input when this is `false` - essentially a test that this feature isn't accidentally on for everyone. (We could later also implement a nice error message saying that this feature isn't enabled, and have that reflected in the test.)

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -37,6 +37,18 @@ export type Dependency<TSplitCondition> = $ReadOnly<{
   name: string,
 }>;
 
+// TODO: Convert to a Flow enum
+type ContextMode = 'sync' | 'eager' | 'lazy' | 'lazy-once';
+
+type RequireContextParams = $ReadOnly<{
```
motiz88, line 43: Seems like we have this definition both here and in `types.flow.js` - we probably need just one of the two?

motiz88 (changes requested):
This is really close - some nits, and I would love some more test coverage to ensure all the various permutations of `getKeyForDependency` are handled correctly, but seems basically good to go from an implementation standpoint.

packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
EvanBacon: imo we had this before https://github.com/facebook/metro/pull/821/commits/7c19f6e244611b114b6aa72e1ef996ac4e9f5acd not sure how to proceed short of recreating those tests.

motiz88: Those tests were in terms of `DefaultModuleDependencyRegistry`. I'm thinking instead we should clone the whole `collects require context arguments` test multiple times, so that each test uses `collectDependencies` and covers a particular scenario/edge case in isolation:

1. contexts differ only by directory
2. contexts differ only by `recursive`
3. contexts differ only by regex pattern
4. contexts differ only by regex flags
5. contexts differ only by `mode`
6. default values in case of `undefined` or missing arguments
7. multiple contexts that are identical
8. contexts differ only by trailing slash in directory (e.g `./dir` vs `./dir/`)
9. ...

motiz88: Also, importantly, a test for `require('./')` and `require.context('./')` being distinct.

PR #822, opened 2022-05-28
feat: add `require.context` support

EvanBacon:
Continued work for adding `require.context` to Metro, which started in https://github.com/facebook/metro/pull/821. The overall design is discussed in https://github.com/microsoft/rnx-kit/discussions/1515. The user-facing API is closely modelled on [Webpack's `require.context`](https://webpack.js.org/guides/dependency-management/#requirecontext).

**This feature is experimental and unsupported.** We will document and announce this for general consumption once it's stable.

## How it works

When we encounter a dependency that has `contextParams` (see https://github.com/facebook/metro/pull/821), we will now generate a corresponding *virtual module* in the dependency graph, with dependencies on every file in the file map that matches the context params.

```
┌─────────┐  require.context('./ctx', ...)   ┌−−−−−−−−−−−−−−−−−−┐     ┌──────────┐
│ /bundle │ ───────────────────────────────▶ ╎ <virtual module> ╎ ──▶ │ /ctx/bar │
└─────────┘                                  └−−−−−−−−−−−−−−−−−−┘     └──────────┘
                                               │
                                               │
                                               ▼
                                             ┌──────────────────┐
                                             │     /ctx/foo     │
                                             └──────────────────┘
```

Crucially, we keep this context module up-to-date as file change events come in, so that HMR / Fast Refresh continues to work reliably and transparently. This accounts for the bulk of the complexity of the implementation and tests.

## Main pieces of the implementation

* [collectDependencies] Done in https://github.com/facebook/metro/pull/821: Extract `require.context` calls as dependencies with added metadata. (Behind a flag)
* [metro-file-map] HasteFS: API for querying the file map for the set of files that match a particular resolved context.
* [DeltaBundler] DeltaCalculator: Logic to mark a previously-generated context module as dirty when the set of files it matches has changed.
* [DeltaBundler] graphOperations: Logic to "resolve" context params to a virtual module stored in the graph object, and forward the resolved params to the transformer to generate the module's body and dependencies.
* [DeltaBundler] graphOperations: API for querying the graph for the set of currently active contexts that match a particular file. (Used by DeltaCalculator)
* [metro] transformHelpers, contextModuleTemplates: Logic to generate the contents of a virtual context module by querying the file map. Includes various templates to implement the different `require.context` modes.
* [metro-runtime] `require()`: A stub for `require.context` that helps us throw a meaningful error if the feature is not enabled.

Tests:
* `require-context-test`: a new integration test suite that builds and executes various bundles that use `require.context`. In particular this covers the subset of the `require.context` runtime API that we support.
* `DeltaCalculator-context-test`: a new test suite covering the changes to DeltaCalculator that are specific to `require.context` support.
* `traverseDependencies-test`: expanded and refactored from its previous state. Covers the changes to graphOperations.

## Future work

At the moment, every time we invalidate a context module, we scan the entire file map to find the up-to-date matches and then regenerate the module from scratch (including passing it through the transformer).

Two open areas of investigation are:
1. Optimising the *initial* scan over the file map - e.g. representing it as a path tree to drastically limit the number of files we need to match against.
2. Optimising the incremental case - e.g. directly using the file addition/deletion events we receive from the file map to update the generated module in-place.

At least (2) is essential before we treat this feature as stable.

There's also room to generalise the "virtual modules" concept/infrastructure here to support other use cases. For now everything is very much coupled to the `require.context` use case.

@EvanBacon's original PR summary follows.

----

<details>

## Summary

- Continued work for adding `require.context` to Metro https://github.com/facebook/metro/pull/821
- This PR adds the ability to match files given a "require context". This is similar to the existing method for matching against a regex, but this enables users to search upwards in a directory, search shallow, and match a regex filter.
- Adds ability to pass a `Buffer` to `transformFile` as a sort of virtual file mechanism. This accounts for most the changes in `packages/metro/src/DeltaBundler/Transformer.js`, `packages/metro/src/DeltaBundler/Worker.flow.js`, `packages/metro/src/DeltaBundler/WorkerFarm.js`.
- Since we collapse `require.context` to `require` I've also added a convenience function in dev mode which warns users if they attempt to access `require.context` without enabling the feature flag.
- Made `DeltaCalculator` aware of files being added.
- `graphOperations` has two notable changes:
  1. When resolving dependencies with context, we attach a query to the absolute path (which is used for indexing), this query has a hex hash of the context -- used hex instead of b64 for filename safety.
  2. We pass the require context to `processModule` and inside we transform the dependency different based on the existence of a context module.
- When collecting the delta in `_getChangedDependencies` we now iterate over added and deleted files to see if they match any context modules. This is only enabled when the feature flag is on.
- In `packages/metro/src/lib/contextModule.js` we handle a number of common tasks related to context handling.
- In `packages/metro/src/lib/contextModuleTemplates.js` we generate the virtual modules based on the context. There are a number of different modules that can be created based on the `ContextMode`. I attempted to keep the functionality here similar to Webpack so users could interop bundlers. The most notable missing feature is `context.resolve` which returns the string path to a module, this doesn't appear to be supported in Metro. This mechansim is used for ensuring a module must be explicitly loaded by the user. Instead I wrapped the require values in getters to achieve the same effect.
- We implement the `lazy` mode as well but this requires the user to have async bundles already setup, otherwise the module will throw a runtime error.

### Notice

I've withheld certain optimizations in order to keep this PR simple but still functional. We will want to follow up with a delta file check to require context so we aren't iterating over every file on every update. This feature can be seen in my [test branch](https://github.com/facebook/metro/compare/main...EvanBacon:%40evanbacon/context/graph-resolver).

In my [test branch](https://github.com/facebook/metro/compare/main...EvanBacon:%40evanbacon/context/graph-resolver?expand=1), I built the feature on top of #835 so I know it works, should only require minor changes to graphOperations to get them in sync.

## Test plan

- [ ] Unit tests -- pending @motiz88 approving the implementation.

### Local testing

`metro.config.js`
```js

const path = require("path");

config.watchFolders = [
  path.resolve(__dirname),
  path.resolve(__dirname, "../path/to/metro"),
];

config.transformer = {
  // `require.context` support
  unstable_allowRequireContext: true,
};

module.exports = config;
```

index.js
```tsx
console.log(require.context("./").keys());
```

Start a dev server, when changes to the file system occur, they should be reflected as automatic updates to the context. This does lead to the issue of having multiple contexts triggering multiple sequential reloads, I don't think this is a blocking issue though.

Also tested with async loading enabled, and the context:  `require.context("./", undefined, undefined, 'lazy')`.

### Behavior

Notable features brought over to ensure this `require.context` functions as close to the original implementation as possible:

- `require.context` does not respect platform extensions, it will return every file matched inside of its context.
- `require.context` can match itself.
- A custom 'empty' module will be generated when no files are matched, this improves the user experience.
- All methods in `require.context` are named to improve the debugging.
- We always match against a `./` prefix. This prefix is also returned in the keys.
- Modules must be loaded by invoking the context as a function, e.g. `context('./module.js')`

</details>


packages/metro/src/node-haste/DependencyGraph.js
```diff
@@ -223,6 +223,19 @@ class DependencyGraph extends EventEmitter {
...
 
+  /** Given a search context, return a list of file paths matching the query. */
+  resolveContext(
+    from: string,
+    context: {
+      /* Should search for files recursively. Optional, default `true` when `require.context` is used */
+      recursive: boolean,
+      /* Filename filter pattern for use in `require.context`. Optional, default `/^\.\/.*$/` (any file) when `require.context` is used */
+      filter: RegExp,
+    },
```
motiz88, line 234: Nits:

1. Make read only
2. Drop defaults and optionality which don't apply here.
 
```suggestion
    context: $ReadOnly<{
      /* Should search for files recursively. */
      recursive: boolean,
      /* Filter relative paths against a pattern. */
      filter: RegExp,
    }>,
```

packages/metro/src/node-haste/DependencyGraph.js
```diff
@@ -223,6 +223,19 @@ class DependencyGraph extends EventEmitter {
     this._haste.end();
   }
 
+  /** Given a search context, return a list of file paths matching the query. */
+  resolveContext(
```
motiz88, line 227: This implementation doesn't resolve the paths (as in "module resolution"), so let's reflect that in the name.

```suggestion
  matchFilesWithContext(
```

packages/metro-file-map/src/HasteFS.js
```diff
@@ -84,6 +84,37 @@ export default class HasteFS {
...
 
+  /** Given a search context, return a list of file paths matching the query. */
+  matchFilesWithContext(
+    root: Path,
+    context: {
+      /* Should search for files recursively. */
+      recursive: boolean,
+      /* Filter files against a pattern. */
+      filter: RegExp,
+    },
```
motiz88, line 95:
```suggestion
    context: $ReadOnly<{
      /* Should search for files recursively. */
      recursive: boolean,
      /* Filter relative paths against a pattern. */
      filter: RegExp,
    }>,
```

motiz88 (changes requested):
I would fold this into the upcoming PR which uses this to implement `require.context`, to make sure any edge cases are ironed out & tested - I don't want to have to ship breaking changes to `DependencyGraph` at that point or integration bugs to slip through. Overall looks great though.

packages/metro/src/lib/contextModuleTemplates.js
```diff
@@ -0,0 +1,138 @@
...
+        // Ensure relative file paths start with `./` so they match the
+        // patterns (filters) used to include them.
+        if (!filePath.startsWith('.')) {
+        filePath = `.${path.sep}` + filePath;
+        }
+        const key = JSON.stringify(filePath);
+        // NOTE(EvanBacon): Webpack uses `require.resolve` in order to load modules on demand,
+        // Metro doesn't have this functionality so it will use getters instead. Modules need to
+        // be loaded on demand because if we imported directly then users would get errors from importing
+        // a file without exports as soon as they create a new file and the context module is updated.
```
motiz88, line 28: > Errors from importing a file without exports

What errors? Are these built into Metro or are they a result of application logic?

Does Webpack not eagerly evaluate the inner `require`s as part of the `require.context()` call?

packages/metro/src/lib/contextModuleTemplates.js
```diff
@@ -0,0 +1,138 @@
...
+        return getEmptyContextModuleTemplate(modulePath, id);
+    }
+    switch (mode) {
+        case 'eager':
+        return getLoadableContextModuleTemplate(
+            modulePath,
+            files,
+            id,
+            // NOTE(EvanBacon): It's unclear if we should use `import` or `require` here so sticking
+            // with the more stable option (`require`) for now.
```
motiz88, line 110: My guess is `require`, but the deciding factor is what Webpack does.

packages/metro/src/lib/transformHelpers.js
```diff
@@ -103,6 +123,60 @@ function removeInlineRequiresBlockListFromOptions(
...
+          config.resolver.assetExts,
+        ),
+        inlineRequires: removeInlineRequiresBlockListFromOptions(
+          modulePath,
+          inlineRequires,
+        ),
+      },
+      Buffer.from(template),
+    );
+  };
```
motiz88, line 177: Could we reuse `getTransformFn` for some of this instead of forking it?

packages/metro/src/lib/transformHelpers.js
```diff
@@ -60,6 +68,16 @@ async function calcTransformerOptions(
   const getDependencies = async (path: string) => {
     const dependencies = await deltaBundler.getDependencies([path], {
       resolve: await getResolveDependencyFn(bundler, options.platform),
+      transformContext: await getTransformContextFn(
```
motiz88, line 71: I'm not 100% sure we want a `transformContext` function here, especially if we're eventually looking to make `DeltaBundler` smarter in terms of updating the context modules  incrementally instead of rescanning the whole file map every time. Might be better to bake the context behaviour into DeltaBundler (while making some aspects of it configurable here if necessary).

motiz88 (changes requested):
@EvanBacon this seems broadly in line with what we've discussed. Here's a round of comments. I would love to take another look once there are tests for the new functionality and CI is passing.

packages/metro/src/DeltaBundler/Transformer.js
```diff
@@ -119,15 +122,34 @@ class Transformer {
...
+        );
+
+    data.result.dependencies.forEach(dependency => {
+      if (dependency.data.contextParams) {
+        // Convert JSON regular expression into RegExp.
+        dependency.data.contextParams = toRequireContext(
+          dependency.data.contextParams,
+        );
+      }
+    });
```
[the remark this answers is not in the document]
EvanBacon: We are converting the Regex from a serialized JSON object into a Regex class, this seemingly has to be done after the worker returns the JSON results.

packages/metro/src/lib/contextModuleTemplates.js
EvanBacon: > Does Webpack not eagerly evaluate the inner requires as part of the require.context() call?

No. The context module is seemingly always loaded on demand with `ctx(moduleId)`, the different `mode`s change how the module will be imported when requested (promise, load promise once, require, etc.).

> ...

My comment here is in regards to the edge case that occurs when a file is first created, the module will attempt to import a module with no exports, resulting in an error.

packages/metro-file-map/src/HasteFS.js
```diff
@@ -84,6 +85,48 @@ export default class HasteFS {
...
+    root: Path,
+    context: $ReadOnly<{
+      /* Should search for files recursively. */
+      recursive: boolean,
+      /* Filter relative paths against a pattern. */
+      filter: RegExp,
+    }>,
+  ): Array<Path> {
+    const files = [];
+    const prefix = '.' + path.sep;
```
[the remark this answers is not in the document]
EvanBacon: Appears to be normalized just like module imports.

packages/metro/src/lib/transformHelpers.js
EvanBacon: Dropped this function.

packages/metro/src/DeltaBundler/DeltaCalculator.js
```diff
@@ -180,9 +188,16 @@ class DeltaCalculator<T> extends EventEmitter {
     if (type === 'delete') {
       this._deletedFiles.add(filePath);
       this._modifiedFiles.delete(filePath);
-    } else {
+      this._addedFiles.delete(filePath);
+    } else if (type === 'add') {
```
motiz88, line 192: Are there any tests covering `'add'` events? I don't see any in `DeltaCalculator-test.js`.

packages/metro/src/DeltaBundler/graphOperations.js
```diff
@@ -30,6 +30,7 @@
 
 'use strict';
 
+import type {RequireContextParams} from '../ModuleGraph/worker/collectDependencies';
```
motiz88, line 33: General note - we need extensive tests in `traverseDependencies-test` for all the new functionality in this file - e.g. as the bare minimum we need tests that show:

* how `require.context` is handled in the initial traversal
* how adding a `require.context` in an existing module is reflected in a delta update
* how modifying the set of modules matched by a `require.context` is handled
* how multiple `require.context`s pointing to the same context are handled (from the same origin module, and from multiple origin modules)
* how parallel edges (`require.context` + `require` with the same path) are handled
* what happens to a module that is required *both* directly and via a context (e.g. when we remove the direct `require` and retraverse, does the context keep it in the graph? and vice versa)

packages/metro/src/ModuleGraph/worker/collectDependencies.js
```diff
@@ -39,18 +39,20 @@ export type Dependency<TSplitCondition> = $ReadOnly<{
...
-type ContextMode = 'sync' | 'eager' | 'lazy' | 'lazy-once';
+export type ContextMode = 'sync' | 'eager' | 'lazy' | 'lazy-once';
 
-type ContextFilter = {pattern: string, flags: string};
+export type ContextFilter = {pattern: string, flags: string};
 
-export type RequireContextParams = $ReadOnly<{
+export type RequireContextParams = {
+  /** Absolute file path pointing to the root directory of the context. */
+  from?: string,
```
motiz88, line 48: We should split this data structure - `collectDependencies` will *never* populate `from` with an absolute path because those aren't cacheable, whereas later in the pipeline we need `from` to *always* be present.

motiz88 (changes requested):
Incomplete pass of feedback, will make more time next week.

packages/metro-file-map/src/__tests__/HasteFS-test.js
```diff
@@ -0,0 +1,52 @@
...
+import HasteFS from '../HasteFS';
+
+describe('matchFilesWithContext', () => {
+  it(`matches files against context`, () => {
+    const hfs = new HasteFS({
+      rootDir: '/',
+      files: new Map([]),
+    });
+
+    // $FlowFixMe: mocking files
```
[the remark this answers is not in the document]
EvanBacon: Because I want the test to be focused on `matchFilesWithContext` and this is the path of least resistance. Going through the constructor means the mock paths will be resolved against the file system.

I'd need to mock out path to skip this resolution, which imo creates a more confusing test.

packages/metro/src/DeltaBundler/DeltaCalculator.js
EvanBacon: Seems like [this tests](https://github.com/facebook/metro/blob/419caf75656c373cdb832a45297d40d0c1b19cfb/packages/metro/src/DeltaBundler/__tests__/DeltaCalculator-test.js#L268) adding files already.

packages/metro/src/DeltaBundler/graphOperations.js
EvanBacon: I'm guess "packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js" is the test file for `packages/metro/src/DeltaBundler/graphOperations.js`?

packages/metro/src/DeltaBundler/graphOperations.js
```diff
@@ -455,18 +477,35 @@ function resolveDependencies<T>(
   const maybeResolvedDeps = new Map();
   for (const dep of dependencies) {
     let resolvedDep;
-    try {
+
+    // `require.context`
+    const {contextParams} = dep.data;
+    if (contextParams) {
+      contextParams.from = path.join(parentPath, '..', dep.name);
```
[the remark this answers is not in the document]
EvanBacon: In the default case we resolve `dep.name` as a path component, this assumption builds on how the existing functionality works.

packages/metro-file-map/src/__tests__/HasteFS-test.js
```diff
@@ -0,0 +1,52 @@
...
+ * @format
+ */
+
+import HasteFS from '../HasteFS';
+
+describe('matchFilesWithContext', () => {
+  it(`matches files against context`, () => {
+    const hfs = new HasteFS({
+      rootDir: '/',
+      files: new Map([]),
```
EvanBacon:
```suggestion
      files: new Map([ /* Ignored and mocked out later. */ ]),
```

packages/metro/src/DeltaBundler/graphOperations.js
```diff
@@ -450,6 +459,10 @@ function removeDependency<T>(
     decrementImportBundleReference(dependency, graph);
   }
 
+  if (dependency.data.data.contextParams) {
+    graph.privateState.resolvedContext.delete(dependency.absolutePath);
```
motiz88, line 463: This isn't the right place for this - it should happen in `releaseModule`. That's a bug, because if there's a reference to the same context elsewhere in the graph, we'd end up deleting it while it's still in use.

packages/metro/src/DeltaBundler/graphOperations.js
```diff
@@ -469,7 +482,28 @@ function removeDependency<T>(
   }
 }
 
+/**
+ * Collect a list of context modules which include a given file.
+ */
+function getContextModulesMatchingFilePath<T>(
+  graph: Graph<T>,
+  filePath: string,
+  modifiedFiles: Set<string>,
```
motiz88, line 491:
```suggestion
  modifiedFiles: $ReadOnlySet<string>,
```

motiz88 (changes requested):

motiz88:
I think beyond this point I mainly want to spend some time on the DeltaCalculator tests and see if I can (1) simplify the mocking setup for the existing tests, (2) merge the contest test cases into that file for readability + maintainability, (3) convince myself that all the interesting cases are covered.

packages/metro/src/lib/contextModuleTemplates.js
```diff
@@ -0,0 +1,160 @@
...
+        file,
+      )}; } },\n`;
+    });
+  return `Object.defineProperties({}, {${mapString}})`;
+}
+
+function getEmptyContextModuleTemplate(modulePath: string, id: string): string {
+  return `
+function metroEmptyContext(request) {
+  let e = new Error("No modules for context '" + ${JSON.stringify(id)} + "'");
```
motiz88, line 53: @EvanBacon I'm making minor changes to this code and noticed an inconsistency - an empty context module will throw here, but trying to get a missing module from a non-empty context does not throw. Is this Webpack's behaviour as well?

motiz88: Ideally I would like to remove this ID from the body of the module entirely - otherwise we run into machine-independence issues (Windows/POSIX paths, paths needing to be relative to the project root...) which can cause all sorts of downstream problems with caching etc.

For this kind of diagnostic, we should instead pass the numeric `module.id` into a dedicated helper in `require.js` where we have access to module system metadata (such as paths) that module bodies cannot generally access.

packages/metro/src/DeltaBundler/__tests__/traverseDependencies-test.js
```diff
@@ -1391,6 +1400,138 @@ describe('edge cases', () => {
...
+    // Create a context module
+    Actions.addDependency('/bundle', '/ctx', undefined, undefined, {
+      contextParams: {
+        recursive: true,
+        mode: 'sync',
+        filter: {pattern: '.*', flags: ''},
+      },
+    });
+
+    Actions.createFile('/ctx?ctx=7855fe0b1074e361e66650cb2e83816836dc652a');
```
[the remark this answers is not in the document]
motiz88: Ah, I see why this was necessary - tweaked the mocks so that we only need to `createFile` when it semantically makes sense.

packages/metro/src/DeltaBundler/__tests__/DeltaCalculatorContext-test.js
```diff
@@ -0,0 +1,362 @@
...
+    expect(traverseDependencies).toHaveBeenCalledTimes(1);
+    expect(traverseDependencies.mock.calls[0][0]).toEqual(['/bundle']);
+  });
+
+  it('should not do unnecessary work when adding a context module file after deleting it', async () => {
+    await deltaCalculator.getDelta({reset: false, shallow: false});
+
+    // First delete a file
+    fileWatcher.emit('change', {
+      eventsQueue: [{type: 'delete', filePath: '/ctx?ctx=xxx'}],
```
motiz88, line 343: I don't get what this test is trying to simulate: the filesystem watcher will never emit events for `/ctx?ctx=xxx` because there is no such file on disk. I'll try to rewrite this test case to cover a more realistic scenario.

motiz88:
I've come to believe we should merge the existing DeltaCalculator + traverseDependencies test suites. It can be done fairly mechanically by rewriting the traverseDependencies tests we have in terms of a DeltaCalculator instance, but it's a chance to also rethink the way we mock out the transformer/filesystem and hopefully simplify things some more. Anyway, I won't block this PR on that.

packages/metro-file-map/src/HasteFS.js
```diff
@@ -103,3 +150,10 @@ export default class HasteFS {
     return this._files.get(relativePath);
   }
 }
+
+function normalizeSlashes(path: string): string {
+  if (/^\\\\\?\\/.test(path) || /[^\u0000-\u0080]+/.test(path)) {
```
robhogan, line 155: I think this needs a comment and/or a function rename - I don't think it's very clear what the intention is. 

It looks like the first test checks for [Win32 file namespaces](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces), but I'm not entirely sure why.

The second test ensures that we return the original path if it contains non-ASCII-128 characters, right? So `'foo\bar'` -> `'foo/bar'` but `foo\bā` -> `foo\bā` - on the face of it that doesn't seem i18n friendly - why the different treatment?

packages/metro/src/DeltaBundler/DeltaCalculator.js
```diff
@@ -10,14 +10,15 @@
...
 'use strict';
 
-import type {DeltaResult, Graph, Options} from './types.flow';
-
-const {
+import {markModifiedContextModules} from './graphOperations';
+import {
   createGraph,
   initialTraverseDependencies,
   reorderGraph,
```
robhogan, line 17: (nit)
```suggestion
import {
  createGraph,
  initialTraverseDependencies,
  markModifiedContextModules,
  reorderGraph,
```

packages/metro/src/lib/contextModule.js
```diff
@@ -0,0 +1,94 @@
...
+  );
+}
+
+/** Match a file against a require context. */
+export function fileMatchesContext(
+  testPath: string,
+  context: RequireContext,
+): boolean {
+  // NOTE(EvanBacon): Ensure this logic is synchronized with the similar
+  // functionality in `metro-file-map/src/HasteFS.js` (`matchFilesWithContext()`)
```
robhogan, line 65: (Same comments apply as there)

packages/metro/src/lib/contextModuleTemplates.js
```diff
@@ -0,0 +1,148 @@
...
+
+  files
+    .slice()
+    // Sort for deterministic output
+    .sort()
+    .forEach(file => {
+      let filePath = path.relative(modulePath, file);
+
+      // NOTE(EvanBacon): I'd prefer we prevent the ability for a module to require itself (`require.context('./')`)
+      // but Webpack allows this, keeping it here provides better parity between bundlers.
```
robhogan, line 29: Yikes, but agreed!

robhogan (commented):
*(Imported PRs need internal review by someone other than the importer, sharing feedback here for transparency)*

Left a few nits and minor comments around path handling that need explanation if not changes, but substantively this looks good to me. Thanks for working on this - must be one of the biggest feature contributions we've ever had!

packages/metro-file-map/src/HasteFS.js
```diff
@@ -85,6 +85,53 @@ export default class HasteFS {
...
+    }>,
+  ): Array<Path> {
+    const files = [];
+    const prefix = './';
+
+    for (const file of this.getAbsoluteFileIterator()) {
+      const filePath = fastPath.relative(root, file);
+
+      const isRelative =
+        filePath && !filePath.startsWith('..') && !path.isAbsolute(filePath);
```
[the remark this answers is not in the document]
EvanBacon: isAbsolute check appears to be extraneous now.

packages/metro-file-map/src/HasteFS.js
EvanBacon: My original intention was to make this path work on windows and fail otherwise, that requirement doesn't appear necessary anymore so I'll simply inline the replacement.

packages/metro-file-map/src/HasteFS.js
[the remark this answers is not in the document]
robhogan: > I'm not sure normalization is required for the the `..` check to work as this paradigm is cross-platform.

If it's not normalised it might not appear at the start of the path - eg `foo/../../bar` is valid but normalises to `../bar`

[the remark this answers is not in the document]
robhogan: > isAbsolute check appears to be extraneous now.

I think it's still necessary for Windows tbh, looks like we get absolute paths back when there's no valid relative path between locations:

```
node -e "console.log(path.win32.relative('C:/foo', 'D:/bar'))";
D:/bar
```

kraenhansen:
I know this PR says

> This feature is experimental and unsupported. We will document and announce this for general consumption once it's stable.

Are there plans to stabilise this? I've opened https://github.com/facebook/react-native/pull/41421 in case it's already there.

robhogan:
The main thing holding us back *was* that the pre-symlink filesystem implementation didn't support enumerating a subtree, so `require.context` was an O(project) transform. Now we have [subtree scans](https://github.com/facebook/metro/commit/4c9b0d2ba160596242b33309889616ffc3784a06#diff-c501fb5253191df1cde600bfdfcecd0357cd8a85f321a3bb7f25ba935dc6e46bL155-R155), and Expo/@EvanBacon have battle-tested the implementation, I don't see any reason not to promote this to stable (and always-on).

karlhorky:
For anyone looking for a copy+paste `metroRequire.d.ts` for TypeScript types for Metro `context.require()` to add to your own project until https://github.com/facebook/react-native/pull/41421 is merged, @EvanBacon has one here:

- https://github.com/EvanBacon/Metro-Context-Modules-Demo/blob/3ad84a12b40c51060ee50d0b0065d02fd8ce3b43/types/metroRequire.d.ts

dominictobias:
This feature is great thanks. But a real shame it doesn't work with anything other than relative paths. Has to be:

```ts
const req = require.context(
  '../../../../../node_modules/@metamask/contract-metadata/images',
  false,
  /\.svg$/,
)
```

Which is not only dependent on the location of the file, but also where a package manager decides to install something in a monorepo (in the root or project). It would be ideal if we could do:

```ts
const req = require.context(
  '@metamask/contract-metadata/images',
  false,
  /\.svg$/,
)
```

PR #823, opened 2022-05-29
Allow Metro to run directly from source

motiz88:
## Summary

Inspired by #696, this change automatically registers Babel to compile Metro's packages on-the-fly if we can detect that we're in the Metro monorepo. This applies in the following cases:

1. Someone `require()`s the `metro` package.
2. Someone runs the Metro CLI. (For example, using the new `yarn start` convenience script.)
3. Metro starts the built-in transform worker.

Each of the entry points needs to be runnable in the target version of Node without being transpiled. In some cases (e.g. `metro/src/index.js` and `metro/src/DeltaBundler/Worker.js`) this required moving the existing code to secondary files and creating trivial wrappers around them.

> **Note for maintainers of advanced integrations:** If you've imported the Metro repo as-is into your own source tree (e.g. as a submodule) and have Jest tests that rely on importing Metro packages from source, you'll need to add the following to the top of your Jest setup file.
> ```js
> require('metro-babel-register').unstable_registerForMetroMonorepo = () => {};
> ```
> This will avoid conflicts between the Metro repo's automatic Babel registration and Jest's own Babel transformer.
> Please note that `unstable_registerForMetroMonorepo` is not a supported API and is not covered by semver.

### Development workflow

1. Create a new folder with an empty `package.json` outside the Metro checkout.
2. `yarn link` the `metro` package from source.
3. `yarn metro`. Observe that Metro starts successfully.
4. Require the `metro` package programmatically:
```
node -e 'console.log(require("metro"))'
{
  runMetro: [AsyncFunction: runMetro],
  loadConfig: [AsyncFunction: loadConfig],
  createConnectMiddleware: [AsyncFunction: createConnectMiddleware],
  runServer: [AsyncFunction],
  runBuild: [AsyncFunction],
  buildGraph: [AsyncFunction],
  attachMetroCli: [Function]
}
```

### (Simulated) release workflow
1. From the Metro checkout root, `yarn run build-clean && yarn run build && lerna run prepare-release`.
2. Create a new folder with an empty `package.json` outside the Metro checkout.
3. `yarn add ../metro/packages/metro` (adjust path accordingly)
4. `yarn metro`. Observe that Metro starts successfully.
5. Ensure (e.g. with grep) that `yarn.lock` does not contain `metro-babel-register` or `@babel/register`, i.e. by definition Metro cannot be requiring them accidentally.


EvanBacon:
For reference, this can be used in an Expo project by adding the following to your `metro.config.js`:

```js
const path = require("path");

config.watchFolders = [
  path.resolve(__dirname),
  // Path to your local metro fork
  path.resolve(__dirname, "../metro"),
];

module.exports = config;
```

Otherwise you'll get the error:
``` 
Bundling failed 14590ms
SHA-1 for file /Users/evanbacon/Documents/GitHub/metro/packages/metro-runtime/src/polyfills/require.js (/Users/evanbacon/Documents/GitHub/metro/packages/metro-runtime/src/polyfills/require.js) is not computed.
         Potential causes:
           1) You have symlinks in your project - watchman does not follow symlinks.
           2) Check `blockList` in your metro.config.js and make sure it isn't excluding the file path.
```

Can also get the following if you don't add everything to the `watchFolders`:

```
Unable to resolve module @babel/runtime/helpers/classCallCheck from /Users/evanbacon/Documents/GitHub/metro/packages/metro-runtime/src/modules/HMRClient.js: @babel/runtime/helpers/classCallCheck could not be found within the project or in these directories:
  ../../metro/node_modules
  ../../node_modules
  ../../../../node_modules
> 1 | /**
  2 |  * Copyright (c) Meta Platforms, Inc. and affiliates.
  3 |  *
  4 |  * This source code is licensed under the MIT license found in the
```

motiz88:
@EvanBacon that's a good callout. Ideally we should make that work with no special config (outside the scope of this PR though). The handling of these injected runtime dependencies (Babel runtime, Metro runtime, Fast Refresh runtime) is problematic to begin with - there are a number of these failure modes, where the actual resolution can be outside of the project, or to the wrong version, etc. We need a better architecture here (e.g. teaching the resolver about runtime modules built into Metro).

motiz88:
> teaching the resolver about runtime modules built into Metro

Discussion: https://github.com/microsoft/rnx-kit/discussions/1582

PR #833, opened 2022-06-17
fix: handle http scenario for pathToSource in _processDebuggerGetScri…

### LINKED ISSUE #832: processDebuggerGetScriptSource in metro-inspector-proxy/src/Device.js is unable to return source
bug

When I use flipper to debug `index.bundle?platform=android&dev=true&minify=false&app=com.rn0682&modulesOnly=false&runModule=true` without enabling the JavaScript source map, it shows
```
ENOENT: no such file or directory, open '/Users/wood/Desktop/RN0682/http:/localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.rn0682&modulesOnly=false&runModule=true'
```

image

And I found 

https://github.com/facebook/metro/blob/1a105ea4f91ff385b2fb5f4c5fcfedab3ec630dc/packages/metro-inspector-proxy/src/Device.js#L488-L510

This always throws an exception and makes me unable to access bundled js file.

`git clone https://github.com/wood1986/RN0682`
`yarn install`
`yarn android`
use flipper to debug

It should not read from this fs if metro server is used

wood1986:
fix #832 

image

### DIFF
--- a/packages/metro-inspector-proxy/package.json
+++ b/packages/metro-inspector-proxy/package.json
@@ -19,6 +19,7 @@
   "dependencies": {
     "connect": "^3.6.5",
     "debug": "^2.2.0",
+    "node-fetch": "^2.2.0",
     "ws": "^7.5.1",
     "yargs": "^15.3.1"
   }
--- a/packages/metro-inspector-proxy/src/Device.js
+++ b/packages/metro-inspector-proxy/src/Device.js
@@ -22,6 +22,7 @@ import type {
 import * as fs from 'fs';
 import * as path from 'path';
 import WS from 'ws';
+import fetch from 'node-fetch';
 
 const debug = require('debug')('Metro:InspectorProxy');
 
@@ -173,10 +174,10 @@ class Device {
       },
     });
 
-    socket.on('message', (message: string) => {
+    socket.on('message', async (message: string) => {
       debug('(Debugger) -> (Proxy)    (Device): ' + message);
       const debuggerRequest = JSON.parse(message);
-      const interceptedResponse = this._interceptMessageFromDebugger(
+      const interceptedResponse = await this._interceptMessageFromDebugger(
         debuggerRequest,
         debuggerInfo,
       );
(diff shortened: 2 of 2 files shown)


motiz88:
Hey @wood1986, thanks for the PR! Unfortunately I don't think this is the right solution - the *client* should perform this fetch, not the inspector proxy (which is part of the server). This only works based on assumptions on external conditions, e.g. that `http://localhost:8081` on the client device happens to be forwarded to port 8081 on the server.

wood1986:
@motiz88 Thanks for the review!!

Based on your hint, I found that 

https://github.com/facebook/react-native/blob/b5ff26b0b97d6cd600bdb9c33af866971ef14f9c/ReactCommon/hermes/inspector/chrome/cli/main.cpp#L169-L185

send a message to metro and this handles the message 

https://github.com/facebook/metro/blob/100ca6ffcc3ab99b82fac87b41be708040be2e9e/packages/metro-inspector-proxy/src/Device.js#L176

In short, the communication is between react native app and metro not flipper and metro. If it is latter (flipper < - > metro), then I agree what you say.

Other than that, this is a websocket communication. I do not see metro server handles any request from `http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.rn0682&modulesOnly=false&runModule=true&cachePrevention=1`

So I do not know how I can make client side fetch the source

wood1986:
I deleted my previous comment because it was wrong.

I have just realized `Debugger.getScriptSource` is a standard Chrome DevTool protocol.

`Flipper < - > Chrome DevTool < - > metro`

Chrome DevTool is just an iframe inside Flipper. Flipper initializes Chrome DevTool. After that, Chrome DevTool will deal with Metro directly and Flipper does not have a role. 

How to tell flipper to hit `http://localhost:8081/index.bundle?platform=android&dev=true&minify=false&app=com.rn0682&modulesOnly=false&runModule=true&cachePrevention=1` directly?

wood1986:
Just ping again in case you forget after the weekend

wood1986:
Any update??

raon0211:
This seems to be fixed from v0.73.0.

https://github.com/facebook/metro/commit/db19b06bdd6d2fbbe109e4f3be4b3af3489c1f1c

robhogan:
Sorry we never got back to this. As above, it was fixed by db19b06bd, later generalised by 6690b393c, and the whole package was moved to RN in 15c993f83.

PR #835, opened 2022-06-17
Use explicit key for tracking dependencies in bundler

motiz88:
Summary:
Changes `collectDependencies` so it is responsible for providing a key with each extracted dependency. This key should be *stable* across builds and when dependencies are reordered, and *must* be locally unique within a given module.

NOTE: This is a similar concept to the [`key`](https://reactjs.org/docs/lists-and-keys.html#keys) prop in React. It's also used for a similar purpose - `traverseDependencies` is not unlike a tree diffing algorithm.

Previously, the `name` property ( = the first argument to `require` etc) *implicitly* served as this key in both `collectDependencies` and DeltaBundler, but since the addition of `require.context` dependency descriptors in https://github.com/facebook/metro/pull/821, this is no longer strictly correct. (This diff therefore unblocks implementing `require.context` in DeltaBundler.)

Instead of teaching DeltaBundler to internally re-derive suitable keys from the dependency descriptors - essentially duplicating the logic in `collectDependencies` - the approach taken here is to require `collectDependencies` to return an *explicit* key as part of each descriptor.

NOTE: Keys should be considered completely opaque. As an implementation detail (that may change without notice), we hash the keys to obfuscate them and deter downstream code from depending on the information in them.

Note that it's safe for multiple descriptors to resolve to a single module (as of D37194640 (https://github.com/facebook/metro/commit/fc29a1177f883144674cf85a813b58567f69d545)), so from DeltaBundler's perspective it's now perfectly valid for `collectDependencies` to not collapse dependencies at all (e.g. even generate one descriptor per `require` call site) as long as it provides a unique key with each one.

WARNING: This diff exposes a **preexisting** bug affecting optional dependencies (introduced in https://github.com/facebook/metro/pull/511) - see FIXME comment in `graphOperations.js` for the details. This will require more followup in a separate diff.

Differential Revision: D37205825

### DIFF
--- a/packages/metro/src/DeltaBundler/Serializers/helpers/__tests__/bytecode-test.js
+++ b/packages/metro/src/DeltaBundler/Serializers/helpers/__tests__/bytecode-test.js
@@ -33,14 +33,14 @@ beforeEach(() => {
         'bar',
         {
           absolutePath: '/bar',
-          data: {data: {asyncType: null, locs: []}, name: 'bar'},
+          data: {data: {asyncType: null, locs: [], key: 'bar'}, name: 'bar'},
         },
       ],
       [
         'baz',
         {
           absolutePath: '/baz',
-          data: {data: {asyncType: null, locs: []}, name: 'baz'},
+          data: {data: {asyncType: null, locs: [], key: 'baz'}, name: 'baz'},
         },
       ],
     ]),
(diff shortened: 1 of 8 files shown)


motiz88:
cc @EvanBacon - see if you can work on top of this branch for your next set of `require.context` changes. I'm unlikely to land it before I'm back from PTO in early July though. The good news: I don't expect to make further changes to this area of DeltaBundler, except maybe to fix the aforementioned bug around optional dependencies.

PR #849, opened 2022-07-26
Fixed typo causing option to not be applied

danilobuerger:
## Summary

Follow up to https://github.com/facebook/metro/commit/1b6dd6f05a18836f169f49a4e3b1730d1467e84a

The option did not get applied as the object was passed to require instead of babel as a result of the `)` being set in the wrong place.

Very sorry to cause you extra work @dmitryrykun @motiz88


dmitryrykun:
Hi @danilobuerger, thank you for the fix! Could you please also fix eslint error from the CI.

danilobuerger:
@dmitryrykun Thanks & fixed!

PR #854, opened 2022-08-19
feat: skip extraneously duplicating AST during transformation

EvanBacon:
## Summary

We split the parse/transform steps up to accommodate Hermes parsing, but this 
defaults to cloning the AST which increases the transformation time by a fair amount.

You get this behavior by default when using Babel's `transform` method directly, which is what other bundlers do.

## Test plan

Bundling [`three.js`](https://unpkg.com/three@0.77.0/three.js) in production mode:

`$ metro build ./example/index.js --out dist/index.js --platform ios --reset-cache --max-workers 0`

Before:
```
babel.transform: 1.961s
```

After:
```
babel.transform: 1.047s
```


EvanBacon:
Digging into the failures. Looks like the `generateFunctionMap` function is mutating the ast and leaving it in a malformed state.

robhogan:
Quick update - we hoped to get this in for last week's release but that wasn't possible. Updating Babel internally is causing a bit of a cascade headache.
 - `@babel/traverse@7.14.7` requires `@babel/parser@^7.14.7` and `@babel/types@^17.14.5`
 - `@babel/parser@7.14.7` breaks on Flow-readonly private class field declarations (`class Foo { +#foo; }`), fixed some time before `7.20`
 - `@babel/types@7.17.12` introduces a `require('..')` that `node-zip` can't resolve.
 - Some combination of updates result in a bunch of (benign, I think) test snapshot changes.
 - A `@babel/types` update breaks our undefined reference detection on Flow enums.
 - ...

There's probably a combination of versions that works with a bit of lockfile hacking, but I'd rather fix the issues and make a clean upgrade to latest (7.20). Working on it..

robhogan:
We've just come out of a holiday infra code freeze, landing those Babel updates now.

robhogan:
The Babel ^7.20 updates have landed, but this PR is now showing failures on public CI - I'm AFK but looks like it could be a Babel bug.

Will try to have a look next week but just FYI in case you get chance in the meantime.

robhogan:
Follow-up:

Looks like the problem is not that the AST is mutated, but that `traverse.cache` holds cached state keyed on AST Node, and that state is dependent on context - when we run through `traverse` for `generateFunctionMappings`, the cache created at `traverse.cache.paths.get(sourceAst)` has no `hub` property and is effectively polluted for Babel's purposes.

I can't find where this was ever fixed - I suspect it might not have been, but "updating" resulted in multiple `@babel/traverse` versions in `node_modules`, and the competing consumers ended up not sharing module state.

This looks like exactly the problem described by @loganfsmyth in the issue above.

A few options come to mind:
1. Work out how to call `traverse` within `generateFunctionMappings` in a way that populates the cache identically to Babel (ie, providing a `hub`). Babel does that [here](https://github.com/babel/babel/blob/859daa3c185d18901106d8e6c39898ba1ce8e8f4/packages/babel-core/src/transformation/file/file.ts#L72-L78).
2. Push/pop the whole `traverse.cache` around `generateFunctionMappings` - same effect as multiple `@babel/traverse` installations, but a bit hacky.
3. Use `traverse.clearNode(sourceAST)` after traversing with `generateFunctionMappings` - this is a simple public API, but it costs an extra traversal.
4. Use `traverse.cache.paths.delete(sourceAST)` to snipe the problematic bit of the cache without clearing the whole tree - this uses a private API and leaves other paths polluted with unknown consequences.

motiz88:
Ah, Babel path caching hell, my old nemesis. Thanks for taking a look @robhogan!

This is also a prime example of why replacing the transformer with something like SWC would be helpful from a correctness standpoint as well as (likely) performance.

robhogan:
I've gone with option 3 - the `clearNode` approach uses a documented API and despite the traversal it performs it's plenty quick enough (<0.1ms for `three.js`) to maintain the overall perf win.

If there are no objections (@motiz88, @EvanBacon) I'll merge this in time for tomorrow's scheduled Metro release.

robhogan:
For visibility, I think this is the least-worst prerequisite: https://github.com/facebook/metro/pull/906 

(Clearing the node was no good either, because it wipes metadata that may be needed upstream - in particular it broke a bunch of our Jest tests because we use this transformer in Jest setup, and Jest relies on path properties.)

vzaidman:
For visibility, the new workaround for that is implemented here: https://github.com/facebook/metro/pull/1340

PR #865, opened 2022-09-06
Resolve aliased module from closest package directory

### LINKED ISSUE #860: metro-resolver bug: resolve logic assumes fromModule name does not have a path separator
metro-resolver bug: resolve logic assumes fromModule name does not have a path separator (e.g., @foo/bar)

expected behavior: resolve logic works regardless of whether fromModule has a path separator

Example package.json:
`{ "name": "@foo/bar", "browser": { "util": "./lib/polyfills/util.js" } }`

Say the @foo/bar package is in /path/to/node_modules/@foo/bar, then `resolve()` will check /path/to/node_modules/@foo for ./lib/polyfills/util.js instead of /path/to/node_modules/@foo/bar. This is because https://github.com/facebook/metro/blob/main/packages/metro-resolver/src/resolve.js#L73 slices the `originModulePath` at the next path separator after node_modules. This does not work for modules with a path separator in the name.

SlyryD:
## Summary

Fixes #860 by resolving realModuleName from the closest package directory.

Example package.json:
`{ "name": "@foo/bar", "browser": { "util": "./lib/polyfills/util.js" } }`

Example values:
originModulePath := "/path/to/node_modules/@foo/bar/lib/index.js"
realModuleName := "./lib/polyfills/util.js"

### Before (wrong originModuleDir)

originModuleDir := "/path/to/node_modules/@foo"

### After (correct originModuleDir)

originModuleDir := "/path/to/node_modules/@foo/bar"

## Test plan

TBD


SlyryD:
@rafeca any feedback here? Looks like the original code went in 5 years ago, so it's interesting I'm the first person to hit it.

robhogan:
Hi @SlyryD, thanks for the report and PR - this is a little surprising because afaik scoped packages have been fully supported for some time. I'll take a look tomorrow but might ask for a repro case.

SlyryD:
Hi @robhogan, thanks for the response! Admittedly I'm having trouble reproing it outside the private repo where I see the issue (which has a different package manager than yarn). Here's my attempt to repro the issue:
1. Clone https://github.com/SlyryD/metro-repro
2. Run `yarn` at root
3. Run `yarn run bundle` at packages/baz
I'm getting a differnt issue currently, which I think has to do with "@rnx-kit/metro-resolver-symlinks". Maybe it's a good starting point for your investigation? Thanks!

SlyryD:
> Hi @robhogan, thanks for the response! Admittedly I'm having trouble reproing it outside the private repo where I see the issue (which has a different package manager than yarn). Here's my attempt to repro the issue:
> 
> 1. Clone https://github.com/SlyryD/metro-repro
> 2. Run `yarn` at root
> 3. Run `yarn run bundle` at packages/baz
> 
> I'm getting a different issue currently, which I think has to do with "@rnx-kit/metro-resolver-symlinks". I'm starting to think that could be the problematic package, though this PR does fix issues in the private repo.

Never mind, I got the repro:
1. Clone https://github.com/SlyryD/metro-repro
2. Run yarn at root
3. Run yarn run bundle at packages/car

SlyryD:
> > Hi @robhogan, thanks for the response! Admittedly I'm having trouble reproing it outside the private repo where I see the issue (which has a different package manager than yarn). Here's my attempt to repro the issue:
> > 
> > 1. Clone https://github.com/SlyryD/metro-repro
> > 2. Run `yarn` at root
> > 3. Run `yarn run bundle` at packages/baz
> > 
> > I'm getting a different issue currently, which I think has to do with "@rnx-kit/metro-resolver-symlinks". I'm starting to think that could be the problematic package, though this PR does fix issues in the private repo.
> 
> I got a repro, but it's a little bit different because it doesn't have to do with scoped packages. That being said, this PR fixes the issue too:
> 
> 1. Clone https://github.com/SlyryD/metro-repro
> 2. Run yarn at root (in main branch)
> 3. Run yarn run bundle at packages/car

Any update now that I've provided a repro

SlyryD:
I provided a repro for an issue above. Any updates? Thanks!

robhogan:
Hi @SlyryD  - I just checked out the repro.

`"browser": { "util": "./lib/polyfills/util.js" }`

This isn't standard `node_modules` resolution, is it? It looks like https://github.com/defunctzombie/package-browser-field-spec#replace-specific-files---advanced , which Metro doesn't support AFAIK.

The PR is also makes some incorrect assumptions about package layout - the `package.json` isn't necessarily in the parent directory of the requiring module and it looks like this would break with nesting, or indeed when the module is in the same directory as the `package.json`.

If the issue only appears when using the `rnx-kit` resolver, I'd start by filing an issue over there, or otherwise reduce the repro to something that demonstrates Metro's `resolveRequest` is behaving incorrectly. If you can do that, please do reopen an issue and we'll work from there. 

I'm going to close the PR as we need to concretely identify an issue and then work out the best fix that doesn't break other cases - thanks for your effort though.

PR #874, opened 2022-09-28
Add configuration for disabling preferNativePlatform

scottmas:
## Summary

Currently, it is impossible to disable `preferNativePlatform`. This PR enables it

## Test plan

In an application directory, create a `blah.native.js` file, then import it in your bundle entry point without its extension. If `preferNativePlatform` is enabled (or not present) in `metro.config.js`, the bundle should succeed. If it is disabled, the bundle should fail. 

## Note
I wasn't able to fully get my local environment working and some tests and flow checks were failing. I also did not update the snapshots, which will need to be refreshed.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -154,6 +154,12 @@ Type: `boolean`
 
 Whether to disable [looking up modules in `node_modules` folders](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders). This only affects the default search through the directory tree, not other Metro options like `extraNodeModules` or `nodeModulesPaths`. Defaults to `false`.
 
+#### `preferNativePlatform`
+
+Type: `boolean`
+
+If set to false, will disable lookups for `.native` file extensions (used in react-native) and will use only extensions specified in `platforms`.
+
 #### `emptyModulePath`
 
 Type: `string`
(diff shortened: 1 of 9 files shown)


huntie:
Hi @scottmas, would you be able to share more context about your use case? There should be one or more ways to work around this currently. The team is reconsidering `.native.js` handling in general, and this input will be valuable! 🙂 In particular, we want to get the design right when adding more config options to Metro.

Separately, please consider updating failing tests in future PRs, as this gives us more confidence about what changed!

scottmas:
Thanks for responding! In general on react-native projects I've found `.native.js` to be quite handy. However, the task I'm currently working on (foolishly or not, I don't know yet) is to make the react-native cli work for _web_ bundles. By setting the `platforms` property in `metro.config.js` I was able to mostly enable it, but `.native.js` files kept getting included in the bundle since the `preferNativePlatform` property always was always enabled.   

I understand why you added the property since handling `.native.js` extensions isn't _quite_ the same as adding an additional platform. But to be honest, a more "correct" approach would likely be to keep `platforms` property as is but then let users specify compatibility for _multiple_ platforms when creating their bundle. E.g. `metro bundle --platform=ios --platform=native`. That said, it would be tricky to make that change backwards compatible since literally everyone in the ecosystem assumes a single platform. So best of luck deciding on your approach.

PR #876, opened 2022-09-28
Normalize file paths for require context on Windows

byCedric:
## Summary

On Windows, the `require.context` resolves modules to non-posix paths. This seems to diverge from the expected behavior compared to Webpack, where the modules are resolved as posix paths. Because of this, we are running into unfortunate issues like https://github.com/expo/router/issues/13.

When inspecting the [`contextModuleTemplates`](https://github.com/facebook/metro/blob/main/packages/metro/src/lib/contextModuleTemplates.js#L44-L46), using [this example repo](https://github.com/byCedric/expo-router-context-windows-issue), you can see this happening.

image

This change normalizes the file path keys to be posix, matching the default Webpack behavior. It does not change the absolute file reference. (based on [this piece of code](https://github.com/facebook/metro/blob/main/packages/metro/src/Assets.js#L201-L204))

## Test plan

- Added a windows style test case, copied from [this test in `metro-file-map`](https://github.com/facebook/metro/blob/main/packages/metro-file-map/src/lib/__tests__/normalizePathSep-test.js#L21-L26)
- Ran the tests on my local Windows machine
  image
- Patched the [test repo](https://github.com/byCedric/expo-router-context-windows-issue) with [this change](https://github.com/byCedric/expo-router-context-windows-issue/blob/main/patches/metro%2B0.72.3.patch) and validated the output on Windows
  image

### DIFF
--- a/packages/metro/src/lib/__tests__/contextModuleTemplates-test.js
+++ b/packages/metro/src/lib/__tests__/contextModuleTemplates-test.js
@@ -49,4 +49,15 @@ describe('getContextModuleTemplate', () => {
 
     expect(template).toMatchSnapshot();
   });
+
+  test('creates posix paths on windows for sync template', () => {
+    jest.resetModules();
+    jest.mock('path', () => jest.requireActual('path').win32);
+    const { getContextModuleTemplate: getWindowsTemplate } = require('../contextModuleTemplates');
+    const template = getWindowsTemplate('sync', 'c:/path/to/project/src', [
+      'C:\\path\\to\\project\\src\\foo.js',
+    ]);
+    expect(template).toMatch(/foo\.js/);
+    expect(template).toMatchSnapshot();
+  });
 });
--- a/packages/metro/src/lib/__tests__/contextModuleTemplates-test.js
+++ b/packages/metro/src/lib/__tests__/contextModuleTemplates-test.js
@@ -53,7 +53,9 @@ describe('getContextModuleTemplate', () => {
   test('creates posix paths on windows for sync template', () => {
     jest.resetModules();
     jest.mock('path', () => jest.requireActual('path').win32);
-    const { getContextModuleTemplate: getWindowsTemplate } = require('../contextModuleTemplates');
+    const {
+      getContextModuleTemplate: getWindowsTemplate,
+    } = require('../contextModuleTemplates');
     const template = getWindowsTemplate('sync', 'c:/path/to/project/src', [
       'C:\\path\\to\\project\\src\\foo.js',
     ]);
--- a/packages/metro/src/lib/contextModuleTemplates.js
+++ b/packages/metro/src/lib/contextModuleTemplates.js
@@ -10,6 +10,8 @@
  */
 
 import * as path from 'path';
+import * as os from 'os';
+import slash from 'slash';
 import type {ContextMode} from '../ModuleGraph/worker/collectDependencies';
 
 function createFileMap(
@@ -34,10 +36,11 @@ function createFileMap(
       if (!filePath.startsWith('.')) {
         filePath = `.${path.sep}` + filePath;
       }
-      // NOTE(byCedric): On Windows, normalize the backslashes to forward slashes to match the behavior in Webpack.
-      if (path.sep === '\\') {
-        filePath = filePath.replace(/\\/g, '/');
+
+      if (os.platform() === 'win32') {
+        filePath = slash(filePath);
       }
+
       const key = JSON.stringify(filePath);
       // NOTE(EvanBacon): Webpack uses `require.resolve` in order to load modules on demand,
       // Metro doesn't have this functionality so it will use getters instead. Modules need to
--- a/packages/metro/src/lib/contextModuleTemplates.js
+++ b/packages/metro/src/lib/contextModuleTemplates.js
@@ -28,17 +28,17 @@ function createFileMap(
     .forEach(file => {
       let filePath = path.relative(modulePath, file);
 
+      if (os.platform() === 'win32') {
+        filePath = slash(filePath);
+      }
+
       // NOTE(EvanBacon): I'd prefer we prevent the ability for a module to require itself (`require.context('./')`)
       // but Webpack allows this, keeping it here provides better parity between bundlers.
 
       // Ensure relative file paths start with `./` so they match the
       // patterns (filters) used to include them.
       if (!filePath.startsWith('.')) {
-        filePath = `.${path.sep}` + filePath;
-      }
-
-      if (os.platform() === 'win32') {
-        filePath = slash(filePath);
+        filePath = `./${filePath}`;
       }
 
       const key = JSON.stringify(filePath);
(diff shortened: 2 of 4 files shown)


packages/metro/src/lib/contextModuleTemplates.js
```diff
@@ -34,6 +34,10 @@ function createFileMap(
       if (!filePath.startsWith('.')) {
         filePath = `.${path.sep}` + filePath;
       }
+      // NOTE(byCedric): On Windows, normalize the backslashes to forward slashes to match the behavior in Webpack.
+      if (path.sep === '\\') {
```
jacdebug, line 38: Lets use `os.platform` here 
```suggestion
      if (os.platform === 'win32') {
```

packages/metro/src/lib/contextModuleTemplates.js
```diff
@@ -34,6 +34,10 @@ function createFileMap(
...
+        filePath = filePath.replace(/\\/g, '/');
```
jacdebug, line 39: We are using npm package slash for this usecase, better reuse the same. 

```suggestion
        filePath = slash(filePath);
```

packages/metro/src/lib/__tests__/contextModuleTemplates-test.js
```diff
@@ -49,4 +49,15 @@ describe('getContextModuleTemplate', () => {
 
     expect(template).toMatchSnapshot();
   });
+
+  test('creates posix paths on windows for sync template', () => {
+    jest.resetModules();
+    jest.mock('path', () => jest.requireActual('path').win32);
+    const { getContextModuleTemplate: getWindowsTemplate } = require('../contextModuleTemplates');
```
jacdebug, line 56: nit: run prettier-format
```suggestion
    const {
       getContextModuleTemplate: getWindowsTemplate,
     } = require('../contextModuleTemplates');
```

jacdebug (changes requested):
Thanks for the PR, looks good some little suggestions.

packages/metro/src/lib/contextModuleTemplates.js
jacdebug: Also by using os.platform you can remove the comment as code itself is self explanatory.

packages/metro/src/lib/contextModuleTemplates.js
huntie: Note that we've used `slash` as a `devDependency` within `metro-file-map`, and it is only referenced once. Therefore probably *not* worth including it here — `filePath.replaceAll(path.sep, '/')` should do.

packages/metro/src/lib/contextModuleTemplates.js
huntie: nit: If we move this logic above the preceding `if` statement, it's slightly cleaner to have normalised `filePath` first, before operating on it (e.g. adding relative path prefix).

packages/metro/src/lib/__tests__/contextModuleTemplates-test.js
byCedric: Done

packages/metro/src/lib/contextModuleTemplates.js
byCedric: Alright, let me revert that.

byCedric: Done

packages/metro/src/lib/contextModuleTemplates.js
byCedric: Does Metro have a standard way of mocking `os.platform()` to be `win32`? I'd prefer to add a test that runs in "windows mode" because I don't think there is a windows CI runner on Metro atm.

jacdebug (approved):

packages/metro/src/lib/contextModuleTemplates.js
huntie: Not AFAIK, but can be achieved with Jest mocks. I think the added unit test makes sense — it's an intentional and direct test against the fork in logic implemented. IMO (weakly held, see below) it's nice that the base tests deal without the concern of POSIX/Windows paths except where we expect specific handling.

At the same time, we would harden test coverage if all tests ran for both path types, however this will not be solved immediately via a separate `win32` test runner — as the input strings to all tests would need to be changed appropriately (and at which point could be handled on a Unix host). I guess we could explore some abstractions, possibly:
- Simple: `describe.each()` (inside an abstraction) across suitable sets of tests for "posix paths" and "win32 paths" which mock relevant `os` and `path` methods in the `win32` case, and the use of `path.normalize()` for every input string in these tests (and snapshot/`path.sep`-aware assertions).
    - What we get: All test cases are run at development time. Tests and snapshot tests are cleanly grouped by path behaviour (where use of `path.normalize()` alone will cause snapshot mismatches across platforms).
- Fancy: Setting up "posix paths" and "win32 paths" **globally** for certain tests matching an extension pattern (e.g. `.multiplatform.test.js`, akin to the multiplatform Jest setup of [`jest-expo/universal`](https://github.com/expo/expo/blob/master/packages/jest-expo/universal/jest-preset.js) — with [output like this](https://user-images.githubusercontent.com/2547783/193242764-aefda859-87bc-49e2-a8d5-2e77166c473c.png)). In this case, we might opt to label these "unix" and "win32 (simulated)" with a wider set of mocked differences.
    - Improvements: More transparent behaviour when opted-in via test filename (no extra `describe` wrapper, snapshots automatically suffixed). Fancy CLI output for test cases (platform tag). Single source of truth for handling in test setup scripts.

With these ideas listed out, you might agree that our existing case-by-case handling is simple and adequate 😅.

Of course, this is not the only functional difference we may encounter on Windows, so I'm in favour of adding a `win32` CI test job as a baseline. cc @motiz88

dmytrorykun:
@byCedric could you please rebase your branch a the latest main

packages/metro/src/lib/contextModuleTemplates.js
```diff
@@ -26,14 +27,19 @@ function createFileMap(
     .forEach(file => {
       let filePath = path.relative(modulePath, file);
 
+      if (os.platform() === 'win32') {
+        filePath = filePath.replaceAll(path.sep, '/');
```
huntie, line 31: Sorry! Just noted that `String.replaceAll` isn't available in our minimum supported version of Node (14.x) — you were good beforehand 😅.

```suggestion
        filePath = filePath.replace(/\\/g, '/');
```

byCedric:
> @byCedric could you please rebase your branch a the latest main

Done! 😄

byCedric:
> Done! @dmytrorykun 😄

Done again! @dmytrorykun 😄

dmytrorykun:
@byCedric could you please rebase your branch on the latest main once again

byCedric:
> @byCedric could you please rebase your branch on the latest main once again

Done, but can we please get this merged before the next change? 😄

PR #915, opened 2023-01-18
Add `resolver.unstable_enableSymlinks` and remove symlinks+watchman restriction

robhogan:
Summary:
**Note: this change does not mean that symlinks are respected by the resolver (yet)!**

Add `resolver.unstable_enableSymlinks` (default: `false`) option to Metro config, and pass it though to the existing `enableSymlinks` option in `metro-file-map`.

At this point this is purely for ease of testing end-to-end behaviour as we incrementally change Fast Refresh and resolution to cope with symlinks.

Removes the Jest-inherited prohibition on using `enableSymlinks` with `useWatchman`, which is no longer necessary or relevant to the new implementation.

**Bikeshed: Why under `resolver`?**
Symlink support is mostly relevant to the resolver, in that if the resolver resolves to a file through a link, that fact is transparent to the rest of Metro. Third-party solutions for symlink-based projects have focused on the resolver. The fact that 90% of the work for this solution is within `metro-file-map` is an implementation detail of in-memory resolution. That said, we'll have an opportunity to revisit this when the config is promoted from `unstable_`

Changlog: [Experimental] - Add `resolver.unstable_enableSymlinks` config for first phase of symlink support.

Reviewed By: jacdebug

Differential Revision: D42552674

### DIFF
--- a/packages/metro-config/src/__tests__/__snapshots__/loadConfig-test.js.snap
+++ b/packages/metro-config/src/__tests__/__snapshots__/loadConfig-test.js.snap
@@ -83,6 +83,7 @@ Object {
       ],
     },
     "unstable_enablePackageExports": false,
+    "unstable_enableSymlinks": false,
     "useWatchman": true,
   },
   "serializer": Object {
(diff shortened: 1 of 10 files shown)


robhogan:
I guess the bot missed it but this landed in https://github.com/facebook/metro/commit/ed5545561d15de39fe6ea999bc34840afb79d8da

PR #926, opened 2023-02-14
TreeFS: Symlink-enabled, prefix-tree implementation of `FileSystem`

robhogan:
Summary:
Currently, `HasteFS` is the only implementation of the `FileSystem` interface, used by Metro to check for file existence during resolution, and to retrieve metadata (file hashes, etc) during transformation. `HasteFS` is based on a simple, performant `Map` of  real paths to metadata.

This diff introduces `TreeFS`, which is based on an in-memory prefix tree representation, split on path separators, of all watched files.

By using a prefix tree, we can lookup files whose paths segments are directory symlinks or files which are themselves links to other files.

We use `TreeFS` in preference to `HasteFS` when `config.resolver.unstable_enableSymlinks` (default: `false`) is `true`.

Reviewed By: motiz88

Differential Revision: D42552999


packages/metro-file-map/src/lib/TreeFS.js
```diff
@@ -0,0 +1,386 @@
...
+    }
+    return files;
+  }
+
+  /**
+   * Given a search context, return a list of file paths matching the query.
+   * The query matches against normalized paths which start with `./`,
+   * for example: `a/b.js` -> `./a/b.js`
+   */
+  matchFilesWithContext(
```
robhogan, line 122: FYI @EvanBacon - this is an O(subtree) implementation of `matchFilesWithContext`, so once TreeFS is the default (a bit of benchmarking + optimisation to do) we'll have addressed [this](https://github.com/facebook/metro/blob/a0ba9945ea0bcb99e2048eb76dafafa03a0e1709/packages/metro/src/lib/transformHelpers.js#L154) and can probably promote `require.context` to stable 🚀

PR #929, opened 2023-02-17
metro-minify-terser: Don't mutate input options

robhogan:
Summary:
Mitigates https://github.com/terser/terser/issues/1341

Terser sets properties inlcuding `source_map` and `_destroy_ast` on the given `options.output` (or `options.format`) object, which can affect subsequent calls where we re-use the same config object.

Here we take a shallow copy of the given configuration, so `terser` doesn't mutate `metro-minify-terser`'s input.

Changelog: [Fix] Mitigate potential source map mismatches with concurrent transformations (Terser [#1341](https://github.com/terser/terser/issues/1341))

Reviewed By: jacdebug, motiz88

Differential Revision: D43362977

### DIFF
--- a/packages/metro-minify-terser/src/__tests__/terser-issue-1341-test.js
+++ b/packages/metro-minify-terser/src/__tests__/terser-issue-1341-test.js
@@ -1,0 +1,92 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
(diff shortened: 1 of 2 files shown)


robhogan:
Merged in https://github.com/facebook/metro/commit/5e27b3b8464f02a47ff6401c23cda0ec6658c9b1

PR #951, opened 2023-03-21
Add content type support to metro server.

aleqsio:
## Summary

I'm a maintainer of `expo-image` on the web platform.
To support rendering svg assets by passing the URI to the `` src prop, it's necessary for the asset url to return the correct content type.

The solution follows the code described in this issue, and was tested to work and allow using SVG assets on web in an expected way.

## Test plan

SVG returned as text without the patch in this PR:
image

SVG rendered correctly on asset URL after applying the patch:
image

Expected content type:
image

### DIFF
--- a/packages/metro/src/Server.js
+++ b/packages/metro/src/Server.js
@@ -456,6 +456,7 @@ class Server {
       if (process.env.REACT_NATIVE_ENABLE_ASSET_CACHING === true) {
         res.setHeader('Cache-Control', 'max-age=31536000');
       }
+      res.setHeader("Content-Type", mime.lookup(path.basename(assetPath)));
       res.end(this._rangeRequestMiddleware(req, res, data, assetPath));
       process.nextTick(() => {
         log(createActionEndEntry(processingAssetRequestLogEntry));


packages/metro/src/Server.js
```diff
@@ -456,6 +456,7 @@ class Server {
       if (process.env.REACT_NATIVE_ENABLE_ASSET_CACHING === true) {
         res.setHeader('Cache-Control', 'max-age=31536000');
       }
+      res.setHeader("Content-Type", mime.lookup(path.basename(assetPath)));
```
motiz88, line 459: We already have this in `_rangeRequestMiddleware` - let's deduplicate the logic to set the header, and ensure range requests still work after this change.

motiz88 (changes requested):
Hey @aleqsio! Thanks for this fix. I have a couple of minor comments but this mostly looks good!

Can you add a test for this under https://github.com/facebook/metro/blob/main/packages/metro/src/Server/__tests__/Server-test.js#L783 ?

aleqsio:
Sorry, codespaces recreated the branch for some reason instead of just ammending it, made a second PR :)

PR #953, opened 2023-03-21
Add missing content type header for assets to metro server.

### LINKED ISSUE #394: Content-Type not set
In metro@0.51.1 src/Server.js sets the correct Content-Type in `_rangeRequestMiddleware` only if the request is a range request.
There is a `res.setHeader("Content-Type", mime.lookup(path.basename(assetPath[1])));` missing just after `process.env.REACT_NATIVE_ENABLE_ASSET_CACHING`

aleqsio:
## Summary

I'm a maintainer of `expo-image` on the web platform.
To support rendering svg assets by passing the URI to the `` src prop, it's necessary for the asset url to return the correct content type.

The solution follows the code described in [this issue](https://github.com/facebook/metro/issues/394), and was tested to work and allow using SVG assets on web in an expected way.

Fixes #394.

## Test plan

SVG returned as text without the patch in this PR:
image

SVG rendered correctly on asset URL after applying the patch:
image

Expected content type:
image


aleqsio:
@motiz88 Sorry for reopening the PR, github codespaces UI removed the branch instead of updating it with recent changes 😕 
This is same as https://github.com/facebook/metro/pull/951, but I've removed the redundant line and added requested tests.

motiz88 (changes requested):

motiz88 (approved):

PR #961, opened 2023-03-30
Fix missing content-length header when fetching assets

byCedric:
## Summary

This PR relates to PR #953 and adds the `Content-Length` property when loading assets.

While this isn't mandatory for the apps, it's extremely useful for some debugging possibilities. At Expo, we are working on debugging improvements through Chrome DevTools, and one of these improvements is adding support for the Network inspector.

On the native side, we are limited to what we can infer automatically. One of these limitations is intercepting response bodies. We can't intercept these response bodies if it's larger than 1mb, and because of that, we rely on the `Content-Length` header to communicate the size of the response.

## Test plan

image

I can provide a repository to try this debugger, if it's necessary. You can also test this change by manually loading an existing asset (e.g. `http://localhost/assets/icon.png?platform=android`) and check the response headers in the same debugger.

### DIFF
--- a/packages/metro/src/Server.js
+++ b/packages/metro/src/Server.js
@@ -401,6 +401,8 @@ class Server {
       return data.slice(dataStart, dataEnd + 1);
     }
 
+    res.setHeader('Content-Length', String(Buffer.byteLength(data)));
+
     return data;
   }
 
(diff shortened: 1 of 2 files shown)


huntie (approved):
Looks good to me, thanks!

PR #967, opened 2023-04-17
feat: add support for preserving comments in code

tido64:
## Summary

Add support for preserving comments in code. Useful for preserving `/*#__PURE__*/` annotations that aid tree shaking.

## Test plan

Test case here: https://github.com/microsoft/rnx-kit/issues/2365

|        |    Size |
| :----- | ------: |
| Before | 1567764 |
| After  | 1258413 |
| Diff   | -309351 |


robhogan:
Thanks! This seems pretty well defined - I wonder whether we could just go straight to stable with this. WDYT @motiz88?

packages/metro-transform-worker/src/__tests__/index-test.js
```diff
@@ -583,3 +584,23 @@ it('counts all line endings correctly', async () => {
...
   );
 });
+
+it('allows preservation of comments', async () => {
+  const result = await Transformer.transform(
+    {...baseConfig, unstable_preserveComments: true},
+    '/root',
+    'local/file.js',
+    Buffer.from('/*#__PURE__*/arbitrary(code);', 'utf8'),
+    // $FlowFixMe[prop-missing] Added when annotating Transformer.
```
motiz88, line 594: Can we avoid introducing new `$FlowFixMe`s?

motiz88:
> This seems pretty well defined - I wonder whether we could just go straight to stable with this.

@robhogan: Makes sense to me. I guess this is something all transformers (even non-Babel-based ones) would be able to implement, so no concerns there. The only real gap IMO is that we should document this in https://facebook.github.io/metro/docs/configuration/#transformer-options if stable. To avoid confusion, I wouldn't explicitly mention tree-shaking, and instead just briefly describe what the option itself does.

packages/metro-transform-worker/src/__tests__/index-test.js
```diff
@@ -583,3 +584,23 @@ it('counts all line endings correctly', async () => {
...
+    {
+      dev: false,
+      minify: false,
```
motiz88, line 597: We probably need a test for `minify: true` to clarify how these options interact.

Also, 💡 maybe we should just preserve all comments when not minifying? Similar to Babel's defaults: https://babeljs.io/docs/options#shouldprintcomment

motiz88: (By which I mean, maybe we can avoid introducing this option altogether.)

robhogan: I think there's a case for moving this responsibility entirely to the minifier, and just setting `comments: true`. For a case like this when you need comments in the serializer, that shouldn't prevent subsequent minification.

tido64: I think minify and comments should be two separate things. I'm fine with `minify` affecting the default value of `preserveComments` though. @robhogan's suggestion seems to be the best option here. I can set `comments: true` in `metro-transform-worker`, and `comments: false` in the default minifier config to preserve the current behaviour. Let me know what you prefer.

motiz88: > I think minify and comments should be two separate things.

Can you elaborate? What's the use case?

My preference is for the following setup:
1. The transformer should always preserve comments (before minification) if it can, without us adding a new option to control this.
2. The minifier should default to stripping comments, but this can be changed using the already-existing [`minifierConfig`](https://facebook.github.io/metro/docs/configuration/#minifierconfig) option.

tido64: I think we are in agreement. I just misunderstood your first comment. 😄

tido64: Updated. Let me know if I've misunderstood anything.

packages/metro-config/src/defaults/index.js
```diff
@@ -111,6 +111,7 @@ const getDefaultValues = (projectRoot: ?string): ConfigT => ({
       },
       output: {
         ascii_only: true,
+        comments: false,
```
motiz88, line 114: 1. I don't see a need to specify this explicitly.
2. Our default minifier, Terser, has a [documented](https://terser.org/docs/api-reference#format-options) default of `"some"`, set [here](https://github.com/terser/terser/blob/e072c218498ff5bcacab268f8813d5480ba0d288/lib/cli.js#L102), so this overrides that and may be an unwanted change in behaviour. Per 1, let's just remove this explicit Metro default.

```suggestion
```

tido64: Previously, we had `comments: false` in Babel, which basically omits all comments, so no comments ever reach the minifier. To preserve that behaviour, we need to set `comments: false` here. If you still want to remove it, that's fine with me, but it will produce a different output than today.

motiz88: I see! Yeah, in that case we should probably land this with a behaviour-preserving default so it can go out as a patch, then align with terser's default in the next major. (Would you mind sending a separate PR with that?) 

The only thing I don't get is: if this is supposed to be behaviour-preserving, why did the buildGraph test snapshot change? Do we have a good test covering the actual default from Metro's config?

tido64: It changed because it's unminified, so the comments have yet to be stripped out.

I can send out a separate PR to remove this line.

tido64: Thanks for merging this PR! Here's the follow-up: https://github.com/facebook/metro/pull/972

PR #974, opened 2023-05-02
Support ';&' as an alternative to '?' in bundle URLs

robhogan:
Summary:
Recent versions of JavaScriptCore strip query strings from `Error.prototype.stack`, which breaks our use of query strings to carry bundle build parameters. See https://github.com/facebook/react-native/issues/36794 for context.

This allows Metro to accept an alternative format where we interpret the special reserved character sequence `;&`, the first time it appears in a URL *path*, as equivalent to `?`.

So that this does not break custom `rewriteRequestUrl` implementations, we (temporarily) pass a normalised URL (`;&` replaced with `?`), and then reverse that operation on the URL returned by `rewriteRequestUrl`. *[Simplifying the details here somewhat - see the implementation for specifics].*

Changelog:
```
**[Feature]**: Support alternative JavaScriptCore-safe URL format (`;&` as `?`).
```

Differential Revision: D45477505

### DIFF
--- a/packages/metro/src/Server.js
+++ b/packages/metro/src/Server.js
@@ -56,6 +56,7 @@ const ResourceNotFoundError = require('./IncrementalBundler/ResourceNotFoundErro
 const bundleToString = require('./lib/bundleToString');
 const formatBundlingError = require('./lib/formatBundlingError');
 const getGraphId = require('./lib/getGraphId');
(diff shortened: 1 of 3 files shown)


robhogan:
Superseded by https://github.com/facebook/metro/pull/989

PR #991, opened 2023-05-27
Allow client-side device identifiers in inspector proxy

### LINKED ISSUE #985: Improve inspector proxy device identifier

Feature

This is an idea I shared with @huntie during the App.js Conf, but something I would love other opinions on.

Metro currently registers each device using an incremental ID generator inside the inspector proxy, when using the inspector through CDP. This is a simple approach, but it has some downsides.

For example, if an app fully restarts while keeping the inspector open, the same device is registered under a different ID. This forces users to restart the inspector, or it won't be connected to the right websocket address.

Instead of using incremental IDs, we might be able to leverage a more "stable identifier", such as the app installation ID (or even a generated UUID on the device). With that, the URL of the websocket address shouldn't change much. If an app gracefully restarts, the websocket disconnection should free up the previous registered device ID.

With this, the inspector connection _should_ be a bit more stable. But I wonder what other people think.

byCedric:
## Summary

Fixes #985 

This allows devices connecting to the inspector proxy to provide a client-side unique string identifier. With this, we can pass along a unique identifier that doesn't change when reconnecting, e.g. after a hard crash.

It makes the overall debugging experience way more stable since you don't have to restart the debugger in between crashes or full restarts. Allowing users to keep the debugger open also lets the debugger "remember" certain things, such as set breakpoints.

> ⚠️ When the client doesn't specify this unique identifier, it still falls back to the incremental identifier (the `fallbackDeviceId` in this PR)

When a collision occurs, the old device's connection is closed. But, if both the device and app names are equal to the new device connection, the debugger connection is kept open.

## Next steps

If this is landed, we still need changes in React Native (the `&device=...` query param). I have no strong opinions on what identifier is used, but it should follow these rules:
- Should be unique per device or emulator/simulator (and thus, also per platform)
- Must not change when restarting the app
- Must be a string that's URL safe

On Android, I had good success using the [`Secure.ANDROID_ID`](https://developer.android.com/reference/android/provider/Settings.Secure#ANDROID_ID). It may change after a factory reset, which doesn't matter for our use case.

## Test plan

- Create a new project, and enable building from source on both Android & iOS
- **android**: edit [this file](https://github.com/facebook/react-native/blob/main/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/devsupport/DevServerHelper.java#L282-L289) in `node_modules/react-native` and add a hardcoded `&device=testingandroid` query param.
- **ios**: edit [this file](https://github.com/facebook/react-native/blob/main/packages/react-native/React/DevSupport/RCTInspectorDevServerHelper.mm#L43-L53) in `node_modules/react-naive` and add a hardcoded `&device=testingios` query param.
- Connect the debugger to the running app
- Force close the app, which should cause a "reconnect" warning in the debugger
- Open the app again, and press "reconnect" in the debugger
- _Due to the stable identifiers, the URL won't change and the above scenario should work fine_

Also test without these `&device=...` query param since that should work as it does now (with incremental identifiers).

### DIFF
--- a/packages/metro-inspector-proxy/src/Device.js
+++ b/packages/metro-inspector-proxy/src/Device.js
@@ -134,6 +134,10 @@ class Device {
     return this._name;
   }
 
+  getApp(): string {
+    return this._app;
+  }
+
   getPagesList(): Array<Page> {
     if (this._lastConnectedReactNativePage) {
       const reactNativeReloadablePage = {
@@ -210,6 +214,35 @@ class Device {
       debug('(Debugger) <- (Proxy)    (Device): ' + message);
       return sendFunc.call(socket, message);
     };
+  }
+
+  // Handles cleaning up a duplicate device connection, by client-side device ID.
+  // 1. Checks if the same device is attempting to reconnect for the same app.
+  // 2. If not, close both the device and debugger socket.
+  // 3. If the debugger connection can be reused, close the device socket only.
+  //
+  // This allows users to reload the app, either as result of a crash, or manually
+  // reloading, without having to restart the debugger.
+  handleDuplicateDeviceConnection(newDevice: Device) {
+    if (
+      this._app !== newDevice.getApp() ||
+      this._name !== newDevice.getName()
+    ) {
+      this._deviceSocket.close();
+      this._debuggerConnection?.socket.close();
+    }
+
+    const oldDebugger = this._debuggerConnection;
+    this._debuggerConnection = null;
+
+    if (oldDebugger) {
+      oldDebugger.socket.removeAllListeners();
+      this._deviceSocket.close();
+      newDevice.handleDebuggerConnection(
+        oldDebugger.socket,
+        oldDebugger.pageId,
+      );
+    }
   }
 
   // Handles messages received from device:
--- a/packages/metro-inspector-proxy/src/Device.js
+++ b/packages/metro-inspector-proxy/src/Device.js
@@ -216,13 +216,15 @@ class Device {
     };
   }
 
-  // Handles cleaning up a duplicate device connection, by client-side device ID.
-  // 1. Checks if the same device is attempting to reconnect for the same app.
-  // 2. If not, close both the device and debugger socket.
-  // 3. If the debugger connection can be reused, close the device socket only.
-  //
-  // This allows users to reload the app, either as result of a crash, or manually
-  // reloading, without having to restart the debugger.
+  /**
+   * Handles cleaning up a duplicate device connection, by client-side device ID.
+   * 1. Checks if the same device is attempting to reconnect for the same app.
+   * 2. If not, close both the device and debugger socket.
+   * 3. If the debugger connection can be reused, close the device socket only.
+   *
+   * This allows users to reload the app, either as result of a crash, or manually
+   * reloading, without having to restart the debugger.
+   */
   handleDuplicateDeviceConnection(newDevice: Device) {
     if (
       this._app !== newDevice.getApp() ||
--- a/packages/metro-inspector-proxy/src/Device.js
+++ b/packages/metro-inspector-proxy/src/Device.js
@@ -55,7 +55,7 @@ const REACT_NATIVE_RELOADABLE_PAGE_ID = '-1';
  */
 class Device {
   // ID of the device.
-  _id: number;
+  _id: string;
 
   // Name of the device.
   _name: string;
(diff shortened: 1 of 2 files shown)


packages/metro-inspector-proxy/src/InspectorProxy.js
```diff
@@ -162,16 +162,30 @@ class InspectorProxy {
...
+
         const query = url.parse(req.url || '', true).query || {};
+        const deviceId = query.device || fallbackDeviceId;
         const deviceName = query.name || 'Unknown';
         const appName = query.app || 'Unknown';
-        const deviceId = this._deviceCounter++;
+
+        const oldDevice = this._devices.get(deviceId);
+        if (oldDevice) {
+          // Keep the debugger connection alive when disconnecting the device, if possible
```
voideanvalue, line 174: Does the `new Device` created below somehow inherit `_debuggerConnection` from `oldDevice`?

Also, `_debuggerConnection` is a private variable (by naming convention) and ideally shouldn't be accessed outside of the `Device` class

byCedric: > Does the `new Device` created below somehow inherit `_debuggerConnection` from `oldDevice`?

It does, but we might need to add some logic when swapping out the `Device` instances. Basically, in my tests I found that the open debugger automatically triggered the `Device.handleDebuggerConnection` method (background reconnection). This sets the debugger connection to the new `Device` instance.

I think an ideal scenario is to just invoke that method directly during the initialization.

Originally, I had to unset the debugger connection in order to avoid the [device's socket close event handler](https://github.com/facebook/metro/blob/main/packages/metro-inspector-proxy/src/Device.js#L123-L127).

byCedric: > Also, `_debuggerConnection` is a private variable (by naming convention) and ideally shouldn't be accessed outside of the `Device` class

One thing we can do is move the shutdown logic into a handler on the device. With that, we don't need to access `_debuggerConnection`, `_name`, and `_app`. Does that make sense to you? @voideanvalue

huntie: Yup makes sense, I think, feel free to modify the `Device` class (as it's private within this package) to accomplish the new logic needed as well-named methods.

packages/metro-inspector-proxy/src/Device.js
```diff
@@ -212,6 +216,35 @@ class Device {
     };
   }
 
+  // Handles cleaning up a duplicate device connection, by client-side device ID.
+  // 1. Checks if the same device is attempting to reconnect for the same app.
+  // 2. If not, close both the device and debugger socket.
+  // 3. If the debugger connection can be reused, close the device socket only.
+  //
+  // This allows users to reload the app, either as result of a crash, or manually
+  // reloading, without having to restart the debugger.
```
huntie, line 225: nit: Use doc comment.

packages/metro-inspector-proxy/src/InspectorProxy.js
```diff
@@ -162,16 +162,31 @@ class InspectorProxy {
...
 
-        debug(`Got new connection: device=${deviceName}, app=${appName}`);
+        if (oldDevice) {
+          oldDevice.handleDuplicateDeviceConnection(newDevice);
+        }
+
+        this._devices.set(deviceId, newDevice);
+
+        debug(
+          `Got new connection: device=${deviceName}, app=${appName}, id=${deviceId}`,
```
huntie, line 188:
```suggestion
          `Got new connection: name=${deviceName}, app=${appName}, device=${deviceId}`,
```

nit: Let's align this log with the query param names.

huntie (approved):

huntie:
Thanks Cedric! LGTM with small nits 🙂. Will import and it'll receive a final review.

The criteria for a device/session ID sound good — a stable identifier that is unique per device and app. I'm happy to handle the next steps on the React Native side.

packages/metro-inspector-proxy/src/Device.js
byCedric: I used what was used on the other event handlers, but will change this to docblock :)

byCedric:
> If you need, I'm happy to handle the next steps on the React Native side.

That would be great 🙏

motiz88:
> The criteria for a device/session ID sound good — a stable identifier that is unique per device and app.

To nitpick slightly - a device ID and session ID don't sound like the same thing. It's important to consider what happens if we have multiple devices, multiple apps on each device, and possibly even **multiple instances of React Native** in each app (and in the future, multiple JavaScript contexts e.g. Web Workers in each instance of React Native) - all of which could reasonably be connected to one instance of Metro.

This PR looks good, but eventually we should figure out all the different entities and their lifetimes, and flesh out what types of IDs (and persistence/stability) we need at each level.

PR #1009, opened 2023-06-19
expose TerminalReporter for Reporter implementations

colinta:
We are experimenting with creating our own `Reporter` that can intercept `console.log` messages and turn those into a filterable stream that gets subscribed via a WebSocket. IE

```ts
// this is ok I guess:
import { TerminalReporter } from 'metro/src/lib/TerminalReporter';
// this is preferred, "official" export, hence this PR
import { TerminalReporter } from 'metro';

export default class Reporter extends TerminalReporter {
  update(event) {
    switch (event.type) {
      case 'client_log':
        // send 
        break;
      default:
        TerminalReporter.prototype.update(event);
    }
  }
}
```

## Summary

To make this easy, I just want to expose `TerminalReporter` from the module, rather than reach into the file.

## Test plan

Not much needed here, AFAICT.


motiz88:
Thanks @colinta! Would you mind updating https://github.com/facebook/metro/blob/main/packages/metro/types/index.d.ts accordingly?

colinta:
> Would you mind updating

Happy to!

robhogan (approved):

Issue #1026, opened 2023-07-07
ESM project support broken in 0.76/RN 0.72


oblador:
I'm working in a "isomorphic" project with both react-dom and react-native, where the web part is using vite that requires ESM. To make that setup work with RN/Metro that is still heavily reliant on CJS I simply renamed the metro.config.js to metro.config.cjs to force it to resolve using CJS. 

With the latest version of RN (0.72.1) this setup yields `No metro config found`, and if I rename it to `.js` and use esm it throws errors originating from metro for using requires in an esm context. 

My guess this is originates in an update of cosmiconfig or similar. 

Another solution would be to properly support ESM/TS in the config as suggested in https://github.com/facebook/metro/issues/916

robhogan:
Hi @oblador - thanks for this, @huntie is already on the case adding support for `metro.config.cjs`, I'd like to dig in a bit as to whether something's broken.

> but with 0.71.x it works fine.

This is curious. I'm not aware that `metro.config.cjs` has *ever* been supported, but if something's regressed we'll need to pick a fix back for RN72. I wonder if there might've been a build step in your project outputting a side-by-side `.js`?

And just to confirm so I can reproduce, when you say "use ESM", that's by setting `"type": "module"` in your project's `package.json`, or more than that?

> Even though RN setup says Node 16+, metro requires 18+ 

Node 16 should be fine for Metro 0.76.x / RN 0.72.x. We only bumped the requirement to 18 in Metro 0.77, for RN 0.73.

oblador:
> I wonder if there might've been a build step in your project outputting a side-by-side .js?

Nope there isn't, just single-step babel/metro.

> And just to confirm so I can reproduce, when you say "use ESM", that's by setting "type": "module" in your project's package.json, or more than that?

Exactly!

> Node 16 should be fine for Metro 0.76.x / RN 0.72.x. We only bumped the requirement to 18 in Metro 0.77, for RN 0.73.

Looked into this and tracked it down to be caused by the `@types/metro-config` package that has a `metro-config "*"` dependency set up, which resolves to latest (0.77) (using yarn 1.22.17). This package isn't really needed tbf and IMHO the types should be shipped with the package itself and not separate, but I try to stick to the official template setup unless I have good reasons not to, makes upgrading a lot easier.

oblador:
[...]

Issue #916, opened 2023-01-18
Support ES Module / Typescript config


acalvino4:
## Feature request: support typescript / esm config

**What is the current behavior?**

Metro does not look for a `metro.config.ts`, and if it did, I have no idea whether it would be configurable as a module.

This seems necessary sooner or later, as the ecosystem is moving to modules, and majority of people are using typescript (RN just made it the default for new projects).

**What is the desired behavior?**

Metro uses `metro.config.ts` if it exists, and the docs include an example of what this looks like.

karlhorky:
Update:

In the other issue @robhogan mentioned that ESM configs will need to wait until Node.js v20 is used for the `react-native` projects:

https://github.com/react-native-community/cli/issues/2167#issuecomment-1994523341

stefan-schweiger:
Now that we have ESM support for react-native.config files how about metro.config?

CodeWithShreyans:
Is this being reconsidered now?

Sid110307:
Any update?

sam-warren-finnair2:
I ran into this problem today as well. All files apart from this are now in ESM format, so this is quite frustrating having to find a workaround for.

rdnsan:
Any update on this?

robhogan:
Metro now supports ESM and TS (requires Node 22+) configs, I still need to update the docs.

https://github.com/facebook/metro/releases/tag/v0.83.2

Master-Antonio:
> Metro now supports ESM and TS (requires Node 22+) configs, I still need to update the docs.
> 
> https://github.com/facebook/metro/releases/tag/v0.83.2

How to enable? I have 0.83.3 but still this error

<img width="414" height="407" alt="Image" src="https://github.com/user-attachments/assets/b87a21e3-1b7e-419b-ad22-1c6902f6df23" />

```

const { getDefaultConfig, mergeConfig } = require('@react-native/metro-config');
/**
 * Metro configuration
 * https://reactnative.dev/docs/metro
 *
 * @type {import('@react-native/metro-config').MetroConfig}
 */
const config = {};
module.exports = mergeConfig(getDefaultConfig(__dirname), config);
```

robhogan:
@Master-Antonio we’re talking about the config file itself being ESM or TypeScript (eg `metro.config.mts`). I’m not sure what you’re trying to do there, but please open a separate issue.

zowebs:
[...]

PR #1027, opened 2023-07-07
Widen config search paths to include metro.config.cjs

huntie:
Summary:
Resolves https://github.com/facebook/metro/issues/1026.

Changelog: **[Feature]** Widen config search paths to include `metro.config.cjs`

Differential Revision: D47290079

### DIFF
--- a/packages/metro-config/src/loadConfig.js
+++ b/packages/metro-config/src/loadConfig.js
@@ -46,8 +46,12 @@ function overrideArgument<T>(arg: Array<T> | T): T {
 }
 
 const explorer = cosmiconfig('metro', {
-  searchPlaces: ['metro.config.js', 'metro.config.json', 'package.json'],
-
+  searchPlaces: [
+    'metro.config.js',
+    'metro.config.cjs',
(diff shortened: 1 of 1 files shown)


packages/metro-config/src/loadConfig.js
```diff
@@ -46,8 +46,12 @@ function overrideArgument<T>(arg: Array<T> | T): T {
 }
 
 const explorer = cosmiconfig('metro', {
-  searchPlaces: ['metro.config.js', 'metro.config.json', 'package.json'],
-
+  searchPlaces: [
+    'metro.config.js',
+    'metro.config.cjs',
```
oblador, line 51: I think you might have to add `cjs` to `loaders` section too

oblador:
Thanks for a quick fix!

> resolves #916

Are you sure this is true? Doesn't seem to address TS in the config

packages/metro-config/src/loadConfig.js
huntie: Great spot! That got caught by our internal CI too :)

PR #1036, opened 2023-07-15
Use shorter error message for file resolution issues.

EvanBacon:
## Summary

- Part of the resolution error is very verbose and somewhat hard to make sense of.
- This PR condenses the matchable segments to reduce duplicate contents.
  - Add optional `(/index)?` segment to prevent needing a second line.
  - Modify header `none of these files exist:` to account for the single line `no file matched the pattern`. I find this line more helpful as it implies the issue may be due to the pattern and not to the lack of file existence. Generally when users are reporting this issue to Expo CLI, it's because they need to modify `srcExts`.
  - Reduce duplicate references in extensions `(.native|.native.js|.js|.native.json|.json)` -> `(.native)?.(native|js|json)`. This clearly has an issue where `.native.native` would technically match the listed pattern––open to suggestions on how to format this differently. However, `.native` seems like it shouldn't be present regardless. Worth noting that I hadn't noticed the extraneous `.native` before because the length of the error message.
- [reference to discord proposal](https://discord.com/channels/514829729862516747/514832110595604510/1128846939060961411).

Screenshot 2023-07-15 at 6 32 54 PM

## Test plan

- Unit tests cover this well.

### DIFF
--- a/packages/metro-resolver/src/errors/formatFileCandidates.js
+++ b/packages/metro-resolver/src/errors/formatFileCandidates.js
@@ -15,6 +15,59 @@ import path from 'path';
 
 import type {FileCandidates} from '../types';
 
+function groupExtensions(exts: $ReadOnlyArray<string>): string[][] {
+  // Reverse the extensions and split into parts
+  const extParts = exts.map(ext => ext.split('.').reverse());
+
+  // Find the maximum depth of extension parts
+  const maxDepth = Math.max(...extParts.map(parts => parts.length));
+
+  // Initialize groups based on the max depth
+  const groups = Array.from({length: maxDepth}, () => new Set<string>());
+
+  extParts.forEach(parts => {
+    parts.forEach((part, i) => {
+      // Add parts to the corresponding group based on their depth
+      groups[i].add(part);
+    });
+  });
+
+  // Cycle groups and remove duplicates that appear forwards
+  groups.forEach((group, index) => {
+    // Remove duplicates that appear forwards
+    // NOTE: This doesn't support extensions like `.native.native.js`
+    groups.forEach((otherGroup, otherIndex) => {
+      if (index < otherIndex) {
+        otherGroup.forEach(part => group.delete(part));
+      }
+    });
+  });
+
+  // Convert sets back to arrays and reverse groups to correct order
+  return groups.map(group => Array.from(group)).reverse();
+}
+
+function createMatcherPatternForExtensions(
+  exts: $ReadOnlyArray<string>,
+): string {
+  let formatted = '';
+
+  if (exts.length) {
+    // Apply grouping function
+    const groups = groupExtensions(exts);
+
+    formatted += groups
+      .map((group, index) => {
+        return index < groups.length - 1
+          ? `(${group.map(ext => `.${ext}`).join('|')})?`
+          : `.(${group.join('|')})`;
+      })
+      .join('');
+  }
+
+  return formatted;
+}
+
 function formatFileCandidates(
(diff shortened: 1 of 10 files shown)


motiz88:
This looks great! Re: `.native.native`, that seems weird - is it just an artifact of the current tests or does it show up in real-world usage too?

motiz88:
Oh, found the `.native` bug: https://github.com/facebook/metro/blob/c39b51b2df74479abd00b30866e06c35d7a23385/packages/metro-resolver/src/resolve.js#L463-L464

We need to check that `sourceExt` isn't empty.

packages/metro-resolver/src/errors/formatFileCandidates.js
```diff
@@ -11,15 +11,53 @@
...
   }
   let formatted = candidates.filePathPrefix;
+
+  if (allowIndex) {
+    formatted += `(${path.sep}index)?`;
+  }
+
   if (candidates.candidateExts.length) {
-    formatted += '(' + candidates.candidateExts.filter(Boolean).join('|') + ')';
+    const exts = candidates.candidateExts.filter(Boolean);
```
motiz88, line 32: How does this logic handle entries in `sourceExts` that contain a `.`? Not sure if there's an existing test for this, but internally at Meta we have a config with `.fb.js` as a possible extension. Would this logic split it into `fb` and `js`, suggesting that other compound extensions like `.fb.ts` are included (or given special precedence) when they're not? Or would it correctly preserve `.fb.js` as a unit? (Caveat: I'm too jetlagged to try to follow the code right now.)

packages/metro-resolver/src/errors/formatFileCandidates.js
```diff
@@ -11,15 +11,53 @@
...
+    // exts are formatted as `.ios.js, .native.js, .js, .native.json, .json`
+    // we want to split them into groups delimited by the period, then format
+    // like `(.native|.ios).(js|json)`
+
+    // split into groups, filter, reverse to ensure we get the most specific
+    const groups = exts.map(ext => ext.split('.').filter(Boolean).reverse());
+
+    const splits = groups
+      .reduce((acc, group) => {
```
motiz88, line 41: Flow wants an annotation on `acc` here.

packages/metro-resolver/src/errors/formatFileCandidates.js
EvanBacon: This potentially introduces a three-level deep extension (`.native.fb.js`) which could be formatted like:

```
* missing(/index)?(.native)?(.fb)?.(js|json)
```

- `missing.native.fb.js`
- `missing.fb.js`
- `missing.native.js`
- `missing.js`
- `missing/index.native.fb.js`
- `missing/index.fb.js`
- `missing/index.native.js`
- `missing/index.js`


Does this look right to you?

packages/metro-resolver/src/resolve.js
```diff
@@ -460,7 +460,10 @@ function resolveSourceFileForAllExts(
     }
   }
   if (context.preferNativePlatform) {
```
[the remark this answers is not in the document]
EvanBacon: https://github.com/facebook/metro/pull/1045

rshest:
@EvanBacon Note that this still can't be merged, unfortunately, due to both merge conflicts with the mainline, and also some outstanding follow up still required (as requested by Moti).

EvanBacon:
@rshest will follow up when I have more time. Wasn't accounting for the facebook-specific logic when I opened the PR.

PR #1061, opened 2023-08-17
pass custom customTransformOptions and customResolverOptions to bundle_build_started log event

EvanBacon:
## Summary

We use `customTransformOptions` in Expo CLI to inform Metro that we're bundling for a Node.js environment (in addition to the `platform=web`), but the logger has no way of indicating to the user that a bundle is Server-specific. This change would enable us to refine the bundling message based on the additional criteria. We can do this if we skip over the server (exports/builds) but it'd be good to have parity. [Example usage](https://github.com/expo/expo/blob/e4eba8a7c9d8134c10578a47d39ddf5294aca9e4/packages/%40expo/cli/src/start/server/metro/MetroTerminalReporter.ts#L227), [corresponding output](https://github.com/expo/expo/blob/e4eba8a7c9d8134c10578a47d39ddf5294aca9e4/packages/%40expo/cli/src/start/server/metro/__tests__/MetroTerminalReporter-test.ts#L61).


motiz88:
Looks good to me, but there's a Flow error in `packages/metro/src/commands/build.js` (aren't we running Flow in CircleCI? Maybe not on PRs?)

robhogan:
Lint error now is due to an internal callsite that needs updating - I'll take care of that.

PR #1071, opened 2023-08-31
add more info to the `getPolyfills` function

EvanBacon:
## Summary

We use `getPolyfills` in Expo CLI to supply custom global setup for Node.js environments. Currently, these polyfills are applied to bundles for both client and server web environments. This PR would supply additional info that we can use to pick polyfills.


robhogan:
Sorry about the delay with this one. At first glance it seems a *bit* arbitrary to add `customTransformOptions` to this (and not other config), but it's a valid use case and fits with future extensions, so I think we can get this in.

robhogan (approved):

PR #1082, opened 2023-09-13
Prevent Babel registration from transforming plugins with arbitrary config

robhogan:
Summary:
Using `require` after `const registerFn = require('babel/register')` but before `registerFn(config)` causes Babel to transform required code with the default configuration (ie, using a nearby `babel.config.js`, if available).

This was causing the Babel plugins loaded by `metro-babel-register` to be (unnecessarily) transformed according to `babel.config.js`, which actually fails if the plugins/presets referenced in `babel.config.js` themselves require transformation.

This ensures no code is loaded in between registering Babel as a side effect of requiring Babel register, and replacing that hook with something explicitly configured.

## React Native
Changelog: [Internal]

## Metro
```
* **[Fix]:** `metro-babel-register` prevent arbitrary transformation of Babel plugins during registration setup
```

Differential Revision: D49238671

### DIFF
--- a/packages/metro-babel-register/src/babel-register.js
+++ b/packages/metro-babel-register/src/babel-register.js
@@ -22,7 +22,18 @@ const path = require('path');
 let _only /*: $ReadOnlyArray<RegExp | string> */ = [];
 
 function register(onlyList /*: $ReadOnlyArray<RegExp | string> */) {
-  require('@babel/register')({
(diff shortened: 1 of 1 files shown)


robhogan:
Merged in https://github.com/facebook/metro/commit/e8f468d5e4f1b88f1b99c66656cec004d4079831

PR #1085, opened 2023-09-15
bump react-refresh to latest

EvanBacon:
## Summary

React Refresh v4 doesn't work with statically rendered React websites. We've been encouraging Expo Router users to resolve v14 for a couple months now. I haven't observed any issues with the latest version.

I'm opening a PR in React Native too. https://github.com/facebook/react-native/pull/39486

## Test plan

1. In a React Native project's `package.json`:
```json
 "resolutions": {
    "react-refresh": "~0.14.0"
  },
```
2. Start the server with a clear Metro cache.
3. Changes should update while preserving React state.

### DIFF
--- a/packages/metro-runtime/package.json
+++ b/packages/metro-runtime/package.json
@@ -14,7 +14,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "react-refresh": "^0.4.0"
+    "react-refresh": "^0.14.0"
   },
   "devDependencies": {
     "@babel/core": "^7.20.0",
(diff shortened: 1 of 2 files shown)


motiz88:
@EvanBacon Can you please delete [metro-runtime/src/modules/fast-refresh.js](https://github.com/facebook/metro/blob/main/packages/metro-runtime/src/modules/fast-refresh.js) and demote `react-refresh` to a devDependency? (As it's used purely for types in tests)

Now that the RN Babel preset is colocated with the RN runtime in the `react-native` repo, your other PR https://github.com/facebook/react-native/pull/39486 should be sufficient to actually upgrade `react-refresh` in RN.

robhogan:
I'm going to import this as-is and stack Moti's suggestion on top of it (removing the `fast-refresh` module is semver breaking, so I'll give it a separate commit).

PR #1092, opened 2023-10-02
feat: update docusaurus to 2.4.3

OzakIOne:
## Summary

Upgrade docusaurus to 2.4.3

## Test plan

CI Argos https://github.com/OzakIOne/metro-argos/pull/1


slorber:
Thanks, I see.

We'll try to temporarily remove the usage of that package and see if we can make the site run. If it does, we can send a draft PR to upgrade to v3.

PR #1130, opened 2023-11-02
Add function names to module factories in DEV

gaearon:
This adds path-based function names to the generated module factories in development.

```js
__d(function src_some_folder_path_MyComponent_js__module_factory__() {
  // ...
})
```

They're only emitted when `minify` is `false` and `dev` is `true`.

## Motivation

Profiling module initialization is a common need. The current set of tooling React Native provides for it is arguably abysmal.

The primary tool [suggested](https://reactnative.dev/docs/profiling#profiling-android-ui-performance-with-systrace) by React Native documentation is Systrace which literally [does not exist anymore](https://stackoverflow.com/a/74005757). "Profiling with Hermes" suggests using [sampling profiler](https://reactnative.dev/docs/profile-hermes) which seems to barely work — I wasn't able to capture any information with it.

If you try to use Hermes Chrome DevTools integration and press "Record" in the Performance tab, it hangs.

I almost gave up at that point, but @brentvatne kindly suggested that there is a workaround — I need to Shift+Cmd+P to get a different "JS Profiler" Chrome DevTools tab to show up, and that one actually works.

Even that approach [is a little bit broken](https://github.com/facebook/hermes/discussions/1181) but it's the closest to what I want.

Unfortunately, even that tool does not let me see which modules I'm spending time initializing:

Screenshot 2023-11-02 at 17 15 05

As you can see, module factories have blank function names.

This is not the case with web-based tooling. For example, webpack emits module factories as a dictionary like `{ "./src/some/module.js": function() {}, ... }` so the stacks are useful. Since we don't want to change the format entirely, and configuring `fn.name` dynamically at runtime does not make it show up in the trace, I opted for a minimal possible change — to actually emit the module names in development.

As a result, I can see where I'm spending time during module initialization:

Screenshot 2023-11-02 at 17 24 47

### DIFF
--- a/packages/metro-transform-worker/src/index.js
+++ b/packages/metro-transform-worker/src/index.js
@@ -346,6 +346,11 @@ async function transformJS(
   let dependencies;
   let wrappedAst;
 
+  const minify =
+    options.minify &&
+    options.unstable_transformProfile !== 'hermes-canary' &&
+    options.unstable_transformProfile !== 'hermes-stable';
+
   // If the module to transform is a script (meaning that is not part of the
   // dependency graph and it code will just be prepended to the bundle modules),
   // we need to wrap it differently than a commonJS module (also, scripts do
@@ -382,20 +387,20 @@ async function transformJS(
     if (config.unstable_disableModuleWrapping === true) {
       wrappedAst = ast;
     } else {
+      let moduleFactoryName;
+      if (options.dev && !minify) {
+        moduleFactoryName = file.filename;
+      }
       ({ast: wrappedAst} = JsFileWrapping.wrapModule(
         ast,
         importDefault,
         importAll,
         dependencyMapName,
         config.globalPrefix,
+        moduleFactoryName,
       ));
     }
   }
-
-  const minify =
-    options.minify &&
-    options.unstable_transformProfile !== 'hermes-canary' &&
-    options.unstable_transformProfile !== 'hermes-stable';
 
   const reserved = [];
   if (config.unstable_dependencyMapReservedName != null) {
--- a/packages/metro-transform-worker/src/__tests__/__snapshots__/index-test.js.snap
+++ b/packages/metro-transform-worker/src/__tests__/__snapshots__/index-test.js.snap
@@ -240,7 +240,7 @@ Object {
 `;
 
 exports[`transforms an es module with asyncToGenerator 1`] = `
-"__d(function (global, _$$_REQUIRE, _$$_IMPORT_DEFAULT, _$$_IMPORT_ALL, module, exports, _dependencyMap) {
+"__d(function local_file_js__module_factory__(global, _$$_REQUIRE, _$$_IMPORT_DEFAULT, _$$_IMPORT_ALL, module, exports, _dependencyMap) {
   var _interopRequireDefault = _$$_REQUIRE(_dependencyMap[0], \\"@babel/runtime/helpers/interopRequireDefault\\");
   Object.defineProperty(exports, \\"__esModule\\", {
     value: true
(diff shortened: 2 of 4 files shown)


packages/metro-transform-worker/src/index.js
```diff
@@ -382,21 +387,21 @@ async function transformJS(
     if (config.unstable_disableModuleWrapping === true) {
       wrappedAst = ast;
     } else {
+      let moduleFactoryName;
+      if (options.dev && !minify) {
+        moduleFactoryName = file.filename;
```
motiz88, line 392: My main concern here (other than the obvious: we should improve the profiling & debugging tools so that this isn't needed 😅) is around the portability and cache consistency of using this `filename` property.

I'm not sure off the top of my head whether this is (1) an absolute or relative path, (2) a platform-dependent or normalised path. Could you confirm?

* If the path is absolute, then transformer cache artifacts will be sensitive to the project's root directory, which is incorrect. (Cache artifacts should be compatible across machines, including e.g. checkouts at different base locations)
* If the path is absolute *and* platform-dependent, then it would begin with `C:\foo` on Windows but `/foo` on POSIX, which again is non-portable.

gaearon: This is a relative path in my testing — see the screenshot (e.g. `node_modules/blabla/...`).

I'm not sure if it's platform-dependent or not but we replace all non-valid JS ident characters with `_` so a path with `/` and `\` would result in the same identifier.

gaearon: Note that we already emit a path into the DEV artifacts [here](https://github.com/facebook/metro/blob/bbdd7d7c5e6e0feb50a9967ffae1f723c1d7c4e8/packages/metro/src/DeltaBundler/Serializers/helpers/js.js#L101) (the module's "verbose name") but it happens at a different stage so I'm not entirely sure if they're constructed in the same way or come from different sources.

robhogan:
Urgh - this caused native crashes across multiple apps in debug builds post-land, still connecting the dots but there are runtime mem allocation failures - first few stack frames, FYI:

```
SIGABRT:libxplat_hermes_API_HermesAPIAndroid.so:./xplat/hermes/external/llvh/include/llvh/Support/MemAlloc.h:llvh::SmallVectorBase::grow_pod(void*, unsigned int, unsigned int)

hermes::irgen::ESTreeIRGen::serializeScope(hermes::ScopeDesc*, bool)
hermes::irgen::ESTreeIRGen::setupLazyScope(hermes::ESTree::FunctionLikeNode*, hermes::Function*, hermes::ESTree::BlockStatementNode*)
hermes::irgen::ESTreeIRGen::genES5Function(hermes::Identifier, hermes::Variable*, hermes::ESTree::FunctionLikeNode*, bool)
```

gaearon:
Is there any chance that some (relative) paths are still incredibly long, or that Hermes has a limit on function names that isn't present in other JS engines? I was thinking of adding `.slice(0, 50)` or something.

robhogan:
> Is there any chance that some (relative) paths are still incredibly long, or that Hermes has a limit on function names that isn't present in other JS engines? I was thinking of adding `.slice(0, 50)` or something.

String length is a factor, but probably not the determining one. We saw total RAM on Facebook Android (dev builds) go from 1.6GB to 3.8GB with this change, so way more than just the sum of those strings, even though the paths are pretty long.

@tmikov has been working on it and identified a quadratic growth situation in where scopes were repeatedly re-serialised during lazy evaluation, but the fix for that didn't seem to be enough on its own.

Worst case, we could enable this behind a `serializer` config flag in Metro and document the memory use, but the Hermes folks were actively looking into it last I heard so they may want a bit more time.

tmikov:
There is no limit, Hermes is just not optimized for running from source - this problem doesn't exist when running from bytecode. Running from source with lazy compilation is, in a sense, contrary to Hermes's mission of being an optimizing AOT compiler. We basically bolted lazy compilation on top of the AOT compiler and had to use a lot of hacks, because it simply wasn't designed for that. It is a miracle it has worked as well as it has...

We will support this properly in Static Hermes (at tremendous complexity cost), but for now we are trying to figure out exactly what is consuming the memory in "regular" Hermes. There is something consuming quadratic memory.

PR #1134, opened 2023-11-07
feat: pass dev option to resolver

EvanBacon:
## Summary

Have first-class support for resolving based on if the process is bundling for `dev` or not. Ref https://discord.com/channels/514829729862516747/514832110595604510/1171554890456256522

This feature is essentially a built-in version of just passing `dev` as a custom resolver option.

## Test plan

Updated the tests, defaulting to `dev: true`.


motiz88:
Can you update the resolver docs to reflect this?

PR #1136, opened 2023-11-10
Use `resolvePackage` instead of `resolveModulePath` in `resolveHasteName`

### LINKED ISSUE #1128: Package Exports not working
It could be a BUG or maybe I just need some help with the configuration.

- package A has defined `exports` in the `package.json`
- The native app has package A as a dependency
- The metro config has the `unstable_enablePackageExports` enabled
- When we start the app `yarn start` or `npm start` it explodes with an error:
```
Error: While resolving module `package A`, the Haste package `package A` was found. However the module `native` could not be found within the package.
```

- You can check out this repro and try it out => https://github.com/Bekaxp/native-exports-debug
- Also here is a video if it helps: https://github.com/facebook/metro/assets/13329079/3ef10456-f6e5-45aa-8d0a-f16209e5bccd

I would expect that the packages are resolved correctly and the app starts without issues.

My metro config looks like this:
```

module.exports = (() => {

  const { resolver } = config;

  config.resolver = {
    ...resolver,
    unstable_enableSymlinks: true,
    unstable_enablePackageExports: true,
    unstable_conditionNames: ["import", "require"],
  };

  config.watchFolders = [".."];

  return config;
})();
```

I am running it on a MacOS Sonoma 14.0 (Apple M1 Pro)

Thank you very much for any tips and tricks to make it work :)

kraenhansen:
## Summary

This fixes #1128 by calling the `resolvePackage` instead of `resolveModulePath` in `resolveHasteName`.
Only `resolvePackage` has the code to resolve package "exports" and it calls `resolveModulePath` as a fallback.

## Test plan

I've added a failing test which passed as the fix got implemented.

### DIFF
--- a/packages/metro-resolver/src/__tests__/package-exports-test.js
+++ b/packages/metro-resolver/src/__tests__/package-exports-test.js
@@ -455,6 +455,24 @@ describe('with package exports resolution enabled', () => {
         expect(logWarning).not.toHaveBeenCalled();
       });
     });
+
+    describe('haste package', () => {
+      test('should resolve subpath in "exports"', () => {
+        const context = {
+          ...baseContext,
+          resolveHastePackage(name: string) {
+            if (name === 'test-pkg') {
+              return '/root/node_modules/test-pkg/package.json';
+            }
+            return null;
+          },
+        };
+        expect(Resolver.resolve(context, 'test-pkg/foo.js', null)).toEqual({
+          type: 'sourceFile',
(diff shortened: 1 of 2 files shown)


robhogan (approved):

kraenhansen:
I'm wondering if we need to update [the documentation](https://github.com/facebook/metro/blob/main/docs/Resolution.md#resolve_haste) based on this change 🤔

PR #1155, opened 2023-12-14
Restructure `Graph` - separate traversal from state mutation

robhogan:
This is a significant restructuring of the internal operation of `Graph`. Currently, we asynchronously traverse, and resolve, from a set of entry/modified paths, updating the graph's state as we go.

This change performs the same work in two distinct phases:
 1. Draft: Asynchronously traverse the necessary modules, collecting transform and resolution results with *no* mutation - using a separate, ephemeral `GraphTraversal` instance to make that explicit.
 2. Commit: Mutate graph state in a second, synchronous, depth-first traversal of the modified modules.

The motivation for this is twofold:
 - Resolve a category of known bugs where a userland issue, such as a syntax error or unresolvable dependency, would interrupt traversal and potentially leave internal state *partially* updated *without sending an update to the client*. This will now fail in the draft phase, allowing `DeltaCalculator` to repeat the attempt when the userland issue is fixed.
 - Makes the operation of `Graph` easier to reason about, and opens up several opportunities for simplification. For example, the commit phase being synchronous means that ordering is deterministic, and that the need to track `earlyInverseDependencies` is obviated.

Some of the issues fixed here were introduced by Graph delta pruning in D45691844, but others are longstanding. New unit tests cover the major cases.

Changelog:
```
**[Fix]**: Fast Refresh may fail or lose modifications after correcting transform or resolution issues in user code.
```

Differential Revision: D51943094

### DIFF
--- a/packages/metro/src/DeltaBundler/Graph.js
+++ b/packages/metro/src/DeltaBundler/Graph.js
@@ -86,10 +86,6 @@ type Delta<T> = $ReadOnly<{
   touched: Set<string>,
   deleted: Set<string>,
 
-  // A place to temporarily track inverse dependencies for a module while it is
-  // being processed but has not been added to `graph.dependencies` yet.
-  earlyInverseDependencies: Map<string, CountingSet<string>>,
-
   moduleData: Map<string, ModuleData<T>>,
 }>;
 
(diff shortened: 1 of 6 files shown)


robhogan:
Closed by https://github.com/facebook/metro/commit/0e51ad92f3f2fba3f90410d246009e5ade48db7c

PR #1163, opened 2024-01-02
image-size package version set to a static value 1.0.2

AdarshJais:
The image-size package recently released version 1.1.0, which no longer supports Node versions below 18. Due to the caret (^) dependency in the metro package, it automatically upgrades from version 1.0.2 to 1.1.0. However, this poses an issue for dependent packages like react-native (below version 0.72), which allows consumers to use Node versions below 18 and hence the current code makes it incompatible because of the updated image-size package.

## Summary

React Native versions below 0.73 permit the use of Node 16 for development. However, an issue arose suddenly, indicating that the image-size package is incompatible with the current Node version and now requires version 18 or above. Upon investigation, it was discovered that `image-size` is a dependency of the metro, specified with the caret (^) notation. The problem stems from the recent release of image-size version 1.1.0, which no longer supports Node versions below 18. This change is causing installation failures in React Native projects.

### DIFF
--- a/packages/metro/package.json
+++ b/packages/metro/package.json
@@ -29,7 +29,7 @@
     "error-stack-parser": "^2.0.6",
     "graceful-fs": "^4.2.4",
     "hermes-parser": "0.18.2",
-    "image-size": "^1.0.2",
+    "image-size": "1.0.2",
     "invariant": "^2.2.4",
     "jest-worker": "^29.6.3",
     "jsc-safe-url": "^0.2.2",


netroy:
It's not recommended to use nodejs 16 anymore, [even more so that the end-of-life date for v16 was moved to an earlier date because of unmaintained versions of openssl](https://nodejs.org/en/blog/announcements/nodejs16-eol).

It's fine if someone wants to use nodejs 16 on their own, but not project should bear the burden to support a version that nodejs doesn't themselves support anymore. 
Even this project requires a minimum of 18 [since 7 months](https://github.com/facebook/metro/commit/ebaf27ab49eb8b9394107bd7a796b24bc342c483).

netroy:
I've published `v1.1.1` of `image-size` that allows using nodejs v16. 
This change is not needed anymore.

robhogan:
Thanks for flagging this @AdarshJais and thanks again @netroy for jumping on this upstream. The new patch helps all of our React Native users on 0.72, which requires Node >=16.

Technically, React Native 0.71 still supports users on Node 14. For those folks who can't upgrade Node or React Native, I'd currently recommend using Yarn resolutions to force your `image-size` dependency to `1.0.2`. If that turns out to be a significant number of people, we'll need to merge a PR like this to Metro's `v0.73-stable` branch, as that's the version used by RN 0.71. Closing this PR, but thanks again @AdarshJais for kicking off this discussion and helping a bunch of our users.

PR #1172, opened 2024-01-10
Fix `metro-minify-terser` being resolved from the wrong folder

tido64:
## Summary

In a pnpm-like setup, `metro-minify-terser` gets resolved from the wrong folder.

## Test plan

n/a


tido64:
cc @huntie @robhogan

tido64:
@robhogan: Is this no longer needed?

robhogan:
> @robhogan: Is this no longer needed?

Not urgently, since we added the dependency do `metro-transform-worker` in the last release  - thanks for the PR though,  as well as identifying the problem it kicked off a good discussion about how we should be configuring these kinds of dependencies (relative vs absolute module paths vs code injection, etc). Cache portability makes things a little tricky.

We’ll revisit this kind of config holistically in a future breaking release and might merge this PR then - it may be the best approach long term, but if you’d rather close it for now, feel free.

Issue #1167, opened 2024-01-06
[`metro-transform-plugins`] Function calls prefixed with void together with optional chain (?.) are skipped on production (`dev: false`).


Gamote:
As the title suggests, today we found out that the `constant-folding-plugin` replaces this call `void foo?.();` with `undefined;` on production (`dev: false`).

We found this issue while using Expo SDK 50 in production mode. In development mode works just fine because the `constant-folding-plugin` is skipped. For more information and a MRE please see this issue https://github.com/expo/expo/issues/26276.

After some investigation we found out that we can fix this by adding `OptionalCallExpression: unsafe` in the `evaluate` method of the plugin. I am not 100% that this is the way to go as I don't have a lot of experience with Babel plugins and I might miss important implications. Please let me know if you have some guidance and I will be more than happy to proceed with a PR.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/metro-transform-plugins/src/constant-folding-plugin.js b/node_modules/metro-transform-plugins/src/constant-folding-plugin.js
index d0dd8ac..f1e96c6 100644
--- a/node_modules/metro-transform-plugins/src/constant-folding-plugin.js
+++ b/node_modules/metro-transform-plugins/src/constant-folding-plugin.js
@@ -27,8 +27,15 @@ function constantFoldingPlugin(context) {
     };
     path.traverse(
       {
-        CallExpression: unsafe,
         AssignmentExpression: unsafe,
+        CallExpression: unsafe,
+        /**
+         * This will mark `foo?.()` as unsafe, so it is not replaced with `undefined` down the line.
+         *
+         * We saw this case in the wild, where the unary expression `void foo?.()` was replaced with `undefined`
+         * resulting in the expression call being skipped.
+         */
+        OptionalCallExpression: unsafe,
       },
       state
     );
```

Gamote:
I have started a PR, hopefully it is the right direction: https://github.com/facebook/metro/pull/1178.

PR #1178, opened 2024-01-13
fix(#1167): function calls prefixed with `void` together with the optional chain (?.) are skipped

Gamote:
## Summary

Fixes: https://github.com/facebook/metro/issues/1167

`constant-folding-plugin` replaces this call `void foo?.();` with `undefined;`.

This is because the `evaluate()` function marks the optional call expressions as safe, making the `UnaryExpression` visitor think it is safe to replace it with the `value` returned by the `evaluate()`, in this case `undefined`.

This PR makes the `evaluate()` function mark the optional call expressions as unsafe.

Changelog:
```
* **[Fix]**: `constant-folding-plugin`: Don't fold optional function calls (`foo.?()`). (https://github.com/facebook/metro/pull/1178 by @Gamote)
```

## Test plan

I have added 2 tests to cover these changes:
- does not transform optional chained call into `undefined`
- does not transform `void` prefixed optional chained call into `undefined`

### DIFF
--- a/packages/metro-transform-plugins/src/__tests__/constant-folding-plugin-test.js
+++ b/packages/metro-transform-plugins/src/__tests__/constant-folding-plugin-test.js
@@ -361,4 +361,20 @@ describe('constant expressions', () => {
 
     compare([constantFoldingPlugin], code, expected);
   });
+
+  it('does not transform optional chained call into `undefined`', () => {
+    const code = `foo?.();`;
+
+    const expected = `foo?.();`;
+
+    compare([constantFoldingPlugin], code, expected);
+  });
+
+  it('does not transform `void` prefixed optional chained call into `undefined`', () => {
+    const code = `void foo?.();`;
+
+    const expected = `void foo?.();`;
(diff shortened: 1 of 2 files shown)


robhogan (approved):

Gamote:
> Facebook Internal - Builds & Tests

@robhogan Is there anything specific that I can do to fix this?

robhogan:
> > Facebook Internal - Builds & Tests
> 
> @robhogan Is there anything specific that I can do to fix this?

Thanks, but nothing's broken, just a couple of flaky tests. We're just waiting for internal review at the moment, shouldn't be long.

Gamote:
> > > Facebook Internal - Builds & Tests
> 
> > 
> 
> > @robhogan Is there anything specific that I can do to fix this?
> 
> 
> 
> Thanks, but nothing's broken, just a couple of flaky tests. We're just waiting for internal review at the moment, shouldn't be long.

Ah alright. Thank you for the information.

PR #1195, opened 2024-01-19
Fix `inlinePlugin` replacing `__DEV__` in invalid Babel AST locations

kitten:
## Summary

We've encountered this as part of a report on the `expo` repo: https://github.com/expo/expo/issues/25965

There, an issue is caused by a dependency using an export alias named `__DEV__`, which the `inlinePlugin` is trying to replace:

```js
export { DEV as __DEV__ };
```

This only occurs in certain cases, since usually, as part of the iOS/Android builds, we would be applying the `inlinePlugin` only after export statements, where this issue may occur, have been transpiled away.

However, looking at this issue, I found more cases that would fail with the current implementation of the `inlinePlugin`. This is part of what I added in tests but not an exhaustive list:
- Shorthand object methods (`{ __DEV__() {} }`)
- Labelled statements (`__DEV__: {};`)
- Class methods (`class { __DEV__() {} }`)
- Optional property member access (`x?.__DEV__`)

All of these issues can be addressed by letting this replacement use `ReferencedIdentifier` as a visitor, rather than just `Identifier`.

## Test plan

- Tests have been added to `inline-plugin-test.js` to reflect the mentioned cases.

### DIFF
--- a/packages/metro-transform-plugins/src/__tests__/inline-plugin-test.js
+++ b/packages/metro-transform-plugins/src/__tests__/inline-plugin-test.js
@@ -16,45 +16,106 @@ const inlinePlugin = require('../inline-plugin');
 const stripFlow = require('@babel/plugin-transform-flow-strip-types');
 
 describe('inline constants', () => {
-  it('replaces __DEV__ in the code', () => {
-    const code = `
-      function a() {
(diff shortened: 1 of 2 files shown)


motiz88 (approved):
Looks great, thanks for catching this and for adding tests!

PR #1206, opened 2024-02-02
`metro-config`: Add `injectProjectRootAsWatchFolder` argument to `loadConfig ` and provide fallback for `watchFolders`

kraenhansen:
## Summary

As per https://github.com/facebook/react-native/pull/41967#discussion_r1475235748 I would like to expose a way to disable the code path which overrides the `watchFolders` to be able to detect if the user provided a value for it, or not.

## Test plan

I added a test to verify the expected behaviour:

https://github.com/kraenhansen/metro/blob/e6de60117ea1e099cb417131d57de52250338ca2/packages/metro-config/src/__tests__/loadConfig-test.js#L169-L182


kraenhansen:
The integration tests failed, the failure seems unrelated to my changes and they pass when ran locally.
Perhaps it needs a re-run?

PR #1212, opened 2024-02-06
feat: provide `extends` directive for configs

### LINKED ISSUE #977: Feature: Add `extends` config loading support
## 👋🏻 Good First Issue

- An issue or task that is ideal for new open source contributors.
- Please reach out to the issue creator before claiming an issue to discuss design.

## Description

We'd like Metro's configuration to accept an `extends` key, similar to other tools such as ESLint and TypeScript.

This would benefit Metro users in enabling a more convenient and concise API that abstracts away the need for `mergeConfig()` today.

```js
// metro.config.js
module.exports = {
  extends: ['@react-native/metro-config'],
  ...
};
```

Related context:

- https://github.com/facebook/react-native/pull/36502

### Rough design

The implementation for this feature should live in the `metro-config` package.

- Update the relevant `MetroConfig` type constituents to add `extends?: $ReadOnlyArray<string>`.
- Update [`loadMetroConfigFromDisk`](https://github.com/facebook/metro/blob/d0d554381f119bb80ab09dbd6a1d310b54737e52/packages/metro-config/src/loadConfig.js#L180) to read `extends` if present.
    - For each value in `extends`, attempt to `require` the config path relative to the `filepath` of the input config file, and apply the imported object in order via `mergeConfig()`.
    - Throw an error if a config path cannot be resolved.
- Add one or more test cases in `packages/metro/src/integration_tests/`.
- Update documentation to illustrate usage of this feature.

antongolub:
This PR brings declarative config extension via `extends` references. Just like `tsconfig`, `eslint`, `typedoc` and many others.

* Accepts: `string | Record<any, any> | Array<string | Record<any, any>>`
  * 'foo' → require('foo')
  * './foo.js' → require(path.resolve(<parent cfg>, 'foo.js'))
  * ['foo', 'bar']  → merge(base, ...extras.load(...))
  * [{"extends": "./base.json", "foo": "bar"}] → ...

closes #977


antongolub:
@huntie, could you clarify some things?

The actual cosmiconfig preset defines `searchPlaces` without `.yml` and '.es6' (mjs?), but adds appropriate loaders. This looks inconsistent. What is the desired behavior?

```js
const explorer = cosmiconfig('metro', {
  searchPlaces: [
    'metro.config.js',
    'metro.config.cjs',
    'metro.config.json',
    'package.json',
  ],
  loaders: {
    '.json': cosmiconfig.loadJson,
    '.yaml': cosmiconfig.loadYaml,
    '.yml': cosmiconfig.loadYaml,
    '.js': cosmiconfig.loadJs,
    '.cjs': cosmiconfig.loadJs,
    '.es6': cosmiconfig.loadJs,
    noExt: cosmiconfig.loadYaml,
  },
});
```

robhogan:
Hi @antongolub - apologies this sat unreviewed for so long, and sorry to be closing it after that wait.

We've decided against an `extends` construct - see #977 for the reasoning. The ergonomics are covered by `mergeConfig` accepting functions (#1580) and by the array-exporting config form proposed in #1753, and I'd rather not add a third way to do the same thing.

On your question above - it's moot now. We dropped cosmiconfig in #1553 and do our own config file discovery, with ESM and TypeScript configs supported since, so this would need a substantial rewrite regardless.

Thanks for the work and apologies again.

PR #1226, opened 2024-02-21
fix: skip errors that occurred on virtual lines

EvanBacon:
## Summary

When Metro cannot find a module being imported, it shows an error with a small preview of your code around that import. Some build tools automatically add imports while transforming your code. Those added imports can point to a line number that does not exist in your saved file. Before this change, Metro crashed while building that preview, which hid the real "Unable to resolve module" error. Now Metro skips lines that do not exist in the file and still shows the real error with a preview of the lines that do exist.

### Changes
- ModuleResolution.js: skip missing lines when searching for the import location, falling back to the default location.
- Adds an integration test with a Babel plugin that appends an unresolvable import past the end of the file.

## Test Plan
Added integration test "reports resolution errors on locations past the end of the source" in build-errors-test.js. It uses a test transformer that appends import "./does-not-exist" at a line past the end of the file, then verifies the build still fails with "Unable to resolve module ./does-not-exist" and shows the existing code instead of crashing.

## Changelog
[Fix] Fixed a crash when building the error preview for an unresolvable import pointing past the end of the file (for example, imports added automatically during the build). Metro now skips the missing lines and reports the real "Unable to resolve module" error.

Original contribution from https://github.com/react/metro/pull/1226 by evanbacon.


motiz88:
Where is the bad source location coming from? This kinda looks like it masks a genuine bug elsewhere. Do you have repro steps? Also, we have [tests](https://github.com/facebook/metro/commit/a398f34a34a9b385af6c2bd64ea3ad161913b067) for this code now, might be worth adding a regression test to cover the fix.

EvanBacon:
@motiz88 If you have a babel plugin which takes an empty file like `""` and injects a bunch of imports:

```
import "missing"
import "missing2"
```

Then the resolver will throw an error which breaks the formatter because it attempts to read lines that don't exist on-disk.

robhogan:
@vzaidman I've rebased this and added tests to show how it's reproducible with legit transform. Good to land IMO.

(Cheers @EvanBacon ;))

vzaidman:
@robhogan can you please add a comment in tests or in the code or a screenshot on this issue, demonstrating how the thrown error would look like?

robhogan:
> @robhogan can you please add a comment in tests or in the code or a screenshot on this issue, demonstrating how the thrown error would look like?

@vzaidman sure - the basic recipe to repro is a resolution error in a generated import, which is handled fine (an actionable resolution error, pointing to the user's source or falling back gracefully) _except where the generated import happens to be beyond the length of the untransformed source_, which crashes - that's arbitrary, as there's no reason the transformed source should be shorter than the original - in fact it's usually longer.

We've got away with this largely because imports are usually near the top of the file, so it's rare that a resolution failure blames to a high line number. That's obviously pretty fragile!

The error thrown in the terminal is:

```
TypeError: Cannot read properties of undefined (reading 'length')
    at length (packages/metro/src/node-haste/DependencyGraph/ModuleResolution.js:377:64)
    at UnableToResolveError.refineDependencyLocation [as buildCodeFrameMessage] (.../ModuleResolution.js:344:9)
    at new buildCodeFrameMessage (.../ModuleResolution.js:316:35)
    at ModuleResolver.resolveDependency (.../ModuleResolution.js:170:15)
    at DependencyGraph.resolveDependency (.../DependencyGraph.js:290:43)
``` 

Where we'd expect to see:
```
UnableToResolveError: Unable to resolve module ./does-not-exist from .../build-errors/transform-injected-import.js:
```

Running the new Jest test with the fix reverted, the test fails as expected:
```
✕ reports resolution errors on locations past the end of the source
  Expected pattern: /^Unable to resolve module \.\/does-not-exist/
  Received string:  "Cannot read properties of undefined (reading 'length')"
```

PR #1230, opened 2024-03-05
feat: add `unstable_renameRequire` to allow disabling the require rename transform

EvanBacon:
## Summary

Renaming `require` to `_$$_REQUIRE` is an extraneous babel traversal that apparently is used internally at Meta for additional dependency extraction. Since the modules are scoped as iife functions, and there's no special convention to preserve native `require` syntax, this transform does not appear to add any value to standard projects. Per advice from @robhogan, I've added a flag to disable this transform, which we can enable by default in Expo CLI. 

Saves ~165ms when transforming the production react renderer module.


EvanBacon:
Bundling our kitchen sink app (3301 modules):

`bun expo export -p ios --no-bytecode --max-workers=0 --no-minify -c `

- Before: 61,953ms
- After: 58,145ms

I'm finding I can only accurately measure savings with threading disabled. With threads enabled, the results can sometimes be slower, but are ~100ms faster (total average: 27,491ms).

PR #1233, opened 2024-03-13
feat(config): support typescript configs

Titozzz:
## Summary

Allowing the user to user ts config will reduce configuration mistakes

I had to revert https://github.com/facebook/metro/commit/05b9ed4446c31c44f310fce1fbc86f3240084a85 as jiti uses os.tmpdir and it would conflict. But it seems to be running fine with node 18/20 at least

## Test plan

Unsure how to test this, but the code seems pretty straightforward
Happy to discuss if testing is needed

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -3,11 +3,13 @@ id: configuration
 title: Configuring Metro
 ---
 
-A Metro config can be created in these three ways (ordered by priority):
+A Metro config can be created in these five ways (ordered by priority):
 
 1.  `metro.config.js`
-2.  `metro.config.json`
-3.  The `metro` field in `package.json`
+2.  `metro.config.ts`
+3.  `metro.config.cjs`
+4.  `metro.config.json`
+5.  The `metro` field in `package.json`
 
 You can also give a custom file to the configuration by specifying `--config <path/to/config>` when calling the CLI.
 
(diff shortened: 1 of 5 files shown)


robhogan (changes requested):
Thanks for this, this would be a nice feature.

I do have some concerns about the dependency on jiti here - it's a very oddly assembled package that claims zero dependencies, but actually bundles (some version of) Babel as a 1.6MB file, discarding its licence in the process: https://unpkg.com/browse/jiti@1.21.0/dist/ .

That, and the fact that it's modifying the environment enough to break tests in weird ways, means we really can't accept a dependency on it.

As a possible alternative, transforming files is somewhat within Metro's wheelhouse - we obviously already have a dependency on Babel - and based on https://github.com/Codex-/cosmiconfig-typescript-loader/blob/main/lib/loader.ts it looks like implementing a loader is pretty straightforward - if it's only a few lines (plus tests), and a Babel plugin to roll our own, that might be the best option.

Aside: We're planning to update cosmiconfig for ESM support but are currently blocked by a Node<20 bug, details: https://github.com/react-native-community/cli/issues/2167#issuecomment-1807128723

> Allowing the user to user ts config will reduce configuration mistakes

Curious what you mean by this - I can see that TS would be nicer both syntactically and for consistency, but TS should already be checking the JSDoc types in metro.config.js - it's a template bug if not. Was it not working for you?

robhogan:
Thanks again, closing as TS support was released in v0.83.2 and is now documented.

Issue #1236, opened 2024-03-19
Package Exports resolution differs from Node in 2 edge cases


jbroma:
**Do you want to request a *feature* or report a *bug*?**
Bug

Hey metro team! 👋 

I was working on aligning Re.Pack's resolution setup to follow `metro-resolver` with RN defaults closely and found these 2 edge-cases were `metro-resolver` behaves in a weird way when dealing with Package Exports:

#### 1. Root Array Shorthand

For the Root Array shorthand, the implementation differs when compared to Node.js or Webpack's `enhanced-resolve` specification:

```json
{
  "name": "@metro-package-exports/root-shorthand",
  "exports": ["./index.js"]
}
```

Results:

For the import specifier: `@metro-package-exports/root-shorthand`

- **Node.js**: Resolves to `index.js` inside of `@metro-package-exports/root-shorthand`.
- **enhanced-resolve**: Identically resolves to `index.js` within `@metro-package-exports/root-shorthand`.
- **metro-resolver**: Fails to resolve, falling back to file resolution and ultimately resolving to the `@metro-package-exports/root-shorthand` directory itself.

#### 2. Subpath Patterns in `exports`

When using subpath patterns in `exports` to match the import specifier, the behaviors observed are as follows:

```json
{
  "name": "@metro-package-exports/restricted",
  "exports": {
    ".": "./index.js",
    "./features/*.js": "./features/*.js",
    "./features/internal/*.js": null
  }
}
```

Results:

For the import specifier: `@metro-package-exports/restricted/features/internal/b.js`

- **Node.js**: Fails to resolve as `b.js` is not exported from `@metro-package-exports/restricted`.
- **enhanced-resolve**: Also fails to resolve for the same reason as Node.js.
- **metro-resolver**: Successfully resolves to `@metro-package-exports/restricted/features/internal/b.js` without falling back to file resolution.

#### Reproduction repository

https://github.com/jbroma/metro-package-exports

#### Setup

1. Install dependencies via `yarn` v1 (classic).
2. Make the test scripts within the `tester` directory executable by running `chmod +x` on them.
3. Execute `test-root-shorthand.sh` & `test-restricted.sh` scripts.

**What is the expected behavior?**

Resolution in edge cases above should follow Node.js resolution algorithm

huntie:
Thanks @jbroma!

> **1. Root Array Shorthand**

[...]

PR #1238, opened 2024-03-21
fix(metro-resolver): package exports array root shorthand

jbroma:
## Summary

This PR deals with array root shorthand edge case from #1236 

- [x] - fixed array root shorthand spec in tests to match Node.js resolution algorithm
- [x] - aligned implementation of package exports for array root shorthand 
- [x] - added extra `[nonstrict]` test case to highlight that non-declared paths from array should not work normally 

Changelog: [Experimental] Fix package exports array root shorthand resolution to behave like node (edge case)

## Test plan
- [x] - all tests pass


packages/metro-resolver/src/__tests__/package-exports-test.js
```diff
@@ -255,6 +255,51 @@ describe('with package exports resolution enabled', () => {
     });
   });
 
+  describe('array root shorthand', () => {
```
huntie, line 258: Can we group this `describe` group under `describe('main entry point', () => {`? Will make the "should use first valid value" test more understandable in context.

huntie (commented):
Thanks! I'm still trying to find where the previous root shorthand behaviour was documented. Lining up with current Node makes sense 👍🏻

packages/metro-resolver/src/__tests__/package-exports-test.js
```diff
@@ -255,6 +255,51 @@ describe('with package exports resolution enabled', () => {
...
+    });
+
+    test('[nonstrict] should fall back and log warning when subpath is not listed', () => {
+      expect(Resolver.resolve(context, 'test-pkg/index.js', null)).toEqual({
+        type: 'sourceFile',
+        filePath: '/root/node_modules/test-pkg/index.js',
+      });
+      expect(logWarning).toHaveBeenCalledTimes(1);
+      expect(logWarning.mock.calls[0][0]).toMatchInlineSnapshot(
+        `"Attempted to import the module \\"/root/node_modules/test-pkg/index.js\\" which is not listed in the \\"exports\\" of \\"/root/node_modules/test-pkg\\". Falling back to file-based resolution. Consider updating the call site or asking the package maintainer(s) to expose this API."`,
```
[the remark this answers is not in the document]
jbroma: yes, if package exports are defined then import like the one above should not work, see the output from repro [here](https://app.warp.dev/block/LCzgyHtD0J1LHaQfLd7HmD)

packages/metro-resolver/src/__tests__/package-exports-test.js
jbroma: I've slightly modified the structure, it made more sense to move the `[nonstrict]` test to the rest of them under `main entry point`. I've also moved the `describe` group as suggested above 👍

packages/metro-resolver/src/__tests__/package-exports-test.js
```diff
@@ -211,6 +211,40 @@ describe('with package exports resolution enabled', () => {
       `);
     });
 
+    test('[nonstrict] should fall back to file resolution and log warning when subpath is not listed', () => {
```
huntie, line 214: nit: Ideally, can we split this new test into a separate PR? Or convince me why it belongs here :)

jbroma: hmm this is a test that I started implementation with. It was checking for a behaviour that was not spec compliant. I'm not sure if it makes sense to keep it at all now.

I've removed it from this PR and can add it inside of a separate PR if you want, up to you :)

huntie (approved):

jbroma:
@robhogan oh I absolutely agree - there is more to this than what meets the eye :)

I think we should proceed with merging this edge case and handle other scenarios in separate issues / PR's.
The normalisation that you've provided seems like a good starting point to begin with. We can reuse [the repro](https://github.com/jbroma/metro-package-exports) from the #1236 and just add more cases there. 

Is there anything else that needs to be addressed in this PR? I see the FB internal job failed for some reason.

jbroma:
@huntie @robhogan is there anything blocking us from proceeding with this PR?

PR #1239, opened 2024-03-21
fix(metro-resolver): package exports restricted paths

jbroma:
## Summary

This PR deals with restricting internal paths through the most specific subpath pattern edge case from #1236 

- [x] - added test for implicit default condition to match Node.js resolution algorithm
- [x] - aligned implementation of `reduceConditionalExport` to account for implicit default condition

Changelog: [Experimental] Fix implicit default condition to be `null` for subpath patterns (edge case)

## Test plan
- [x] - all tests pass


huntie (approved):
LGTM, importing to run on the wider Meta codebase

PR #1240, opened 2024-03-21
refactor(metro-resolver): more descriptive assertions in should resolve exports target directly tests

jbroma:
## Summary

This PR adds comments to assertions in tests that check that neither `sourceExts` or `platformExts` are expanded when matching exports target directly, which was suggested inside of #1236 by @huntie. The assertions rely on using `[nonstrict]` fallback mechanism as there is no other way to test this functionality right now because `strict` mode is not implemented.

Changelog: [Internal] Refactor Package Exports `sourceExts` and `platformExts` tests that rely on fallback

## Test plan
- [x] - all tests pass

### DIFF
--- a/packages/metro-resolver/src/__tests__/package-exports-test.js
+++ b/packages/metro-resolver/src/__tests__/package-exports-test.js
@@ -115,6 +115,7 @@ describe('with package exports resolution enabled', () => {
           '/root/node_modules/test-pkg/package.json': '',
           '/root/node_modules/test-pkg/index.js': '',
           '/root/node_modules/test-pkg/index-main.js': '',
+          '/root/node_modules/test-pkg/index-exports.js': '',
           '/root/node_modules/test-pkg/index-exports.js.js': '',
           '/root/node_modules/test-pkg/index-exports.ios.js': '',
           '/root/node_modules/test-pkg/symlink.js': {
@@ -225,14 +226,14 @@ describe('with package exports resolution enabled', () => {
       test('without expanding `sourceExts`', () => {
         expect(Resolver.resolve(context, 'test-pkg', null)).toEqual({
           type: 'sourceFile',
-          filePath: '/root/node_modules/test-pkg/index-main.js',
+          filePath: '/root/node_modules/test-pkg/index-exports.js',
         });
       });
 
       test('without expanding platform-specific extensions', () => {
         expect(Resolver.resolve(context, 'test-pkg', 'ios')).toEqual({
           type: 'sourceFile',
-          filePath: '/root/node_modules/test-pkg/index-main.js',
+          filePath: '/root/node_modules/test-pkg/index-exports.js',
         });
       });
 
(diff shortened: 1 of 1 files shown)


packages/metro-resolver/src/__tests__/package-exports-test.js
```diff
@@ -115,6 +115,7 @@ describe('with package exports resolution enabled', () => {
           '/root/node_modules/test-pkg/package.json': '',
           '/root/node_modules/test-pkg/index.js': '',
           '/root/node_modules/test-pkg/index-main.js': '',
+          '/root/node_modules/test-pkg/index-exports.js': '',
```
huntie, line 118: The one thing with this change is we make the test case against expanding `index-exports.js` -> `index-exports.js.js` ambiguous. Instead, let's update the overriden `package.json` fields in the `'should resolve "exports" target directly'` test group.

L216:

```diff
      const context = {
        ...baseContext,
        ...createPackageAccessors({
          '/root/node_modules/test-pkg/package.json': {
-           main: 'index-main.js',
            exports: './index-exports.js',
          },
        }),
      };
```

packages/metro-resolver/src/__tests__/package-exports-test.js
```diff
@@ -225,14 +226,14 @@ describe('with package exports resolution enabled', () => {
       test('without expanding `sourceExts`', () => {
         expect(Resolver.resolve(context, 'test-pkg', null)).toEqual({
           type: 'sourceFile',
-          filePath: '/root/node_modules/test-pkg/index-main.js',
+          filePath: '/root/node_modules/test-pkg/index-exports.js',
```
huntie, line 229: With the other comment and change, I think this is the right assertion instead. Note that, with no option to disable strict behaviour today, we can't assert the resolver throws an error in this case. But instead, a comment makes this sufficiently easy to understand, IMO.

```suggestion
          // [nonstrict] Falls back to index.js based on file resolution
          filePath: '/root/node_modules/test-pkg/index.js',
```

huntie (commented):
Love this 💯 I was drafting the same changes locally 😄. However can we make this even more focused/strict — see inline feedback.

packages/metro-resolver/src/__tests__/package-exports-test.js
jbroma: agreed 💯

packages/metro-resolver/src/__tests__/package-exports-test.js
jbroma: Hmm regarding this, I think we should just add another test cases for testing the fallback, as right now with changes from this PR we directly test not expanding `sourceExts` & `platformExts` while falling back proves the functionality but indirectly, if that makes any sense, WDYT?

huntie: It is indirect, but is the only observable behaviour we have right now 😅 We could also add a `logWarning` assertion in these to make even clearer — but I think the comments are good enough.

jbroma:
@huntie applied your suggestions, revised PR description & added changelog

huntie (approved):

PR #1284, opened 2024-06-05
fix: apply custom serializer to source map endpoint in dev server

EvanBacon:
## Summary

- The source maps for Expo CLI are off by 1, but only in development. The cause is that source maps are generated differently between the server endpoint and the `build` function. In the Expo metro-config, we use the serializer to inject custom pre-modules for things like SSR and client-side environment variables. These changes aren't reflected in the source map endpoint.
- This PR addresses the differences by applying a custom serializer to the graph if one exists, and returning the map from the serializer if one is returned.
- The Expo CLI-side of this PR can be found here https://github.com/expo/expo/pull/29463

Changelog: [Fix] Apply custom serializer to source map graph in `.map` endpoint.

### DIFF
--- a/packages/metro/src/Server.js
+++ b/packages/metro/src/Server.js
@@ -1153,6 +1153,55 @@ class Server {
         prepend = [];
       }
 
+      // If the project defines a custom serializer, then we should run it to ensure any
+      // mutations to the graph are applied in the sourcemap.
+      if (this._config.serializer.customSerializer) {
+        const bundle = await this._config.serializer.customSerializer(
+          entryFile,
+          prepend,
+          graph,
+          {
+            asyncRequireModulePath: await this._resolveRelativePath(
+              this._config.transformer.asyncRequireModulePath,
+              {
+                relativeTo: 'project',
+                resolverOptions,
+                transformOptions,
+              },
+            ),
+            processModuleFilter: this._config.serializer.processModuleFilter,
+            createModuleId: this._createModuleId,
+            getRunModuleStatement:
+              this._config.serializer.getRunModuleStatement,
+            includeAsyncPaths: graphOptions.lazy,
+            dev: transformOptions.dev,
+            projectRoot: this._config.projectRoot,
+            modulesOnly: serializerOptions.modulesOnly,
+            runBeforeMainModule:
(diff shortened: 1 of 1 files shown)


packages/metro/src/Server.js
```diff
@@ -1153,6 +1153,55 @@ class Server {
         prepend = [];
       }
 
+      // If the project defines a custom serializer, then we should run it to ensure any
+      // mutations to the graph are applied in the sourcemap.
```
motiz88, line 1157: It makes sense to me to run the custom serializer consistently (and I vaguely remember seeing cases where this was broken), but what do you mean by "mutations to the graph"? When during bundling do those occur?

motiz88 (commented):
> This doesn't address the same issue occurring in error symbolication

Yeah we should definitely make sure those are all consistent.

packages/metro/src/Server.js
EvanBacon: If someone were to `preModules.splice(0,0, { ... })` to add a new module inside of a custom serializer, then it would technically mutate the arguments used to generate the bundle. I can think of two cases like this:
- [sentry](https://github.com/getsentry/sentry-react-native/blob/0019f64e4d91eddd0c5c8cdc369778caecc4815c/src/js/tools/sentryMetroSerializer.ts#L126-L138) (not exactly, but close) 
- One I just wrote for Expo CLI as a fallback to returning the map: [here](https://github.com/expo/expo/pull/29463/files#diff-fd86a91bdc3cee21b84074550e6ec3a5651e6c6ca2b3dd9b9eccfa9e1b13aa2aR100)

motiz88: Hmm. The serializer API wasn't designed with side-effectful serializers in mind, and the `preModules` parameter specifically is typed as being read-only. Is there a way to achieve what you're trying to do without mutating read-only arguments?
 
image

EvanBacon: Yeah but it's a little unstable. I'm sniffing if the sourceUrl has a `.map` pathname to determine if the serializer should return source maps. This works for the `.map` endpoint, but it doesn't account for `/symbolicate` where the result is `getExplodedSourceMap` and not a standard source map string.

Issue #595, opened 2020-10-13
Unable to resolve module `..\\__generated__\\...` in React Native on Windows with Relay artifactDirectory option


benmurden:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Error module not found

**If the current behavior is a bug, please provide the steps to reproduce and a minimal repository on GitHub that we can `yarn install` and `yarn test`.**
Related to https://github.com/facebook/relay/issues/3201, but thought I should post here because it may be specifically a Metro issue.
```
OS: Windows 10
Node: v12.18.3
react: 16.13.1
react-native: 0.63.3
react-relay: 10.0.1
```
Reproduction repository using the SWAPI server from here https://github.com/graphql/swapi-graphql.

https://github.com/benmurden/react-native-relay-windows-repro

This only happens when using the `artifactDirectory` option in Relay, and only when bundled with React Native i.e. Metro bundler; this all works in React Web.

Sorry, not sure how this one could be tested with `yarn test`. To reproduce:
- Run the SWAPI server
- Clone the repro repo
- Change the `relay.environment.js` to match the port given by SWAPI server
- `yarn install && yarn relay && yarn android`

![image](https://user-images.githubusercontent.com/213062/95803272-a6a49400-0d3a-11eb-8649-e58169df8ceb.png)

Looks like a Windows directory issue, since the bundler seems to have prepended too many upward directory traversals (`..\\..\\..\\`).

**What is the expected behavior?**
App builds and runs on Windows 10 and Android.

RichardLindhout:
Do you have a workaround available?

RichardLindhout:
The isRelativeImport does not work on Windows it sees ..\\generated not as relative while it is ;)
```javascript
function isRelativeImport(filePath: string) {
  return /^[.][.]?(?:[/]|$)/.test(filePath);
}
```

We fixed it by adding 

```javascript

function isRelativeImport(filePath: string) {
if (filePath.startsWith('..\\') {
   return true;
}
  return /^[.][.]?(?:[/]|$)/.test(filePath);
}
```

RichardLindhout:
@benmurden thanks for pointing into the right direction!

benmurden:
@RichardLindhout Thanks! I was inspired to create a PR that will not only resolve the above issue, but also fix current directory relative imports (`.\\`).

#648

robhogan:
[...]

PR #1286, opened 2024-06-12
fix(metro-resolver): use proper `isRelativeImport` check that works with Windows paths

byCedric:
## Summary

Found an issue with the `isRelativeImport` from the resolver on Windows, resulting in the following issues:
- https://github.com/expo/expo/issues/29631
- https://www.reddit.com/r/expo/comments/1dd43gb/unable_to_resolve_module_node/

The summary of this is that Expo resolves the `main` field from the package during [static rendering here](https://github.com/expo/expo/blob/88d15528e68e6119f84a9e2be5f23775c40bc03a/packages/%40expo/cli/src/start/server/metro/MetroBundlerDevServer.ts#L263). On Windows, it results in a path like `.\\node_modules\\expo-router\\entry`, which is correct. Node can resolve this path using `require.resolve`:

image

Unfortunately, the `metro-resolver` can't handle this properly because of a single reason: `isRelativeImport` returns false for `.\\node_modules\\expo-router\\entry`, because the regex only tests posix paths, not Windows paths. It results in the following resolve chain:

1. [link](https://github.com/facebook/metro/blob/5c78f8d539474552b264626b5a99dc0d92243db3/packages/metro-resolver/src/resolve.js#L52-L58) - `resolveModulePath` is never called, as this path is not absolute, and incorrectly marked as not relative either
2. [link](https://github.com/facebook/metro/blob/5c78f8d539474552b264626b5a99dc0d92243db3/packages/metro-resolver/src/resolve.js#L60-L65) - `realModuleName` is also not false, as the module path is not redirected
3. [link](https://github.com/facebook/metro/blob/5c78f8d539474552b264626b5a99dc0d92243db3/packages/metro-resolver/src/resolve.js#L67-L86) - `isDirectImport` also results in `false`, for the same reasons as 1
4. [link](https://github.com/facebook/metro/blob/5c78f8d539474552b264626b5a99dc0d92243db3/packages/metro-resolver/src/resolve.js#L88-L94) - `context.allowHaste` is disabled during Expo static rendering
5. [link](https://github.com/facebook/metro/blob/5c78f8d539474552b264626b5a99dc0d92243db3/packages/metro-resolver/src/resolve.js#L96-L129) - Both `disableHierarchicalLookup` and `extraNodeModules` aren't set by default in Expo projects.
6. [link](https://github.com/facebook/metro/blob/5c78f8d539474552b264626b5a99dc0d92243db3/packages/metro-resolver/src/resolve.js#L131-L147) - `allDirPaths` already contain `<project>\\node_modules`, and when joining with `.\\node_modules\\expo-router\\entry`, it creates paths with `<project>\\node_modules\\node_modules\\...`, which are also incorrect.

<details><summary>Overview of all values through the debugger</summary>

image

</details>

This change adds the Windows path, as a valid path separator, to the `isRelativeImport` test - making it compatible with Windows.

Changelog: [Fix] Detect relative imports in the Metro Resolver correctly on Windows

## Test plan

Only reproducible on Windows.

- `$ npm create expo ./test-metro-resolve`
- `$ cd ./test-metro-resolve`
- `$ npx expo export --platform web`
- Failure due to the resolve issue I just mentioned

### DIFF
--- a/packages/metro-resolver/src/resolve.js
+++ b/packages/metro-resolver/src/resolve.js
@@ -508,7 +508,7 @@ function resolveWindowsPath(modulePath: string) {
 }
 
 function isRelativeImport(filePath: string) {
-  return /^[.][.]?(?:[/]|$)/.test(filePath);
+  return /^[.][.]?(?:[/\\]|$)/.test(filePath);
 }
 
 function normalizePath(modulePath: any | string) {


byCedric:
I see that Node 18.12 just timed out, it didn't fail. Also, might be nice to drop Node 16 (EOL version) some time, and start testing on Node 20 (current LTS version).

byCedric:
Node mostly uses the `node:path` module to resolve modules, which is built for interoperability between Windows and Linux/MacOS. I think that's the reason why it succeeds in Node, but I'm not sure if Node qualifies `/` as a "separator" in the spec.

- If they do, then the assumption that it only can be `/` is wrong.
- If they don't, then we have a bigger issue here since we also `C:\\some\\project` as project root, right?

byCedric:
To summarize, Metro assumes that all module specifiers (`from 'module-specifier'` / `require('module-specifier')` are notated in posix format, even on Windows.

This does not apply to folder references passed to Metro, e.g. the project root.

robhogan:
I set Claude off investigating the landscape on this, including tests on Node, Webpack, Vite, CJS and ESM, absolute and relative (https://github.com/robhogan/metro/pull/11).
 - The ESM spec clearly requires a `/` because specifiers are URLs, never paths. Node classifies anything not beginning `./` as a bare specifier and throws `ERR_INVALID_MODULE_SPECIFIER`. Vite is the only ESM implementation tested that accepts a leading `.\`, and only on Windows.
 - Nobody accepts a leading `.\` anywhere but running on Windows, so accepting it on all platforms would be unspecced behaviour unique to Metro.
 - For CommonJS, Vite and Node tolerate it but Webpack (enhanced-resolve) does not.

So I'd say the potentially acceptable version of this change is narrower - tolerate Windows separators only when running on Windows and only for CommonJS require, aligning with Vite, and Node's undocumented behaviour. But IMO the outcome of a resolution should be independent of the platform the resolver is running on, so I prefer to stay aligned with Webpack on this.

Bottom line: The motivating expo/expo#29631 and #595 were fixed upstream, and there isn’t a clear ecosystem alignment case for changing Metro’s behaviour.  So, closing - but thanks as ever!

Full results running on Windows:

| specifier | Node (CJS) | Node (ESM) | Webpack (enhanced-resolve) | Webpack (build) | Vite (ESM) | Vite (CJS) |
|---|---|---|---|---|---|---|
| `./sub/mod` | yes | yes | yes | yes | yes | yes |
| `../root/sub/mod` | yes | yes | yes | yes | yes | yes |
| `.\sub\mod` | yes | no | no | no | yes | yes |
| `..\root\sub\mod` | yes | no | no | no | yes | yes |
| `./sub\mod` | yes | yes | yes | yes | yes | yes |
| `.\sub/mod` | yes | no | no | no | yes | yes |
| absolute, native seps | yes | no | yes | yes | yes | yes |
| absolute, `/` seps | yes | no | yes | yes | yes | yes |
| absolute, `\` seps | yes | no | yes | yes | yes | yes |
| `file://` URL | no | yes | yes | yes | no | no |

PR #1300, opened 2024-06-26
fix: correctly format logs when saving bundle

szymonrybczak:
## Summary

When using `community-cli-plugin`, output when saving bundle looks like this:
```
❯ npx react-native bundle --platform android --entry-file index.js --dev=false --bundle-output dist/
                Welcome to Metro v0.80.9
              Fast - Scalable - Integrated

info Writing bundle output to:, dist/
```

There's a useless comma in log, because `community-cli-plugin` uses `logger.info` from `@react-native-community/cli-tools` which separates each item with `,` separator. 

## Test plan

Saving bundle output should contain correctly formatted log:

```
info Writing bundle output to: dist/
```

### DIFF
--- a/packages/metro/src/shared/output/bundle.flow.js
+++ b/packages/metro/src/shared/output/bundle.flow.js
@@ -49,7 +49,7 @@ async function saveBundleAndMap(
     ...
   },
   options: OutputOptions,
-  log: (message: string) => void,
+  log: (...args: Array<string>) => void,
 ): Promise<mixed> {
   const {
     bundleOutput,
--- a/packages/metro/src/shared/output/bundle.flow.js
+++ b/packages/metro/src/shared/output/bundle.flow.js
@@ -49,7 +49,7 @@ async function saveBundleAndMap(
     ...
   },
   options: OutputOptions,
-  log: (...args: Array<string>) => void,
+  log: (message: string) => void,
 ): Promise<mixed> {
   const {
     bundleOutput,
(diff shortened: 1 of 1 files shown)


szymonrybczak:
cc @huntie

packages/metro/src/shared/output/bundle.flow.js
```diff
@@ -49,7 +49,7 @@ async function saveBundleAndMap(
     ...
   },
   options: OutputOptions,
-  log: (...args: Array<string>) => void,
```
huntie, line 52: @szymonrybczak Hmm, can we drop this change? Rest looks good!

The interface here is that we expect a logger that is similar to `console.log()`'s API, which accepts multiple args that will be concatenated without a `,`.

- Happy with the other changes, which make this behaviour the most predictable wherever Metro is integrated.
- I'd prefer an upstream alignment in `@react-native-community/cli-tools`to `logger.info`. Our current types in `dev-middleware` assume this: https://github.com/facebook/react-native/blob/c0b5e2e031f0e471b57640be13fe086d36c48a99/packages/dev-middleware/src/types/Logger.js#L12-L15.

szymonrybczak: Yeah, I'm good with deleting this change. We just need to be careful when adding more logging to check if they're correctly formatted 👍

szymonrybczak: Removed in https://github.com/facebook/metro/pull/1300/commits/6f86b48aa16ed8a2b38bfbb3408b6040a82fe001

huntie (approved):

Issue #978, opened 2023-05-10
Feature: package.json "imports" field


huntie:
## 👋🏻 Good First Issue

- An issue or task that is ideal for new open source contributors.
- Please reach out to the issue creator before claiming an issue to discuss design.

## Description

[Incoming]

siddarthkay:
Hello @huntie 
I would love to work on adding this feature when you add some more description on the requirement and expectations :)

huntie:
Hey @siddarthkay! I've added a description now. If you're still happy to take this on, let me know :) (and feel free to [DM me on Twitter](https://twitter.com/alxhnt)).

siddarthkay:
@huntie : thanks for the detailed description, for now its good enough to get me started! for more questions I'll paste comments here!
Could you please assign this issue to me?

jameslawson:
@huntie  From the [discussion](https://github.com/facebook/metro/pull/995#issuecomment-1943329549)  in the linked pull request, I think this issue is up for grabs? (cc: @kraenhansen ) Or maybe @kraenhansen was interested in picking it up?

Either way, I've had a go at this myself here: https://github.com/facebook/metro/pull/1302 – please feel free to check it out and see if the changes there are on the right path 😄  - I managed to get some basic tests to pass 🎉

robhogan:
Yes, thanks! And thanks again @jameslawson for the implementation

PR #1302, opened 2024-07-09
Add support for resolving import subpaths

jameslawson:
## Summary

PR to add support in metro-resolver for resolving [import subpaths](https://nodejs.org/api/packages.html#subpath-imports); the feature discussed in Issue #978  in that reads the`"imports"` key in `package.json`

## Test plan

`npm run test`

## After Landing

- Update [the resolver algorithm doc](https://metrobundler.dev/docs/resolution#resolve) 
- Add support for resolving external packages

### DIFF
--- a/packages/metro-resolver/src/types.js
+++ b/packages/metro-resolver/src/types.js
@@ -52,18 +52,18 @@ export type FileCandidates =
       +candidateExts: $ReadOnlyArray<string>,
     };
 
-export type ExportMap = $ReadOnly<{
-  [subpathOrCondition: string]: string | ExportMap | null,
+export type ExportsLikeMap = $ReadOnly<{
+  [subpathOrCondition: string]: string | ExportsLikeMap | null,
 }>;
 
 /** "exports" mapping where values may be legacy Node.js <13.7 array format. */
 export type ExportMapWithFallbacks = $ReadOnly<{
-  [subpath: string]: $Values<ExportMap> | ExportValueWithFallback,
+  [subpath: string]: $Values<ExportsLikeMap> | ExportValueWithFallback,
 }>;
 
 /** "exports" subpath value when in legacy Node.js <13.7 array format. */
 export type ExportValueWithFallback =
-  | $ReadOnlyArray<ExportMap | string>
+  | $ReadOnlyArray<ExportsLikeMap | string>
   // JSON can also contain exotic nested array structure, which will not be parsed
   | $ReadOnlyArray<$ReadOnlyArray<mixed>>;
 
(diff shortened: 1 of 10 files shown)


packages/metro-resolver/src/types.js
```diff
@@ -78,6 +78,7 @@ export type PackageJson = $ReadOnly<{
   name?: string,
   main?: string,
   exports?: ExportsField,
+  imports?: ExportMap,
   ...
```
kraenhansen, line 82: This type name seem a bit odd now 🤔 Does that need an update?

kraenhansen:
Thanks for picking this up!

packages/metro-resolver/src/types.js
jameslawson: Yeah, while I was writing this it felt odd. While @huntie  originally suggested this, I could also see how renaming the type `ExportMap` to something like `PackageSubpathsMap` as a name that describes both import and export subpaths equally.

robhogan (approved):
Huge thanks for working on this @jameslawson and sorry for the lack of response from us. 

There have been a lot of moving parts in the resolver recently and this would've been a breaking change we couldn't immediately pull in, but there's now [room for it](https://github.com/facebook/metro/commit/c1c80c7727e01951a148e9f5a3d06196e98eaf72) and a new, stable `fileSystemLookup` to use.

If you're happy to rebase, this is looking really close.

robhogan:
We'll need to update [the resolver algorithm docs](https://metrobundler.dev/docs/resolution#resolve) as well, but I'm happy to take care of that when importing.

Beachman4:
Node allows you to alias to an external package, could you modify your implementation to allow that? Currently `reduceExportMap` throws an error because exports can't map to an external package.

robhogan:
> Node allows you to alias to an external package, could you modify your implementation to allow that? Currently `reduceExportMap` throws an error because exports can't map to an external package.

That's a great point, thanks for spotting that.

I think that can land as a follow-up though, I'm wary of holding this PR up further, but I've noted it as something that we need to do asap and before claiming support for `package.json#imports`.

Beachman4:
> > Node allows you to alias to an external package, could you modify your implementation to allow that? Currently `reduceExportMap` throws an error because exports can't map to an external package.
> 
> That's a great point, thanks for spotting that.
> 
> I think that can land as a follow-up though, I'm wary of holding this PR up further, but I've noted it as something that we need to do asap and before claiming support for `package.json#imports`.

Yeah I think that's fair. 

Just one last comment based on my usage of this WIP feature:

When there are references to other packages in the imports field, the current implementation will short circuit and throw an error as it considers those values to be invalid, which they are until it is implemented, but I think it would be better to compile the list of invalid values and echo that to the user, but still continue. The reason being that there could be valid values further down the import list and I think it's ideal if those are considered and processed.

The reason I would suggest this, is lets say this is released. I then pull down that change and add a custom resolveRequest function to my config to handle the external packages until support is added, but since that short circuit is there, this won't process the "good" subpath imports.

jameslawson:
@robhogan hey Rob, when you get a chance would you mind having another look at this PR. I've tried to address all comments previously raised; please consider it "Ready to Review". Thank you so much!

itsUndefined:
Any chance this PR can be picked up again? Would be awesome if Metro would support import subpaths for monorepo absolute paths support.

PR #752, opened 2022-01-06
Update hash function for node 17 compatibility

tobua:
When running a RN project currently with node 17 it will fail because a deprecated hash function is used with the [node crypto library](https://nodejs.org/en/blog/release/v17.0.0/#openssl-3-0). The error is `ERR_OSSL_EVP_UNSUPPORTED` and will always appear unless a flag is explicitly set. Only the `metro-cache` package is affected. The stable version node 17 has been around for 3 months and will become active LTS in April of this year.

**Summary**

Switching from a 128 bit `md4` hash to another 128 bit `md5` hash that's safe is enough to fix the issue. However, note that I'm neither a hash function nor a metro cache expert.

**Test plan**

Tested with stable node 17, with LTS node 16 and also works fine with lowest supported node 12 version.
An already built project can be upgraded without the need to clear the cache.

Fixes #722

Couldn't find any related issues in the main RN repository.

PR #1310, opened 2024-08-02
Add gh actions for building and testing the repo

vzaidman:
Introduced config that triggers GH actions that build, lint, and test the repo.

These are triggered when a pull request is created or it's latest commit is updated.

If a new job is triggered, the previous is abandoned.

* Please notice that tests are currently disabled for windows because they fail. This will be fixed in a subsequent diff.

Test:
* See the Github actions on this PR: 
Screenshot 2024-08-02 at 17 14 07

* Example of a run cancelled by pushing a new commit before the current is processed by Github actions: https://github.com/facebook/metro/actions/runs/10216106999

### DIFF
--- a/.github/actions/yarn-install/action.yml
+++ b/.github/actions/yarn-install/action.yml
@@ -1,0 +1,18 @@
+name: Install Dependencies
+inputs:
+  node-version:
+    required: false
+    default: '20.x'
+
+runs:
+  using: "composite"
+  steps:
+  - uses: actions/setup-node@v4
+    with:
+      node-version: ${{ inputs.node-version }}
+      cache: yarn
+      cache-dependency-path: yarn.lock
+      check-latest: true
+  - name: Install Dependencies
+    run: yarn install --frozen-lockfile --non-interactive --ignore-scripts
+    shell: bash
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,0 +1,51 @@
+name: facebook/metro/build-and-test
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+# head_ref is per PR, so this ensures that updating the latest PR commit
+# will cancel the previous run of the workflow and trigger a new one
+concurrency:
+  group: "build-and-test-${{ github.head_ref }}"
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  run-js-checks:
+    runs-on: ubuntu-latest
+    name: "Type check, lint, smoke test"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/yarn-install
+      - run: yarn typecheck
+      - run: yarn typecheck-ts
+      - run: yarn lint
+      - run: yarn test-smoke
+
+  test-with-coverage:
+    runs-on: ubuntu-latest
+    name: "Tests with coverage"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/yarn-install
+      - run: yarn test-coverage
+      - run: "./.github/scripts/install_codecov.sh"
+      - run: "./codecov -t ${{ secrets.CODECOV_TOKEN }} -f ./coverage/coverage-final.json"
+
+  test:
+    strategy:
+      matrix:
+        runs-on: ['ubuntu-latest']
+        node-version: ['18.0', '18.x', '20.x']
+    name: "Tests [Node.js ${{ matrix.node-version }}, ${{ matrix.runs-on }}]"
+    runs-on: ${{ matrix.runs-on }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/yarn-install
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Run Jest Tests
+        run: yarn jest --ci --maxWorkers 4 --reporters=default --reporters=jest-junit --rootdir='./'
--- a/.github/scripts/install_codecov.sh
+++ b/.github/scripts/install_codecov.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+# Install Codecov Uploader
+# See https://docs.codecov.com/docs/codecov-uploader#using-the-uploader-with-codecovio-cloud
+
+CODECOV_URL="https://uploader.codecov.io"
+
+curl "${CODECOV_URL}/verification.gpg" | gpg --no-default-keyring --keyring trustedkeys.gpg --import
+curl -Os "${CODECOV_URL}/latest/linux/codecov"
+curl -Os "${CODECOV_URL}/latest/linux/codecov.SHA256SUM"
+curl -Os "${CODECOV_URL}/latest/linux/codecov.SHA256SUM.sig"
+
+gpgv codecov.SHA256SUM.sig codecov.SHA256SUM
+shasum -a 256 -c codecov.SHA256SUM
+
+chmod +x codecov


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,52 @@
...
+    - uses: actions/checkout@v4
+    - uses: ./.github/actions/yarn-install
+    - run: yarn test-coverage
+    - run: "./.github/scripts/install_codecov.sh"
+    - run: "./codecov -t ${{ secrets.CODECOV_TOKEN }} -f ./coverage/coverage-final.json"
+
+  test:
+    strategy:
+      matrix:
+        runs-on: ['ubuntu-latest'] # 'windows-latest'
```
robhogan, line 42: IIRC we have a bunch of tests currently not passing on Windows unfortunately and this probably can't be enabled on `main` in the short term, but in CircleCI we have the convention of [`windows` branches to trigger CI on Windows](https://github.com/facebook/metro/blob/e3aaa0bac2d91598646341dfd5621c28cdf7370f/.circleci/config.yml#L133), which is useful if we're looking at a specific test suite.

Alternatively... if we could always run on Windows but have Jest automatically skip known-broken Windows tests *only* on Windows - maybe with a [`--filter`](https://jestjs.io/docs/cli#--filterfile) file, that'd be a much better setup.

package.json
```diff
@@ -94,5 +94,8 @@
   "dependencies": {},
   "engines": {
     "node": ">=18"
+  },
+  "volta": {
+    "node": "18.0.0"
```
robhogan, line 99: Is this related to the PR somehow?

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,52 @@
+name: facebook/metro/build-and-test
+on:
+  workflow_call:
+  pull_request:
+    types: [opened, synchronize]
+
+# head_ref is per PR, so this ensures that updaing the latest PR commit
```
robhogan, line 7:
```suggestion
# head_ref is per PR, so this ensures that updating the latest PR commit
```

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,52 @@
...
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  run-js-checks:
+    runs-on: ubuntu-latest
+    name: "Typescript, Link, Smoke Test"
```
robhogan, line 20:
```suggestion
    name: "TypeScript, Lint, Smoke Test"
```

.github/actions/yarn-install/action.yml
```diff
@@ -0,0 +1,23 @@
...
+      - 'min-supported'
+
+runs:
+  using: "composite"
+  steps:
+  - uses: actions/setup-node@v4
+    with:
+      # if both node-version-file and node-version are provided, node-version is used
+      node-version-file: 'package.json'
+      node-version: ${{ inputs.node-version-type == 'lts' && 'lts/*' || ''  }}
```
robhogan, line 18: I don't fully understand the use of `node-version-file`/`node-version` here - can we not use a matrix like the example at https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-a-matrix-for-your-jobs#example-using-a-single-dimension-matrix ?

robhogan (commented):
Thanks for taking this on! Just a couple of nits/questions

.github/workflows/build-and-test.yml
vzaidman: I was planning to fix the windows tests on a next commit and will enable it. WDYT?

package.json
vzaidman: Yes, this tells on which version of node tests will run.
Without it, setting `node-version-file: 'package.json'` will take version `20.2` based on the engine set to `"node": ">=18"`.
But I'd like to test the minimum node version we support and LTS instead.
See the improved comment on `.github/workflows/build-and-test.yml` below.

.github/actions/yarn-install/action.yml
vzaidman: I've updated the comment. please see if you agree with this approach. I can certainly hardcode the versions, but then the minimum supported version on `package.json` will be maintained separately from that hardcoded version in the pipelines, which is not the end of the world, but with no single point of truth they might diverge.

robhogan: My main concern here is the CI configuration effectively leaking into the repo (the `volta` field) in a way that's not obvious if you're looking at it from the `package.json` side. And, if you're a contributor actually using Volta, this is effectively breaking by forcing you to develop using 18.0.0, right?

> with no single point of truth they might diverge.

This is a fair point but I don't see it as a disaster:

 - Bumping the CI minimum without bumping Metro's minimum isn't a concern, I can't see a reason we'd ever do that voluntarily.
 - If we bumped Metro's minimum but forgot the CI minimum, the worst that happens is CI breaks some time later when we actually break compatibility, so we notice and update it. 
 - We probably *should* have an E2E test that builds and publishes, then installs Metro using something like Verdaccio, like RN does - then we'd fail fast because the `engines` requirement would cause installation to fail.
 
There are upsides to hardcoding the matrix - it's cleaner and more GHA-native (automatically reflected in test names, etc). I think that'd be my preference tbh. `['current', 'lts/*', 'lts/-1', '18.0']` would give us good coverage.

vzaidman: gotcha. I also spoke with @huntie about that.
could you please take a look at the updated code?

package.json
vzaidman: removed this change.

.github/actions/yarn-install/action.yml
```diff
@@ -0,0 +1,27 @@
...
+  current-node-version:
+    required: false
+    default: '20.2'
+
+runs:
+  using: "composite"
+  steps:
+  - uses: actions/setup-node@v4
+    with:
+      node-version: ${{ inputs.node-version-type == 'current' && inputs.current-node-version || inputs.min-supported-node-version }}
```
robhogan, line 22: I'm not sure what value this `node-version-type` indirection gives us - why not pass `node-version` through directly from the matrix, and then we can set it once at the top level?

.github/workflows/build-and-test.yml
robhogan: Afaik fixing tests on Windows isn't trivial but yeah we can follow up. I'd probably remove the commented-out `'windows-latest'` for now though.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,51 @@
...
+    - run: yarn test-coverage
+    - run: "./.github/scripts/install_codecov.sh"
+    - run: "./codecov -t ${{ secrets.CODECOV_TOKEN }} -f ./coverage/coverage-final.json"
+
+  test:
+    strategy:
+      matrix:
+        runs-on: ['ubuntu-latest'] # 'windows-latest'
+        node-version-type: ['current', 'min-supported']
+    name: "Tests [${{ matrix.runs-on }} machine, ${{ matrix.node-version-type }} Node.js]"
```
huntie, line 43: nit, also ideally let's print the numeric Node version, e.g. `Tests [Node.js 20, ubuntu-latest]`

```suggestion
    name: "Tests [Node.js ${{ matrix.node-version-type }}, ${{ matrix.runs-on }}]"
```

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,51 @@
...
+
+# head_ref is per PR, so this ensures that updating the latest PR commit
+# will cancel the previous run of the workflow and trigger a new one
+concurrency:
+  group: "build-and-test-${{ github.head_ref }}"
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash
```
huntie, line 14: Q: Do we have any compatibility issues with zsh (assuming as default)? I'd imagine everything we'd need would work, since zsh is a superset of Bash.

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,51 @@
...
+
+jobs:
+  run-js-checks:
+    runs-on: ubuntu-latest
+    name: "Typescript, Lint, Smoke Test"
```
huntie, line 19: nit: This is _Flow_, TypeScript, Lint ... — can we genericise?
```suggestion
    name: "Static analysis"
```

or "Lint", "Run JS checks"

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,51 @@
...
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/yarn-install
+      - run: yarn typecheck
+      - run: yarn typecheck-ts
+      - run: yarn lint
+      - run: yarn test-smoke
+
+  test-with-coverage:
+    runs-on: ubuntu-latest
+    name: "Test Coverage Calculation"
```
huntie, line 30: nit (all tests need to run to calculate coverage)
```suggestion
    name: "Tests with coverage"
```

.github/actions/yarn-install/action.yml
```diff
@@ -0,0 +1,27 @@
...
+    type: choice
+    options:
+      - 'current'
+      - 'min-supported'
+  min-supported-node-version:
+    required: false
+    default: '18.12'
+  current-node-version:
+    required: false
+    default: '20.2'
```
huntie, line 15: I reckon we can reduce to `18.x`, `20.x` as specifiers at the GitHub level so we can simplify this. Then we could receive one `node-version` arg as a string, and more easily test 3 versions in future if we need to.

.github/workflows/build-and-test.yml
robhogan: This used to be `Type check, lint, smoke test`, which is probably more accurate tbh. But yeah, it might be cleaner if we separated the smoke test from static analysis.

.github/workflows/build-and-test.yml
robhogan: I'm in favour of eliminating `node-version-type` and just using a `node-version` matrix, which still supports aliases like `lts/*`, `lts/-n`, `current`.

.github/actions/yarn-install/action.yml
robhogan: I would like to have a test on `18.0` though since we claim to support it, and there are some differences vs later minors. `['current', 'lts/*', 'lts/-1', '18.0']` (which resolves to 21.x, 20.x, 18.x, 18.0), IMO.

.github/actions/yarn-install/action.yml
vzaidman: It certainly can be done but I wanted to avoid the different workflows using the action (true there's only one at this point) from passing their own `node-version` to avoid setting it separately in different files.

.github/workflows/build-and-test.yml
vzaidman: Unfortunately zsh is not installed on ubuntu images (and installing it is +20sec to each run)
We can use macOS machines instead, but even there zsh is not "officially supported" but only supported as a "custom shell" running a template: https://github.com/actions/runner-images/issues/264 
I still tested it to ensure it doesn't work out of the box, and it indeed doesn't:
https://github.com/facebook/metro/actions/runs/10286164783

.github/actions/yarn-install/action.yml
vzaidman: Sounds good to me, however @huntie disagreed that it makes sense to test the minimum and lts versions before.
WDYT @huntie?

vzaidman: BTW @robhogan, `current` resolves to `v22.6.0`.
https://github.com/facebook/metro/actions/runs/10286511772/job/28467353526

robhogan: Yeah that's fair - essentially it's unstable (though fwiw these CI signals aren't land blocking yet). I'd like to test 18.0 (which should be stable), I'd be OK to use `22.x`, `20.x` instead of the aliases.

It *would* still be good to know asap if we're incompatible with a new Node major, because we don't put a "maximum" constraint on `engines` so Node releases can break users (it's [happened before](https://github.com/facebook/metro/pull/752)), but ideally we'd run that as a separate scheduled (eg nightly) run rather than anything that potentially signals a diff-time failure unrelated to the diff. We can follow up with that.

vzaidman: OK. switched to `['18.0', '18.x', '20.x']` and added to my TODO list to introduce a nightly job that tests `lts/*` and `current`.

robhogan:
@vzaidman LGTM, thanks for iterating - if you want to import I *think* the diff should now be auto-stamped because I've approved it here (never actually tried that tbh!).

.github/actions/yarn-install/action.yml
robhogan: One more for the nightly TODO list then - we should have a test that doesn't use (maybe deletes) `yarn.lock` and pulls the latest versions of all our dependencies, so we can see if a dependency update breaks us somehow.

vzaidman: np

vzaidman: @huntie please approve as well

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,51 @@
...
+      - uses: ./.github/actions/yarn-install
+      - run: yarn typecheck
+      - run: yarn typecheck-ts
+      - run: yarn lint
+      - run: yarn test-smoke
+
+  test-with-coverage:
+    runs-on: ubuntu-latest
+    name: "Tests with coverage"
+    steps:
```
huntie, line 31: Wait, this indentation looks wrong. Can we triple-check this, and enable Prettier on YAML files in our repo?

vzaidman: We already have validation on yml.

Both idents are valid in yml apparently: https://stackoverflow.com/questions/17014460/yaml-indentation-for-array-in-hash
Also, evidently it works.

However, I'll make sure they are formatted the same across the file regardless. Thx for noticing.

huntie: Yes, but can we please enable auto-formatting? 🥹 Good to go 👍🏻

PR #1311, opened 2024-08-05
Replace `rimraf` package with `fs.rmSync`

kitten:
## Summary

This is for the deprecation message:
> Rimraf versions prior to v4 are no longer supported deprecation warning.

The `rimraf` package seems to be largely unused in metro, except for `FileStore#_removeDirs`, where it can be replaced with `fs.rmSync`, [which has been added in Node 14.14/16.14/17.3](https://nodejs.org/api/fs.html#fsrmsyncpath-options). Since the minimum supported Node version (via `engines`) is set to `>=18` this seems fine to use.

A `metro-memory-fs` addition of `rmSync` has been added. A comment has been left to note that it's not the most accurate mock method. I'm not sure what the standard of additons to this package are.

Changelog: [Fix] Remove use of `rimraf` and replace it with `fs.rmSync`

## Test plan

- `rmSync` has been added to `metro-memory-fs` and tests have been added for it
- Existing tests should cover changes to `FileStore`, since functionality is unchanged

### DIFF
--- a/packages/metro-cache/package.json
+++ b/packages/metro-cache/package.json
@@ -14,8 +14,7 @@
   "dependencies": {
     "exponential-backoff": "^3.1.1",
     "flow-enums-runtime": "^0.0.6",
-    "metro-core": "0.80.9",
-    "rimraf": "^3.0.2"
+    "metro-core": "0.80.9"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


robhogan (approved):
Perfect, thanks @kitten!

PR #1323, opened 2024-08-15
Update pipelines to allow running "yarn" on node 18

vzaidman:
Summary:
While we support node 18, `typescript-eslint/eslint-plugin` can only be installed on node version `18.18` and above.

This diff forces `yarn install` regardless of packages' `engines` field so we can run tests on the minimum supported version for `metro`.

Differential Revision: D61336394


robhogan (approved):
👍

PR #1327, opened 2024-08-20
Update node-fetch

vzaidman:
Summary:
Updating `node-fetch` to the latest version caused tests in `packages/metro/src/integration_tests/__tests__/server-test.js` to fail. Running each test case inside the file failed individually, but there seems to have been a race condition between the tests causing subsequent tests to fail with the following error:
```
FetchError: request to http://localhost:10028/import-export/index.bundle?platform=ios&dev=true&minify=false failed, reason: socket hang up
```
It happens because the "connection: close" header is removed in the latest version of `node-fetch`: https://github.com/node-fetch/node-fetch/pull/1765 and can be fixed by adding this header manually to fetches.
```
fetch('path', {headers: {'Connection': 'close'}});
```

This diff introduces a workaround where we add that header, however fetch is expected to work even without it when the following bug is fixed: https://github.com/nodejs/node/issues/54484

Also, since node-fetch was only used in this one test, it was actually a good opportunity to remove it from the project's dependencies altogether.

Differential Revision: D61336391

### DIFF
--- a/packages/metro/package.json
+++ b/packages/metro/package.json
@@ -48,7 +48,6 @@
     "metro-transform-plugins": "0.80.10",
     "metro-transform-worker": "0.80.10",
     "mime-types": "^2.1.27",
-    "node-fetch": "^2.2.0",
     "nullthrows": "^1.1.1",
     "serialize-error": "^2.1.0",
     "source-map": "^0.5.6",
@@ -70,6 +69,7 @@
     "metro-memory-fs": "0.80.10",
     "mock-req": "^0.2.0",
     "mock-res": "^0.6.0",
+    "node-fetch": "^2.7.0",
     "stack-trace": "^0.0.10"
   },
   "license": "MIT",
--- a/packages/metro/src/integration_tests/__tests__/server-test.js
+++ b/packages/metro/src/integration_tests/__tests__/server-test.js
@@ -20,13 +20,20 @@ jest.unmock('cosmiconfig');
 
 jest.setTimeout(60 * 1000);
 
+// Workaround for https://github.com/nodejs/node/issues/54484:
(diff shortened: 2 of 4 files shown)


robhogan:
Interesting, feels like a Node bug to me...

```js
const {createServer} = require('node:http');

const port = 8080;
const url = 'http://localhost:' + port

async function main() {
    const server1 = createServer((req, res) => res.end());
    await new Promise(r => server1.listen(port, r));
    await fetch(url);
    await new Promise(r => setTimeout(r, 0));
    await new Promise(r => server1.close(r));

    const server2 = createServer((req, res) => res.end());
    await new Promise(r => server2.listen(port, r));
    await fetch(url); // Throws
    await new Promise(r => server2.close(r));
}

main();
```

```
node:internal/deps/undici/undici:13193
      Error.captureStackTrace(err);
            ^

TypeError: fetch failed
    at node:internal/deps/undici/undici:13193:13
    at async main (.../fetch-test.js:15:5) {
  [cause]: SocketError: other side closed
      at Socket.<anonymous> (node:internal/deps/undici/undici:6025:28)
      at Socket.emit (node:events:532:35)
      at endReadableNT (node:internal/streams/readable:1696:12)
      at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
    code: 'UND_ERR_SOCKET',
    socket: {
      localAddress: '::1',
      localPort: 61899,
      remoteAddress: undefined,
      remotePort: undefined,
      remoteFamily: undefined,
      timeout: undefined,
      bytesWritten: 338,
      bytesRead: 122
    }
  }
}

Node.js v22.3.0
``

robhogan:
Opened https://github.com/nodejs/node/issues/54484

PR #1340, opened 2024-08-30
re-implement babel traverse cache pollution bug workaround using a second copy of babel traverse

vzaidman:
Summary:
### The issue
Traverse in babel 7 is currently polluting the cache with nodes that have no `hub` entry (see https://github.com/babel/babel/issues/6437). Since the traverse cache is used by other babel operations that expect `hub` to be present, this breaks the code in all kind of unexpected scenarios.
While this issue is fixed in babel 8 (not released at this moment), it's still an issue in the latest version of babel 7.

### The previous workaround
We implemented a workaround for it in https://github.com/facebook/metro/pull/854#issuecomment-1336499395 however in the latest version of `babel/traverse`, that workaround cannot be used anymore (more about that below).

### The new workaround
Instead, we are implementing a different workaround that installs traverse twice, including installing it's cache file twice. We use the second copy of traverse `babel/traverse--for-generate-function-map` only in `forEachMapping`, allowing the rest of the system to use the traverse caching without the pollution issue mentioned above.

### Why the previous workaround stopped working
Due to the use of a `let export` in the latest version of traverse cache, and how it's used in the latest version, we can't re-write `traverse.cache.path` anymore.
This cache is exported with a `let export`:
https://github.com/babel/babel/blob/5ebab544af2f1c6fc6abdaae6f4e5426975c9a16/packages/babel-traverse/src/cache.ts#L6-L20
And it compiles to:
```
let pathsCache = exports.path = new WeakMap();
function clearPath() {
  exports.path = pathsCache = new WeakMap();
}
```
and then used like this:
```
function getCachedPaths(hub, parent) {
  // ...
  pathsCache.get(hub);
  // ...
}
```
Which means that re-writing the export like we used to do breaks the traverse cache because `exports.path` is re-written, but not `pathsCache` while the latter is used inside the file:
```
const previousCache = traverse.cache.path;
  traverse.cache.clearPath();
  traverse(ast, { /* settings */ });
  // this line is breaking the traverse cache
  traverse.cache.path = previousCache;
```

Differential Revision: D61917782

### DIFF
--- a/packages/metro-source-map/package.json
+++ b/packages/metro-source-map/package.json
@@ -13,6 +13,7 @@
   },
   "dependencies": {
     "@babel/traverse": "^7.20.0",
+    "@babel/traverse--for-generate-function-map": "npm:@babel/traverse@^7.20.0",
     "@babel/types": "^7.20.0",
     "flow-enums-runtime": "^0.0.6",
     "invariant": "^2.2.4",
--- a/packages/metro-source-map/src/generateFunctionMap.js
+++ b/packages/metro-source-map/src/generateFunctionMap.js
@@ -17,7 +17,8 @@ import type {NodePath} from '@babel/traverse';
 import type {Node} from '@babel/types';
 import type {MetroBabelFileMetadata} from 'metro-babel-transformer';
 
-import traverse from '@babel/traverse';
+// $FlowFixMe[cannot-resolve-module] - resolves to @babel/traverse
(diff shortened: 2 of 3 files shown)


vzaidman:
Merged as https://github.com/facebook/metro/pull/1343/commits/e9b7d663ce79616820650fe161d4f9f0176677cc

PR #1343, opened 2024-09-02
update babel and fix tests accordingly

vzaidman:
Updated babel packages in `package.json`, removed `yarn.lock` files and recreated them by running `yarn install` also fixed the following:
* Updated jest snapshot tests failing
* Updated babel types and corrected typings accordingly:
  - The latest babel 7 introduces the following features we need to adjust our types to:
    * `JSXNamespacedName` is removed from valid `CallExpression` args ([PR](https://github.com/babel/babel/pull/16421))
      * `JSXNamespacedName` is used for namespaced XML properties in things like `<div namespace:name="value">`, but `fn(namespace:name)` doesn't make any sense.
    * Dynamic imports are enabled behind a new flag `createImportExpressions` ([PR](https://github.com/babel/babel/pull/15682)), introducing calls such as `import(foo, options)`. These complicate the expected values passed to `import` to be more than just strings.
      * Since these are behind a flag that is not expected to be enabled, we can throw an error for now and whoever uses it can add a support to it if needed later.

Differential Revision: D61543660


packages/metro-source-map/package.json
```diff
@@ -12,8 +12,9 @@
     "cleanup-release": "test ! -e build && mv src build && mv src.real src"
   },
   "dependencies": {
-    "@babel/traverse": "^7.20.0",
-    "@babel/types": "^7.20.0",
+    "@babel/traverse": "^7.25.3",
+    "@babel/traverse--for-generate-function-map": "npm:@babel/traverse@^7.25.3",
```
jerone, line 16: Why was this dependency with referer created?

jerone: Our security tools are highlighting this as a potential security issue.

vzaidman: @jerone- hey! the PR summary including this information was cut. Thanks for noticing. I've updated the summary to include the reasoning behind installing `@babel/traverse` twice.

The tool flagging it as a potential security issue is concerned about "detected resolved URL for package with a different name" but this is totally expected, so I think you can ignore or mark this concern as resolved.

jerone: Thanks for the update. I have whitelisted this change.

The tool that giving the warning was lockfile linting: https://github.com/lirantal/lockfile-lint/tree/main/packages/lockfile-lint-api#validators

PR #1366, opened 2024-10-06
Replace strip-ansi with native module

stianjensen:
stripVTControlCharacters from Node core utilities provide the same functionality, and is available since Node 16.11. This project currently requires minimum Node 18.18.

See https://github.com/es-tooling/ecosystem-cleanup/issues/122 for similar efforts to replace this dependency in other projects.

### DIFF
--- a/packages/metro-config/package.json
+++ b/packages/metro-config/package.json
@@ -25,8 +25,7 @@
   "devDependencies": {
     "@types/connect": "^3.4.35",
     "metro-babel-register": "0.80.11",
-    "pretty-format": "^29.7.0",
-    "strip-ansi": "^6.0.0"
+    "pretty-format": "^29.7.0"
   },
   "engines": {
     "node": ">=18.18"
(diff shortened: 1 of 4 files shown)


robhogan:
Thanks for this @stianjensen! Always nice to remove a dependency.

robhogan (approved):

stianjensen:
I'm not so familiar with flow, so I'm not quite sure where the typings for util that are failing now come from.

robhogan:
They're mostly compiled in to Flow, and can fall behind for newer APIs. 

You can locally suppress with `// $FlowIssue`, but I'm happy to take care of that when I import.

stianjensen:
I also managed to get type support merged into flow, but ignoring the types for now probably also works:
https://github.com/facebook/flow/commit/19ca1f251e17cf7230ea8fee365d085d57c4e938

I'll leave it up to you!

PR #1373, opened 2024-10-15
fix or filter with a comment windows tests and run CI tests on windows and macOS

vzaidman:
## Summary
* Make tests pass on by fixing 2 tests files and ignoring 27 test files categorized into 4 categories of reasons for their failure.
* This allows us to run on CI **1308 tests on across 96 files on Windows** ignoring around 700 tests (**64% out of the 2038 tests that pass on macOS**)
* Run tests on MacOS as well, meaning that the tests will now run on Ubuntu, Windows, MacOS

Changelog: [Internal]

## Test plan
`yarn jest` on Windows.
image

CI passes, testing on Windows, Linux, and MacOS:
* PR: https://github.com/facebook/metro/actions/runs/11381806744 image
* Nightly: https://github.com/facebook/metro/actions/runs/11381806756 (fails the same way the current nightly tests fail on current node. Please ignore)

### DIFF
--- a/.github/workflows/build-test-and-deploy.yml
+++ b/.github/workflows/build-test-and-deploy.yml
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: ['ubuntu-latest']
+        runs-on: ['ubuntu-latest', 'windows-latest', 'macos-latest']
         node-version: [
           '18.18.0', # minimum supported
           'lts/-1',  # pre-latest lts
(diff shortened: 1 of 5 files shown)


.github/workflows/build-test-and-deploy.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: ['ubuntu-latest']
+        runs-on: ['ubuntu-latest', 'windows-latest', 'macos-latest']
```
vzaidman, line 49: should we also test on macOS?

PR #1378, opened 2024-10-31
feat: optimize getMaxWorkers using os.availableParallelism

screetBloom:
## Summary

Optimize  the maximum number of worker parallelism using os.availableParallelism

`os.cpus().length` returns the number of CPUs in the physical device, which is static. It should be replaced with the maximum number of computations the system can execute simultaneously. In Node.js, this is provided by `os.availableParallelism()`.

Currently, Metro requires Node.js version 18.18 or higher, which includes `os.availableParallelism`.

## Test plan

Green CI

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -143,7 +143,7 @@ If `true`, Metro will use a stable mapping from files to transformer workers, so
 
 Type: `number`
 
-The number of workers to use for parallel processing in Metro. Defaults to approximately half of the number of cores available on the machine, as reported by [`os.availableParallelism()`](https://nodejs.org/api/os.html#availableparallelism).
+The number of workers to use for parallel processing in Metro. Defaults to approximately half of the number of cores available on the machine, as reported by [`os.availableParallelism()`](https://nodejs.org/api/os.html#osavailableparallelism).
 
 :::note
 1. Values exceeding the number of available cores have no effect.
--- a/packages/metro/src/lib/getMaxWorkers.js
+++ b/packages/metro/src/lib/getMaxWorkers.js
@@ -14,8 +14,7 @@
 const os = require('os');
 
 module.exports = (workers: ?number): number => {
-  const cores = os.cpus().length;
   return typeof workers === 'number' && Number.isInteger(workers)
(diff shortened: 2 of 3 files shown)


robhogan (approved):
LGTM, thanks!

(Will need the CLA agreement above before I can merge though)

packages/metro/src/lib/getMaxWorkers.js
```diff
@@ -14,7 +14,7 @@
 const os = require('os');
 
 module.exports = (workers: ?number): number => {
-  const cores = os.cpus().length;
+  const cores = os.availableParallelism();
```
robhogan, line 17:
```suggestion
  // $FlowFixMe[prop-missing] Missing Flow lib def for availableParallelism
  const cores = os.availableParallelism();
```

docs/Configuration.md
```diff
@@ -143,7 +143,7 @@ If `true`, Metro will use a stable mapping from files to transformer workers, so
 
 Type: `number`
 
-The number of workers to use for parallel processing in Metro. Defaults to approximately half of the number of cores available on the machine, as reported by [`os.cpus()`](https://nodejs.org/api/os.html#oscpus).
+The number of workers to use for parallel processing in Metro. Defaults to approximately half of the number of cores available on the machine, as reported by [`os.availableParallelism()`](https://nodejs.org/api/os.html#availableparallelism).
```
robhogan, line 146:
```suggestion
The number of workers to use for parallel processing in Metro. Defaults to approximately half of the number of cores available on the machine, as reported by [`os.availableParallelism()`](https://nodejs.org/api/os.html#osavailableparallelism).
```

screetBloom:
Additional notes:

1. The current implementation of Math.max(1, Math.ceil(cores * (0.5 + 0.5 * Math.exp(-cores * 0.07)) - 1)) uses approximately 50%-60% of the CPU cores, which may not achieve the highest performance in CI deployments.

2. I will try to design a method to set the optimal number of workers based on different CPU architectures in CI deployments soon. Maximize the use of all performance cores of the CI machine.

PR #1390, opened 2024-11-17
Exclude React Compiler runtime from inline requires

gaearon:
## Summary

Similar to https://github.com/facebook/metro/pull/1126. This call is in every component on every render and it is not beneficial to inline it. On the contrary, you want to reduce the indirection there.

Changelog: [Performance] Exclude React Compiler runtime from inline requires

Changelog:

## Test plan

Before:

Screenshot 2024-11-17 at 18 01 25

After:

Screenshot 2024-11-17 at 18 08 05


gaearon:
cc @robhogan

gaearon:
also cc @poteto fyi

robhogan (approved):

robhogan:
Thanks Dan!

PR #1413, opened 2024-12-21
Watcher backends: use Flow-typed object payloads

robhogan:
Summary:
Ensure `'add'` and `'change'` events fired by watcher backends always have associated `metadata` (and that `'delete'` events never do), by typing change event payloads as objects with the `event: 'change' | 'add' | 'delete'` as a discriminator.

(This is in preparation for refactoring out the distinction between `'change'` and `'add'` from backends, which means they don't need to track files and can do much less IO)

Changelog: Internal

Differential Revision: D67287046
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebook/metro/pull/1413).
* #1416
* #1415
* #1414
* __->__ #1413
* #1412

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,6 @@
 [ignore]
 <PROJECT_ROOT>/packages/.*/build/.*
+<PROJECT_ROOT>/\.hg/.*
 # this transient dep bundles tests with their package, which flow attempts to parse
 # and crashes out as the test includes purposely malformed json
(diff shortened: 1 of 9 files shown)


robhogan:
Merged in https://github.com/react/metro/commit/f7343dad4a9d7f4d2e4b71f07b88038dfac0fa36, https://github.com/react/metro/commit/d0e630be2f5a2efdf7f00e6bad53c4183c8c05d9

PR #1429, opened 2025-01-19
Support circular dependencies and export let with experimentalImportSupport

robhogan:
Summary:
Currently, `experimentalImportSupport` transforms named imports such that they're accessed immediately at the top level, eg:

```js
import {foo} from 'bar';
export function getFoo() {
  return foo;
}
```

Becomes
```js
Object.defineProperty(exports, '__esModule', {
  value: true
});
var foo = require('bar').foo;
function getFoo() {
  return foo;
}
exports.getFoo = getFoo;
```

This immediate, top-level assignment of `require('bar').foo` to `foo` problematic for two reasons:
1. In the case of circular dependencies, the module at `'bar'` may not have been fully initialised, so that `foo` might be undefined at this point.
2. In the case where `bar` defines `export let foo = 'something mutable'`, a reassigment of `foo` within `'bar'` at runtime will not be reflected by the importing module.

Differential Revision: D68394514


thespacemanatee:
Hey @robhogan, I wanted to say thanks for this fix and ask if anything is blocking this from merging. My team wants to try the React Compiler on our Expo project, and this is the current blocker for that. Thanks again!

dannyhw:
Hey just wondering if there's any update? With storybook we're seeing issues with enabling the react compiler that seem like they could be related to this.

PR #1430, opened 2025-01-20
[Internal] Remove unnecessary through2 package

talentlessguy:
## Summary

through2 is a legacy dependency and is no longer required. Metro targets Node 18 where Transform streams are stable and do not need a userland package.

Changelog: [Internal] Remove unnecessary through2 package

## Test plan

All tests pass on Node 18:

Screenshot 2025-01-21 at 00 38 31

### DIFF
--- a/packages/metro-transform-plugins/src/addParamsToDefineCall.js
+++ b/packages/metro-transform-plugins/src/addParamsToDefineCall.js
@@ -23,7 +23,7 @@ function addParamsToDefineCall(
 ): string {
   const index = code.lastIndexOf(')');
   const params = paramsToAdd.map(param =>
-    param !== undefined ? JSON.stringify(param) : 'undefined',
+    param != null ? JSON.stringify(param) : 'undefined',
   );
 
   return code.slice(0, index) + ',' + params.join(',') + code.slice(index);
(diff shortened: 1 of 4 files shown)


packages/metro-transform-plugins/src/addParamsToDefineCall.js
```diff
@@ -23,7 +23,7 @@ function addParamsToDefineCall(
 ): string {
   const index = code.lastIndexOf(')');
   const params = paramsToAdd.map(param =>
-    param !== undefined ? JSON.stringify(param) : 'undefined',
+    param != null ? JSON.stringify(param) : 'undefined',
```
vzaidman, line 26: why the change? I think this was meant to explicitly handle `undefined` and not "`null` or `undefined`"

robhogan: I'm guessing the linter suggested it, but it'd be wrong in this case. 

`null` can be serialised to JSON, `undefined` can't, so the distinction was deliberate.
```suggestion
    param !== undefined ? JSON.stringify(param) : 'undefined',
```

talentlessguy: it was auto format by the linter yeah, not my change

robhogan (approved):
Importing to take a closer look and see if we can avoid a couple of those lint warnings, but LGTM, thanks @talentlessguy !

PR #1438, opened 2025-02-14
Replace this with globalThis

exzos28:
I am using Metro to build a Service Worker for my web application (created with Expo). And I get an error because `this` value  is `undefined` [here](https://github.com/facebook/metro/blob/87f717b8f5987827c75c82b3cb390060672628f0/packages/metro/src/lib/getPreludeCode.js#L28). This piece of code is static and is added to every bundle.

## Summary
Provide a standard way of accessing the global `this` value across environments.


robhogan (approved):

robhogan:
Confirmed `globalThis` is also supported on iOS Safari >= 12.2, Hermes, and Android JSC, so this LGTM. Thanks @exzos28

PR #1439, opened 2025-02-16
metro-file-map: Spawn fewer workers for smaller workloads

robhogan:
Summary:
`metro-file-map` currently spins up `maxWorkers` workers (processes, or threads if `enableWorkerThreads`) for any batch workload. `maxWorkers` is `os.availableParallelism` by default.

When we have a warm cache or Saved State, we may only have a handful of changed files to process, and frequently create more workers than we have files.

This implements a simple heuristic factor `maxFilesPerWorker`, such that we use `Math.ceil(numFiles/maxFilesPerWorker)` parallelism (where 1 "worker" means we process in-band).

The default value of  100 is somewhat finger-in-air as a bunch of factors would determine the optimal value - including threads or not, which host platform, and computational cost of the haste impl and dependency extraction. We could expose this as Metro configuration in future.

Changelog:
 - **[Performance]** Don't start an excessive number of workers for hashing files during startup.

Differential Revision: D69704168

### DIFF
--- a/packages/metro-file-map/src/__tests__/index-test.js
+++ b/packages/metro-file-map/src/__tests__/index-test.js
@@ -1371,9 +1371,18 @@ describe('FileMap', () => {
       dependencyExtractor,
       hasteImplModulePath: undefined,
       maxWorkers: 4,
+      maxFilesPerWorker: 2,
     }).build();
 
-    expect(jestWorker.mock.calls.length).toBe(1);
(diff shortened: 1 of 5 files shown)


robhogan:
Closed by https://github.com/facebook/metro/commit/15bc295808e53da27c140e248184e1bf0fd041c6

PR #1465, opened 2025-03-19
Update docs for "package exports" to promote deferring platform specific requires

kraenhansen:
## Summary

This is a suggestion to update the docs on "PackageExports" to ensure only one platform specific file is ever required.
This aligns with the React Native docs that are linked to: https://reactnative.dev/docs/platform#select
Without this, the source-code for both platforms will be required, which is not ideal and not something we'd want to promote.

Changelog: [Internal] Updating docs of "package exports".

## Test plan

None.

### DIFF
--- a/docs/PackageExports.md
+++ b/docs/PackageExports.md
@@ -191,9 +191,9 @@ We chose not to introduce `"android"` and `"ios"` conditions, due to the prevale
 // src/FooComponent.js
 
 const FooComponent = Platform.select({
-  android: require('./FooComponentAndroid.js'),
-  ios: require('FooComponentIOS.js'),
-});
+  android: () => require('./FooComponentAndroid.js'),
+  ios: () => require('FooComponentIOS.js'),
+})();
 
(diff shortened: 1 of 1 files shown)


robhogan:
> Without this, the source-code for both platforms will be required, which is not ideal and not something we'd want to promote.

Hmm - if that's the case it's a bug. Metro (specifically the `inline-plugin`) should collapse `Platform.select` fully before dependencies are collected for bundling.

So the two are functionally equivalent, but the version in the RN docs collapses to an IIFE for no particular reason afaict.

kraenhansen:
Oh okay, I didn't know the inline-plugin was that smart - that's good to know 👍 (and also a bit magical ✨)

This just stood out to me as a bug when compared to the RN docs 👍

robhogan:
That's fair - without special handling of `Platform.select`, the arrow version would be better because although both modules would be bundled, only the matching one would be evaluated.

Through Metro, the non-matching keys are eliminated at build time.

PR #1476, opened 2025-03-28
Add support for ArrayBuffer in asDeserializedBuffer

camero2734:
## Summary

Running `expo` + `metro` in Bun gives an error like:
```
SyntaxError: src/app?ctx=e5807cc4e93b7ea02120ad0165264d893ac772db: ENOENT: no such file or directory, open '/usr/dev/apps/expo/src/app?ctx=e5807c
```

I drilled down quite a bit and it turns out that the `transform` function _is_ being called with a `fileBuffer` already, so it shouldn't be trying to read the file. The issue is that, in Bun, this seems to be returned as a `Uint8Array` instead of a `Buffer`

So, I've just added a simple check and conversion for ArrayView -> Buffer

## Testing
Here's how to reproduce:
```sh
bun create tamagui@latest --template expo-router
# Name app "bun-test" -- at this point it will try to yarn install and might fail, but we want to use bun
cd bun-test
rm yarn.lock && bun install
bunx --bun expo export --platform web
```

This fails before the change and succeeds after.

## Root cause
It seems to be more specifically that Bun and Node give different results when using `jest-worker` to send a Buffer:
```ts
// test.ts
import { resolve } from "node:path";
import { Worker as JestWorker } from "jest-worker";

const path = resolve(import.meta.dirname, "worker.ts");
async function main() {
  const worker = new JestWorker(path, { enableWorkerThreads: false });
  const result = await worker.giveMeABuffer();
  console.log(result);
}

main();

// worker.ts
export function giveMeABuffer() {
  return Buffer.from("hello");
}
```

```bash
> node --experimental-strip-types test.ts
<Buffer 77 6f 72 6c 64>

> bun run test.ts
Uint8Array(5) [ 119, 111, 114, 108, 100 ]
```

Interestingly, if I set `enableWorkerThreads` to `true`, node _also_ returns a Uint8Array:
```bash
// with { enableWorkerThreads: true }
> node --experimental-strip-types test.ts
Uint8Array(5) [ 119, 111, 114, 108, 100 ]

// with { enableWorkerThreads: true }, but Bun ends up breaking:
> bun run test.ts
NotImplementedError: worker_threads.Worker option "resourceLimits" is not yet implemented in Bun.
 code: "ERR_NOT_IMPLEMENTED"
```

### DIFF
--- a/packages/metro/src/DeltaBundler/Worker.flow.js
+++ b/packages/metro/src/DeltaBundler/Worker.flow.js
@@ -66,6 +66,9 @@ function asDeserializedBuffer(value: any): Buffer | null {
   if (value && value.type === 'Buffer') {
     return Buffer.from(value.data);
   }
+  if (ArrayBuffer.isView(value)) {
+    return Buffer.from(value.buffer, value.byteOffset, value.byteLength);
+  }
   return null;
 }
 


robhogan:
Thanks for the deep dive and fix here :)

> Interestingly, if I set enableWorkerThreads to true, node also returns a Uint8Array:

Does that mean that context modules are broken under Node with `transformer.unstable_enableWorkerThreads`?

camero2734:
@robhogan Yep, seems like it 🙃 

Added `config.transformer.unstable_workerThreads = true;` to the tamagui repro above and ran it with node:

image

PR #1482, opened 2025-04-05
Use in-order scheduling policy for transform workloads (prefer hot workers)

robhogan:
Summary:
The `in-order` scheduling policy was added to `jest-worker` by Micha Reiser in https://github.com/jestjs/jest/pull/10902 , specifically with Metro's use case in mind, but it wasn't actually released until after Micha left Meta, so never made it into Metro.

This alters the scheduling policy so that an incoming job (a file transform) will be allocated to the lowest-numbered available worker, rather than the next higher numbered available worker.

This is more likely to reuse "hot" workers, especially in cases where the workload is small relative to the number of cores - including at the beginning of a typical bundle where we're fanning out from a single entry point.

Also updates Flow types based on latest from [upstream TypeScript](https://github.com/jestjs/jest/blob/main/packages/jest-worker/src/types.ts).

Changelog
```
 - **[Performance]**: Prefer scheduling transforms on available hot workers via in-order scheduling policy
```

Reviewed By: yungsters

Differential Revision: D72391272


MichaReiser:
Oh, I forgot about this. Nice to see this landing. Thanks @robhogan I hope it gives you some performance gains.

PR #1483, opened 2025-04-07
feat(metro): Improve stack frame handling by showing the first valid frame when all frames are collapsed

EvanBacon:
## Summary

When an error comes from the internals of your app it can be especially devastating. This generally happens when you have a versioning issue from upgrading. The stacks are currently optimized for user-space errors which make up the majority of issues, but when all frames are collapsed we should assume the issue is an internal bug and show anything to give the user some sort of hint for what to fix.

Changelog: [Feature] Fallback to showing first collapsed frame when all frames in a stack are collapsed.


robhogan:
Looks like the tests need fixing but the change LGTM, thanks!

robhogan (approved):

PR #1494, opened 2025-04-27
[rfc] Support import.meta with experimentalImportSupport

robhogan:
Currently, Metro leaves `import.meta` accesses untransformed, but provides no `import` pseudoglobal, so (unless polyfilled) this throws at runtime.

`import.meta` is [specified](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-meta-properties) as a writable null-proto object. Though the spec is silent on any properties, it provides a mechanism ([`HostGetImportMetaProperties`](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-hostgetimportmetaproperties)) for hosts to set properties, given a module object. Eg: `url`, `resolve` are defined by Node.js and browsers (and notably `env` by Vite).

We emulate that mechanism here with the `__getImportMetaProperties(module) => {[key: string]: mixed}` optional, framework-defined global function.

### DIFF
--- a/packages/metro/src/integration_tests/__tests__/import-export-test.js
+++ b/packages/metro/src/integration_tests/__tests__/import-export-test.js
@@ -26,7 +26,14 @@ test('builds a simple bundle', async () => {
     entry: 'import-export/index.js',
   });
 
-  const object = execBundle(result.code);
+  const object = execBundle(result.code, {
+    // Framework/host-defined props
+    __getImportMetaProperties: module => {
+      return {
+        url: new URL(String(module.id), 'metro://'),
+      };
+    },
+  });
   const cjs = await object.asyncImportCJS;
 
   expect(object).toMatchSnapshot();
--- a/packages/metro-runtime/src/polyfills/require.js
+++ b/packages/metro-runtime/src/polyfills/require.js
@@ -467,6 +467,22 @@ function loadModuleImplementation(
     }
     moduleObject.id = moduleId;
 
+    if (!moduleObject.importMeta) {
+      let importMeta;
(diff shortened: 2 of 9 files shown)


satya164:
What do you think about adding `import.meta.resolve` in addition to `import.meta.url`?

It's supported in both browsers and node, and would be a replacement for `require.resolve`.

robhogan:
> What do you think about adding `import.meta.resolve` in addition to `import.meta.url`?
> 
> It's supported in both browsers and node, and would be a replacement for `require.resolve`.

The plan is for RN (or Expo) to define properties using the `__getImportMetaProperties` hook, which is analogous to host-defined props in the ES spec - Metro doesn't define `url` either (it does here only in an integration test).

packages/metro/src/integration_tests/__tests__/import-export-test.js
```diff
@@ -26,7 +26,14 @@ test('builds a simple bundle', async () => {
     entry: 'import-export/index.js',
   });
 
-  const object = execBundle(result.code);
+  const object = execBundle(result.code, {
+    // Framework/host-defined props
+    __getImportMetaProperties: module => {
+      return {
+        url: new URL(String(module.id), 'metro://'),
+      };
```
satya164, line 34: so this code will be added to react native itself?

satya164:
> The plan is for RN (or Expo)

I feel like for library authors it'd be better if it's added in RN itself, coz otherwise it'd be hard to rely on a consistent `import.meta` for react native, but ofc react native could always define a base `import.meta` and Expo could add Expo specific things on top of it.

IgnisDa:
any progress on this?

Issue #1493, opened 2025-04-23
`_interopRequireDefault is not a function` when loading local file


gfreeman-karmak:
Metro 0.82.1
React Native 079.1

We have a React Native app, and one of the screens is loading a custom component found within a subfolder. So in the screen:
`import { WarningModal } from './components';`
which points to `./components/index.ts`:
`export { default as WarningModal } from './warning_modal';`

When arriving at that screen in the app we get these errors:

![Image](https://github.com/user-attachments/assets/7b0d2f55-f878-41d7-b3a1-fd08faf7be8f)
![Image](https://github.com/user-attachments/assets/5eaf9d20-5f9c-4a40-bcfc-8f0dc1c35077)

This has worked perfectly fine until we upgraded to RN 0.79.1, which in turn put us on Metro 0.82.1, which now has package export support turned on by default. We never had `unstable_enablePackageExports: true` previously, and when I explicitly set that to false, the import works again. It's odd because this is just importing a local file, and in fact if we change the import to bypass `index.ts` with `import WarningModal from './components/warning_modal', that works too.

Possibly related to #1464 but I think this is a different issue.

robhogan:
Hi @gfreeman-karmak, what's your Metro config and `@react-native/metro-config` version?

gfreeman-karmak:
`@react-native/metro-config` is 0.79.1. Metro config file is below:
```
const { getDefaultConfig, mergeConfig } = require('@react-native/metro-config');

/**
 * Metro configuration
 * https://reactnative.dev/docs/metro
 *
 * @type {import('@react-native/metro-config').MetroConfig}
 */

const defaultConfig = getDefaultConfig(__dirname);

const config = {
    resolver: {
        assetExts: [...defaultConfig.resolver.assetExts, 'lottie'],
        unstable_enablePackageExports: false
    }
};

module.exports = mergeConfig(defaultConfig, config);
```

removing `unstable_enablePackageExports: false` is when we have our issue

mjmasn:
[...]

Issue #1478, opened 2025-03-31
Inconsistent (and broken) bundle output since 0.81.2 - even with identical source code


mjmasn:
Following on from my comment here: https://github.com/reactwg/react-native-releases/issues/856#issuecomment-2757877684

I've been trying to reproduce this issue reliably - no luck so far on my machine but weirdly every build from Bitrise CI is broken on Android. Not really sure what would be different there but I have tried to compare the resulting index.android.bundle files between bitrise and local release build. 

I used `hbc-decompiler` from here: https://github.com/P1sec/hermes-dec to get back to sorta-js from hermes bytecode. i.e. `hbc-decompiler index.android.bundle index.js`

The entire diff is massive because a lot of module ids have changed, presumably due to these issues, and I can't share our app code but I'll paste an interesting example snippet here - seems to be import/require related so hoping it sparks an 'aha' in someone who knows this stuff a bit better than I do.

<table>
<thead>
<tr>
<th>Not Working</th><th>Working</th>
</tr>
</thead>
<tbody>
<tr>
<td>

```js
case 34:
                r4 = r4.Object;
                r5 = r4.assign;
                r4 = r5.bind;
                r2 = r4.bind(r5)();
case 54:
                _closure1_slot1 = r2;
                r3['exports'] = r2;
                r2 = _closure1_slot0;
                r4 = r2.exports;
                r3 = true;
                r4['__esModule'] = r3;
                r3 = r2.exports;
                r2 = r2.exports;
                r3['default'] = r2;
                r3 = _closure1_slot1;
                r2 = r3.apply;
                r0 = arguments;
                r1 = r0;
                r0 = null;
                r0 = r2.bind(r3)(r0, r1);
                return r0;
            }
        };
        var _closure1_slot1 = r1;
        r0['exports'] = r1;
        r2 = r0.exports;
        r1 = true;
        r2['__esModule'] = r1;
        r1 = r0.exports;
        r0 = r0.exports;
        r1['default'] = r0;
        r0 = undefined;
        return r0;
    };
    r6 = 1713;
    r5 = new Array(0);
    r5 = r8.bind(r2)(r7, r6, r5);
    r8 = r0.__d;
```
</td>
<td>

```js
case 34:
                r4 = r4.Object;
                r5 = r4.assign;
                r4 = r5.bind;
                r2 = r4.bind(r5)();
case 54:
                _closure1_slot1 = r2;
                r3['default'] = r2;
                r3 = _closure1_slot1;
[...]

PR #1497, opened 2025-05-03
fix(metro): create a new set for every `importLocationPlugin` file

byCedric:
## Summary

I came across the `file.unstable_importDeclarationLocs` property when investigating an anomaly with the `isESMImport` detection with [`rc-pagination/es/index.js`](https://app.unpkg.com/rc-pagination@5.1.0/files/es/index.js) - used in [`antd` and reported to not-work with `package.json:exports`](https://github.com/expo/expo/discussions/36551#discussioncomment-13015890).

When investigating, I noticed that this file seemingly had 1351 import locations, which ofc is incorrect. Turns out, we don't recreate the import location set when collecting new dependencies - meaning the set contained the import locations of _all files collected_.

image

This creates a new set every time the import location plugin is being expected - it creates a new set in `Pre`, before running any of the visitors.

Changelog: [Fix] Create new import locations set for every individual files

## Test plan

Hard to test, you could add logging or breakpoints [right after this line](https://github.com/facebook/metro/blob/1d90d082c6049ed0712d89dc03223278b608de23/packages/metro-transform-worker/src/index.js#L389). Here you can spot if the location set is going over reasonable sizes.

### DIFF
--- a/packages/metro/src/ModuleGraph/worker/__tests__/importLocationsPlugin-test.js
+++ b/packages/metro/src/ModuleGraph/worker/__tests__/importLocationsPlugin-test.js
@@ -27,6 +27,14 @@ function parse(code: string) {
   }) as BabelNodeFile;
 }
 
+function transformString(code: string) {
+  return transformFromAstSync<MetroBabelFileMetadata>(parse(code), code, {
+    filename: 'file.js',
+    cwd: '/my/root',
+    plugins: [importLocationsPlugin],
+  });
+}
+
 test('gathers source locs of static ESM imports and re-exports', () => {
   const code = `
   // ESM imports that will be transformed by any ESM->CJS transform, we must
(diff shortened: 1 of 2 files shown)


robhogan:
Thanks again for this one - this is really insidious, since on a sufficiently large build you're very likely to have an ESM import in some file somewhere on at least the first 10 lines, and with several different line lengths for each. 

Then any subsequently transformed CJS `require` could be falsely flagged as ESM if it happens to have a matching line number and length as some other `import`. Hence the resolution would be dependent on order of (async) traversal and what's retrieved from the cache, which makes the build (and even the _graph_) non-deterministic. Which is.. bad (and on me - sorry everyone!).

I'll see if we can prove that with a regression test and this will definitely be a candidate for hotfixing back to 0.81 ASAP.

On the face of it there's a good chance this explains + fixes:
 - https://github.com/facebook/metro/issues/1478
 - https://github.com/facebook/metro/issues/1493
 
 CC @mjesun @gfreeman-karmak

mjmasn:
Thanks @byCedric + @robhogan, very glad you were able to identify the issue, running bundler repeatedly now gives identical output each time for us as it did in previous metro versions.

PR #1498, opened 2025-05-03
fix(metro): avoid using `Program` location when looking up ESM imports

byCedric:
This is a backstream port of https://github.com/expo/expo/pull/36594

## Summary

We noticed a couple of interesting resolution failures due to `@babel/runtime/helpers/interopRequireDefault` being incorrectly resolved as ESM import. It happens for the following two cases:
- https://app.unpkg.com/rc-pagination@5.1.0/files/es/index.js
- https://app.unpkg.com/@redux-devtools/ui@1.4.0/files/lib/esm/Button/index.js

Turns out that the `getNearestLocFromPath` - used in the dependencies collection - traverses the babel AST until it finds a location. Unfortunately, when the `Program` is literally a single line (`export { default } from './x'`), it means that all imports will be resolved as that single line.

It conflates generated `require('@babel/runtime/helpers/interopRequireDefault')` lines with this ESM imports, and tries to use the ESM version of babel - which contains `.default`. Resulting in `_interopRequireDefault is not a function` errors.

# TL;DR;
- When collecting imports, we look up the location to see if it's an ESM import
- We use `getNearestLocFromPath` to look up the location of the import
- This iterated up to the `Program` node
- The `Program` node in Babel has a location for the whole program
- If the file or `Program` has just 1 line, that location is mixed up, causing incorrect `isESMImport` results for generated code.

Changelog: [Fix] Avoid using `Program` node location when looking up ESM import location

## Test plan

See added test, and repro: https://github.com/expo/expo/discussions/36551#discussioncomment-13015890

### DIFF
--- a/packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
+++ b/packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
@@ -28,7 +28,10 @@ const {transformFromAstSync} = require('@babel/core');
 const babylon = require('@babel/parser');
 const t = require('@babel/types');
 const dedent = require('dedent');
-const {importLocationsPlugin, locToKey} = require('metro/private/ModuleGraph/worker/importLocationsPlugin');
+const {
+  importLocationsPlugin,
+  locToKey,
+} = require('metro/private/ModuleGraph/worker/importLocationsPlugin');
 const nullthrows = require('nullthrows');
 
 const {any, objectContaining} = expect;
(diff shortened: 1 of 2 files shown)


robhogan (approved):

robhogan:
This is a really good find - huge thanks for the investigation + write up + PR + tests! 

> If the file or Program has just 1 line, that location is mixed up, causing incorrect isESMImport results for generated code.

Just to confirm - this would only affect cases where a single `ImportDeclaration` or `ExportDeclaration` is the entire file, IIUC - since the `loc` of the program node would have to match the import `loc` exactly.

Was there a particular real-world lib that triggered this, maybe we could mention to improve discoverability of the fix?

byCedric:
@robhogan Yes, only single files containing `export { ... } from './..'` are affected by this. But there are quite a few packages that use this pattern. We could manually search for them, but we discovered this through @kudo's tests around `antd` which uses `rc-pagination`.

And when upgrading `@expo/dev-plugins` also @kudo noticed the same behavior for `@redux-devtools/ui`.

There are probably more packages affected by this, but I haven't searched widely on the pattern yet.

PR #1500, opened 2025-05-05
docs: update dead link for inline requires.

Maker-Mark:
## Summary
The link in the current config docs is dead. I updated it! :-)


robhogan (approved):
Thanks @Maker-Mark :)

PR #1508, opened 2025-05-23
Add copyright header to eventemitter3.js.flow

rshest:
Summary: There is a signal from the OSS checkup tool, complaining about a lack of the copyright header in this particular file (see T225155176), the diff adds it to make it happy.

Differential Revision: D75282886


robhogan:
(Diff abandoned because this isn't our copyright, closing)

Issue #1480, opened 2025-03-31
Module Federation Support


zackarychapple:
This issue serves as a central place to track the support of [Module Federation](https://module-federation.io/) in Metro.

Module Federation has become the standard for decomposing large web-based applications. Through Re.Pack and Module Federation, we've seen growing excitement around the potential for federated mobile applications and a standards-based approach to building mini-apps in React Native. For long-term stability, we believe it would be ideal for Module Federation to be supported directly within Metro. This would align Metro with other bundlers that support Module Federation natively, such as webpack, Rspack, and Rolldown, as well as Vite via a community plugin.

Enabling this may require adding new capabilities to Metro and making some adjustments on the Module Federation side. A single tracking issue helps consolidate discussions and progress around these changes.

We’re excited to see this move forward!

cc @thymikee @robhogan @scriptedalchemy @valorkin

martamoc:
@Esemesek

ScriptedAlchemy:
Main areas of consideration.

1) shared modules

when sharing a module, all imports resolve to a "virtual" module who then calls our runtimes loadShare() method

so import "react" => loadShareSync("react")

The system is async, so if the module is not preloaded, in order to make eager module consumption work - we usually make end users have a import() somewhere in the codebase before shared modules are used, then we will hoist all async requirements to that boundary and await the requirements there. 

2) remote modules

follow same construct as shared modules, they are async and we create a virtual module that initializes the containers upfront, then pulls the exposed module from the container as needed or at the async boundary

import 'remote/button' => loadRemote('remote/button')

3) initialization of containers

We loop over the plugin configs to find all remotes associated in the build and build up the configuration, then call init({name, remotes,shared}); this must happen at application startup

banerjeesouvik:
This will be very helpful. In one of our native expo projects, we are trying to create the web build a an MFE, which we will import in another web app. This solves our use case.

zackarychapple:
> We're currently working on a basic prototype to uncover the other requirements, but it seems like we could skip code-splitting altogether for the time being and package everything into the remote container (`remoteEntry.js`).

@ScriptedAlchemy any feedback on this part?

zackarychapple:
Initial prototype has been created and has been uploaded to the module federation organization.

crherman7:
I love that module federation is becoming more of conversation with respect to metro-bundler! Great progress 👌

[...]

PR #1509, opened 2025-05-27
feat: don't override async bundle paths searchParams in prod

jbroma:
## Summary

Part of #1480

When implementing Module Federation for Metro, we're reusing existing lazy bundling for production bundles, and we wanted to clean up the bundlePaths for the production bundles by not adding any query params to it. Since the `searchParams` are entirely based on `sourceUrl`, we can avoid adding `searchParams` by having a clean `sourceUrl`.

This PR does two things:
- [x] - it makes the override of `runModule` & `modulesOnly` params happen only in dev
- [x] - prevents adding `?` to the URL when there are no `searchParams`

Changelog: [Feature] Prevent adding runModule & modulesOnly search params when processing bundle async paths for production bundles

## Test plan

- [x] - existing tests pass
- [x] - new test passes

### DIFF
--- a/packages/metro/src/DeltaBundler/Serializers/helpers/__tests__/js-test.js
+++ b/packages/metro/src/DeltaBundler/Serializers/helpers/__tests__/js-test.js
@@ -198,4 +198,26 @@ describe('wrapModule()', () => {
       `__d(function() { console.log("foo") },0,{"0":1,"1":2,"paths":{"1":"/../bar.bundle?modulesOnly=true&runModule=false"}},"foo.js");`,
     );
   });
+
+  test('async dependency paths do not add query params when dev is false', () => {
+    const dep = nullthrows(myModule.dependencies.get('bar'));
+    myModule.dependencies.set('bar', {
+      ...dep,
+      data: {...dep.data, data: {...dep.data.data, asyncType: 'async'}},
+    });
+    expect(
+      raw(
+        wrapModule(myModule, {
+          createModuleId: createModuleIdFactory(),
+          dev: false,
+          includeAsyncPaths: true,
+          projectRoot: '/root',
+          serverRoot: '/root',
+          sourceUrl: 'http://localhost/Main.bundle?param1=true&param2=1234',
+        }),
+      ),
+    ).toMatchInlineSnapshot(
+      `__d(function() { console.log("foo") },0,{"0":1,"1":2,"paths":{"1":"/../bar.bundle"}});`,
+    );
+  });
 });


jbroma:
@vzaidman taken that from here:
https://github.com/facebook/metro/blob/6f0093c4cdc597dbbde8b22b159b9cb0b7912865/packages/metro/src/DeltaBundler/Serializers/helpers/__tests__/js-test.js#L179

(although that might have been an over-interpretation on my side 😅)

jbroma:
@vzaidman is there anything we can do about this PR to move forward with it? 🙏

robhogan:
@jbroma sorry about the delay with this. 

Could you remind me - is there a functional difference intended here (eg MF using different defaults, so we want to leave these params implicit), or is this purely tidying up?

jbroma:
> @jbroma sorry about the delay with this.
> 
> Could you remind me - is there a functional difference intended here (eg MF using different defaults, so we want to leave these params implicit), or is this purely tidying up?

No worries, I feel like we got a little bit sidetracked and left those PRs be for the past month or so, so sorry about that too  😅

So this PR is mostly about tidying up the bundle output in production mode. This PR is completely optional for Module Federation but it cleans up the split module IDs by removing the query params from them.

This should not be breaking in any way because if the query params in question are already there they will be replaced just like before - and that seems to be the case, even for nested lazily imported bundles.

robhogan:
@jbroma I just had a closer look at this and notice the params we set here are the opposite of the defaults:

https://github.com/facebook/metro/blob/4f2110412e3c4cbe6feb4f4cad27085edf7db1b0/packages/metro/src/lib/parseOptionsFromUrl.js#L72

https://github.com/facebook/metro/blob/4f2110412e3c4cbe6feb4f4cad27085edf7db1b0/packages/metro/src/lib/parseOptionsFromUrl.js#L75

So these params are overrides to Metro defaults, and they're typically necessary when loading additional module definitions into an existing bundle _from the Metro server_ (you don't need the bundle boilerplate hence `modulesOnly: true`, and it's up to the caller to evaluate hence `runModule: false`).

The effect of this PR is that if you don't explicitly specify `modulesOnly=false` (the default) when loading the _entry_ bundle from Metro, then `modulesOnly` won't be overridden to `true` for lazy imports - which potentially breaks lazy bundling from the Metro server.

For your use case, I guess you're loading prebuilt bundles in prod and so the query params are just noise. But I think the more correct approach for that is to manipulate the URL only when loading from the dev server, rather than basing it on whether the particular params are present.

jbroma:
Ah, that makes sense, haven't thought of that implicit edge case. I'll rework this to a dev-only feature then 👍 Thanks for investigating this deeply 🙏

jbroma:
@robhogan I've aligned the PR to prevent overriding async bundle paths in prod only 👍

jbroma:
hey @robhogan 👋 

anything left to do here to move forward with this PR?

PR #1511, opened 2025-06-03
Avoid a second graph build when getting assets

tido64:
## Summary

@dannyvv recently discovered that Metro calls [`buildSubgraph()`](https://github.com/facebook/metro/blob/15fef8ebcf5ae0a13e7f0925a22d4211dde95e02/packages/metro/src/DeltaBundler/buildSubgraph.js#L105) twice during bundling. Once during `build()`: https://github.com/facebook/metro/blob/15fef8ebcf5ae0a13e7f0925a22d4211dde95e02/packages/metro/src/index.flow.js#L431

And again, shortly after, in `getAssets()`: https://github.com/facebook/metro/blob/15fef8ebcf5ae0a13e7f0925a22d4211dde95e02/packages/metro/src/index.flow.js#L435

I *think* the second build should be mostly cached, but it would be better if we can avoid it altogether if possible.

Digging deeper down, it looks like we have the full graph inside `build()`, but throw away the work when returning. I've changed things so that it also returns the graph and made it possible to pass it to `getAssets()` to avoid the second graph build. I don't know if this is the best way to solve this, but it seems to work on my machine™️.

Changelog: [Performance] Avoid a second graph build when getting assets

## Test plan

I've added `console.log` inside `buildSubgraph` to make sure it only gets called once per unique input.

### DIFF
--- a/packages/metro/src/Server.js
+++ b/packages/metro/src/Server.js
@@ -337,9 +337,10 @@ class Server {
         graph,
       }),
       withAssets
-        ? this._getAssetsFromDependencies(graph.dependencies, {
-            platform: bundleOptions.platform,
-          })
+        ? this._getAssetsFromDependencies(
+            graph.dependencies,
+            bundleOptions.platform,
+          )
         : null,
     ]);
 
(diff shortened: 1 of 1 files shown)


robhogan:
Hi @tido64 - thanks for spotting this. Indeed the transform results would normally be cached (where a cache is configured), but resolutions wouldn't, so this is a useful optimisation. 

Is it particularly the use of the `runBuild` API that you're looking at? The reason I ask - I'd prefer to avoid leaking the whole `Graph` from the `shared/output/bundle`, but I think we can tweak things so that it can return `assets` directly (and also benefit from parallelising serialisation and asset enumeration). I'd add a `withAssets: boolean` to `RequestOptions` and encapsulate building assets within `Server.build`.

Does it still work for you if I pull this in and make those changes?

tido64:
> Does it still work for you if I pull this in and make those changes?

Yes, please! As long as we implement the optimization, please feel free to make any changes.

PR #1521, opened 2025-06-22
Fix: platform inline transformation

IvanIhnatsiuk:
## Summary

I noticed that when using esmodule named import of platform dead code elimination of Platform.OS did not work.

In this PR, I added logic to find to check that Platform.OS was used from `require('react-native)`. That way, it doesn't matter what name the react-native import has.

Changelog:

### Before:

```typescript
import { Platform } from 'react-native';

if (Platform.OS === 'android') {
  // code
}
```

#### Transformed:

```typescript
import { Platform } from 'react-native';

var _reactNative = _$$_REQUIRE(_dependencyMap[4]);

if (_reactNative.Platform.OS !== 'android') {
  // code
}
```

### After:

```typescript
import { Platform } from 'react-native';

if (Platform.OS === 'android') {
  // code
}
```

#### Transformed:

```typescript
import { Platform } from 'react-native';

var _reactNative = _$$_REQUIRE(_dependencyMap[4]);

if ('android' !== 'android') {
  // code
}
```

## Test plan

Added a test for the case similar to how code looks in the transformed bundle before being passed to the inline plugin. Also tested this in an app bundle and looked for instances of Platform.OS than should be inlined and made sure there was none.


IvanIhnatsiuk:
@robhogan Sorry for the tag, but I do not know who can I tag to ask about this PR 😄 

Do you have any plans to merge/close this one?

robhogan:
Hi @IvanIhnatsiuk - sorry this got dropped, I think https://github.com/facebook/metro/pull/1641 might've superseded it? That was released in 0.83.4

retyui:
@robhogan I've checked it on the latest ver. of react native (0.84.1)

and issue is still exist  https://github.com/facebook/react-native/pull/56344#issue-4217602180


```
yarn why metro-transform-plugins                                             
├─ metro-transform-worker@npm:0.83.5
│  └─ metro-transform-plugins@npm:0.83.5 (via npm:0.83.5)
│
└─ metro@npm:0.83.5
   └─ metro-transform-plugins@npm:0.83.5 (via npm:0.83.5)
```


Can you merge this PR?

robhogan:
> Can you merge this PR?

@retyui Not as is I'm afraid - it conflicts almost entirely with https://github.com/facebook/metro/pull/1641.

This now needs a closer look to work out why https://github.com/facebook/metro/pull/1641 didn't work as intended, and fix the current implementation.

retyui:
@robhogan I debug current impl. and found the root cause

In general inline-plugin works if we import `Platform` from `react-native` 

---

but some modules in `react-native` codebase use [relative imports](https://github.com/retyui/react-native/blob/fdd6ca5bb11b105284c7ad977ff2208c2f2e66f2/packages/react-native/Libraries/Components/Pressable/useAndroidRippleForView.js#L15 ) for  `Platform`, for example:

```tsx
import Platform from '../../Utilities/Platform'; // <- inline-plugin won't work
```


so inline plugin checks `isPlatformNode` returns `false` as 
this relative imports isn't treated as imported from `react-native` package  


---

As workaround we can use `require`, works well:
```tsx
// 1. source code
const Platform = require('../../Utilities/Platform').default;
export const test = Platform.OS === 'ios' ? 'ios' : 'not-ios';

// 2. code before inline plugin
var Platform = require('../../Utilities/Platform').default;
var test = exports.test = Platform.OS === 'ios' ? 'ios' : 'not-ios';


// isPlatformNode 
//  -> isPlatformOS() 
//       -> isToplevelBinding() // <- true

// this line: https://github.com/thymikee/metro/blob/2fa8a8819d6c88b72fbb414d5c324df29546373e/packages/metro-transform-plugins/src/utils/createInlinePlatformChecks.js#L149
```

---

Code examples for tests:

```tsx
// 1. source code
import Platform from '../../Utilities/Platform';
export const test = Platform.OS === 'ios' ? 'ios' : 'not-ios';

// 2. code before inline plugin
var _Platform = _interopRequireDefault(require("../../Utilities/Platform"));
var test = exports.test = _Platform.default.OS === 'ios' ? 'ios' : 'not-ios';

// ^^^ inline-plugin doesn't work 
```

robhogan:
This is superseded by https://github.com/facebook/react-native/issues/57848 (RN 0.88), which moves the plugin into RN and earlier in the pipeline. Thanks for digging into it and motivating the fix.

PR #1534, opened 2025-07-16
Replace lodash.throttle with es-toolkit's throttle

dayongkr:
## Summary

This pull request replaces the `lodash.throttle` dependency with `es-toolkit/compat/throttle` across the Metro codebase. The motivation for this change is to address several key areas:

1.  **Reduce Bundle Size:** `es-toolkit`'s `throttle` is significantly more lightweight (115 lines vs. 440 lines for `lodash.throttle`). This is achieved by removing unnecessary self-implemented helper functions (like `isObject` and `toNumber`) and simplifying complex internal state management present in the `lodash` version. This directly leads to a smaller overall bundle.
2.  **Improve Performance:** A leaner codebase inherently contributes to better application performance due to reduced parsing and execution overhead.
3.  **Enhance Maintainability:** `lodash.throttle` has not been updated in the past four years, indicating a lack of active maintenance. In contrast, `es-toolkit` is actively maintained, ensuring ongoing support, bug fixes, and performance optimizations.
4.  **Increase Reliability:** `es-toolkit/compat/throttle` has been rigorously tested, passing all `lodash.throttle` tests to ensure 100% compatibility in terms of interface and behavior. It has also been successfully adopted by other prominent libraries like Storybook and Recharts, demonstrating its stability and reliability in real-world scenarios.

This migration allows us to leverage a more modern, performant, and actively maintained throttling utility without introducing any breaking changes or significant migration effort.

Changelog: [Performance] Replace `lodash.throttle` with `es-toolkit/compat/throttle` for reduced bundle size and improved maintainability.

## Test plan

To verify the successful replacement and continued functionality, the following step was taken:

1.  **Run existing Metro test suite:**
    * `yarn test` (or equivalent command to run all Metro unit/integration tests).
    * **Expected outcome:** All tests pass, demonstrating that the new `throttle` implementation maintains the expected behavior and does not introduce regressions.


vzaidman:
Hey! Thanks for your contribution. This would have too little benefits / impact for us to replace the well tested and highly reliable and secure lodash.

raon0211:
Thank you for your thoughtful response. I completely understand your reasoning about stability and avoiding unnecessary churn. However, I'd like to respectfully present a few additional considerations that might change the cost-benefit analysis:

## 1. Bundle impact for users
Even with just 1 function, the size difference is significant:

lodash: ~1.45 kB 
es-toolkit: ~495 bytes 

That's a 66% reduction affecting all projects using metro.

## 2. Zero-risk migration
The migration is literally a drop-in replacement

- import { throttle } from 'lodash';
+ import { throttle } from 'es-toolkit/compat';
Same API, same behavior, zero breaking changes.

## 3. Maintenance reality
While lodash's throttle is stable, the library itself hasn't seen meaningful updates in 2+ years. es-toolkit is actively maintained and already adopted by major projects like Storybook and Recharts for these exact reasons.

I understand if this still feels like unnecessary change, but given the minimal effort required and the collective benefit to your users, would you be open to reconsidering?

Thanks for maintaining such a useful bundler! 🙏

robhogan:
```
$ gdu -sh --apparent-size node_modules/es-toolkit node_modules/lodash node_modules/lodash.throttle
2.5M    node_modules/es-toolkit
1.4M    node_modules/lodash
17K     node_modules/lodash.throttle
 ```

Metro is a Node.js application, so install size would be the only relevant size consideration (though not a particularly important one). `es-toolkit` is actually larger than `lodash`, let alone `lodash.throttle`, which is the package you're proposing to replace. It'd be different if we included `lodash` in bundler output, but we don't.

Besides that, we do tend to be cautious about accepting PRs from folks promoting their own packages, especially when they don't mention their affiliation.

raon0211:
Thank you for the clear feedback and explaining the reasoning! ❤️

Issue #1555, opened 2025-08-13
Call os.hostname on Windows 7 node v18 got error uv_os_gethostname returned ENOSYS (function not implemented)


tony9527167:
Call os.hostname on Windows 7 node v18 got error uv_os_gethostname returned ENOSYS (function not implemented)

Reproduce env & steps: 

- OS Windows 7 64-bit SP1
- node v18.20.8
- create any starter app via Expo, then start it `npx expo start`

It raises error as following

```
NodeError [SystemError]: A system error occurred: uv_os_gethostname returned ENOSYS (function not implemented)
    at Object.<anonymous> (C:\Users\public\demo\node_modules\metro-core\src\Logger.js:7:6)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Module.load (node:internal/modules/cjs/loader:1203:32)
    at Function.Module._load (node:internal/modules/cjs/loader:1019:12)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:177:18)
    at Object.<anonymous> (C:\Users\public\demo\node_modules\metro-core\src\index.js:5:16)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1422:10) {
  code: 'ERR_SYSTEM_ERROR',
  info: {
    errno: -4054,
    code: 'ENOSYS',
    message: 'function not implemented',
    syscall: 'uv_os_gethostname'
  },
  errno: [Getter/Setter: -4054],
  syscall: [Getter/Setter: 'uv_os_gethostname']
}
```

ref: https://github.com/facebook/metro/blob/82f5c61f03986b1bfbe5f55d97186ea1ec68a82b/packages/metro-core/src/Logger.js#L55

I ​​am​​ aware that Microsoft has ended support ​​for​​ Windows 7; however, ​​numerous​​ clients and legacy projects ​​require us to​​ continue using it. ​​A workaround is straightforward​​, and I would like to submit a PR.

robhogan:
Closing as per #1556 - Metro can't support Windows 7 because Node.js doesn't.

PR #1556, opened 2025-08-13
Fix: walk around uv_os_gethostname returned ENOSYS on Windows 7 node v18

tony9527167:
## Summary

Call os.hostname on Windows 7 node v18 got error uv_os_gethostname returned ENOSYS (function not implemented),
I ​​am​​ aware that Microsoft has ended support ​​for​​ Windows 7; however, ​​numerous​​ clients and legacy projects ​​require us to​​ continue using it. ​​A workaround is straightforward​​, and I would like to submit a PR.

See also
https://github.com/facebook/metro/issues/1555

Changelog:

## Test plan

I have confirmed my changes both work on Windows 7 and macOS 15.

### DIFF
--- a/packages/metro-core/src/Logger.js
+++ b/packages/metro-core/src/Logger.js
@@ -52,6 +52,13 @@ export type LogEntry = {
   ...
 };
 
+try {
+  os.hostname();
+} catch (err) {
+  console.warn(err)
+  os.hostname = () => "localhost";
+}
+
 const log_session = `${os.hostname()}-${Date.now()}`;
 const eventEmitter = new EventEmitter();
 


robhogan:
Hi @tony9527167, thanks for looking into this.

You mention Node 18, but does this bug reproduce on Node 20+? Metro's minimum supported Node.js version is 20.19, so if this fix is only relevant to Node 18 there's no point landing it on `main`. However, we could consider targeting a release branch to patch supported older versions.

tony9527167:
> Hi @tony9527167, thanks for looking into this.
> 
> You mention Node 18, but does this bug reproduce on Node 20+? Metro's minimum supported Node.js version is 20.19, so if this fix is only relevant to Node 18 there's no point landing it on `main`. However, we could consider targeting a release branch to patch supported older versions.

I understand most mainstream open-source projects and vendors have already abandoned Windows versions below 10 and Node.js v18 (or older non-LTS versions). Maintaining compatibility in such an outdated ecosystem has become extremely challenging.

I would like to conduct more comprehensive testing, but it's impossible to install Node.js v20 or newer versions (including via nvm) on Windows 7 systems.

My PR contains minimal changes that should be compatible with all Node.js versions, as the `os.hostname` API has been available since [v0.3.3](https://nodejs.org/docs/latest/api/os.html#oshostname).

robhogan:
> It's impossible to install Node.js v20 or newer versions (including via nvm) on Windows 7 systems.

Unfortunately, in that case it won't be possible to install Metro 0.83 or newer on Windows 7, so this patch wouldn't be available to you.

Which Metro version are you actually using? I'm open to merging ~~this fix~~ something equivalent into a release branch to patch a still-supported Metro version (anything used by the last three versions of React Native).

tony9527167:
> > It's impossible to install Node.js v20 or newer versions (including via nvm) on Windows 7 systems.
> 
> Unfortunately, in that case it won't be possible to install Metro 0.83 or newer on Windows 7, so this patch wouldn't be available to you.
> 
> Which Metro version are you actually using? I'm open to merging this fix into a release branch to patch a still-supported Metro version (anything used by the last three versions of React Native).

metro-core 0.82.4
react-native 0.79.5

byCedric:
FWIW, Windows 7 has been end-of-life since Januari 14, 2020. I don't think it makes a lot of sense for open-source projects to maintain enviroments that are not maintained by the owning company (Microsoft) for over 5 years. The same applies to Windows 10, which will be end-of-life in less than 2 months (October 14, 2025).

robhogan:
Closing, as Metro no longer supports Node 18 (nor, therefore, Windows 7)

PR #1565, opened 2025-08-29
Restore compatibility with interopRequireDefault (`import Things from 'metro-pkg'`) in Metro public Node.js APIs.

robhogan:
Summary:
In https://github.com/facebook/metro/pull/1549 (Migrate Metro to ESM syntax), which was intended to be non-breaking, there was one use case I overlooked.

If your code consuming Metro under Node.js uses ESM-style imports transformed to CommonJS with Babel's default interop mode, you could've written any of:

```
// Import default
import Resolver from 'metro-resolver';
Resolver.resolve('./foo');
```

or,

```
// Import all named as namespace
import * as Resolver from 'metro-resolver';
Resolver.resolve('./foo');
```

or, 

```
// Import named
import {resolve} from 'metro-resolver';
resolve('./foo');
```

However, when we switched from a `module.exports` object to ESM-style named exports, the first case would fail, because Babel marks our index file with `__esModule`, which causes `interopRequireDefault` to select the `default` prop (which is now undefined).

The consumer would be transformed to this, which would throw:

```
var _metroResolver = _interopRequireDefault(require("metro-resolver"));
function _interopRequireDefault(e) { return e && e.__esModule ? e : { default: e }; }
_metroResolver.default.resolve('./foo'); // TypeError: Cannot access property resolve of undefined
```

This restores compatibility by adding an explicit default export alongside the named exports, for all public package entry points that were not already `__esModule`s. These will be removed in a future breaking change.

Changelog: Internal - fix unreleased breaking change from https://github.com/facebook/metro/pull/1549

Reviewed By: vzaidman

Differential Revision: D81127738

### DIFF
--- a/packages/metro-babel-transformer/src/index.js
+++ b/packages/metro-babel-transformer/src/index.js
@@ -140,3 +140,11 @@ function transform({
 */
 
 export {transform};
+
+/**
+ * Backwards-compatibility with CommonJS consumers using interopRequireDefault.
+ * Do not add to this list.
+ *
+ * @deprecated Default import from 'metro-babel-transformer' is deprecated, use named exports.
+ */
+export default {transform};
--- a/packages/metro-babel-transformer/types/index.d.ts
+++ b/packages/metro-babel-transformer/types/index.d.ts
@@ -48,3 +48,14 @@ export interface BabelTransformer {
 }
 
 export const transform: BabelTransformer['transform'];
+
+/**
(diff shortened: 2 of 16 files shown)


robhogan:
This was merged in https://github.com/facebook/metro/commit/e7e69b1f429be93aa87419b4e34b6e5760649e09, I guess the bot missed closing this PR.

PR #1566, opened 2025-09-02
Pass `globalPrefix` through to serializer and expose to `getRunModuleStatement`

robhogan:
Summary:
To allow prefixing the global function `__r` (e.g., https://github.com/facebook/metro/pull/1512), expose the prefix to the configured [`getRunModuleStatement`](https://metrobundler.dev/docs/configuration/#getrunmodulestatement), which returns the `__r()` call as a string.

```
 - **[Feature]** Expose `globalPrefix` to `getRunModuleStatement`
```

## Alternative? Runtime prefixing
We could alternatively add prefix at runtime by emitting:
```
'globalThis[__METRO_GLOBAL_PREFIX__ + '__r'](/*...*/)'
```
There are a couple of downsides -
 - These runModule statements are currently outside IIFEs that inject `global` based on (something like) `global = globalThis ?? global ?? window` - we don't rely on the existence of `globalThis` elsewhere (though it's probably safe to now - that's a separate breaking change).
 - Concatenation/interpolation + object access is more verbose and slightly slower, and there's just no need for it in code emitted by Metro itself (as opposed to framework/userland code, which must use it).

Differential Revision: D81476621


jbroma:
this is perfect, thank you @robhogan 🙌

PR #1580, opened 2025-09-11
metro-config: Accept functions as arguments to mergeConfig

robhogan:
Extend the existing `mergeConfig` API so that it accepts functions as well as config objects.

These functions are provided with the leftward merged config as an argument, so that they may re-use or extend it in ways not covered by `mergeConfig`'s simple spread. For example:

# Example

Suppose `awesome-lib` exports `addThirdPartyMagic` for configuring Metro for some nice functionality, and it wants to merge intelligently with previous config:

```js
export function addThirdPartyMagic(baseConfig) {
  return {
    resolver: {
      assetExts: [...baseConfig.resolver.assetExts, 'magic'],
    },
  };
}
```

## Currently

This would require a nest or sequence of `mergeConfig` calls with previous configs explicitly passed to `addThirdPartyMagic`, e.g:

```js
const {mergeConfig} = require('metro-config');
const {getDefaultValues} = require('@react-native/metro-config');
const {addThirdPartyMagic} = require('awesome-lib');

const defaults = getDefaultValues(__dirname);
const myConfig = mergeConfig(defaults, {
   resolver: {
     assetExts: ['gif', ...defaults.resolver.assetExts],
   },
})
module.exports = mergeConfig(myConfig, addThirdPartyMagic(myConfig));
```

There's a lot of boilerplate here, it's easy to get wrong and it doesn't get any nicer when you introduce a second and third library that wants to customise config.

## Proposed
Instead, by allowing functions that supply the the previous config, users can do:

```js
const {mergeConfig} = require('metro-config');
const {getDefaultValues} = require('@react-native/metro-config');
const {addThirdPartyMagic} = require('awesome-lib');

module.exports = mergeConfig(getDefaultValues(__dirname), (baseConfig) => {
   resolver: {
     assetExts: ['gif', ...baseConfig.resolver.assetExts],
   },
}, addThirdPartyMagic);
```

And this scales - every additional library is just another argument to `mergeConfig`.

### Future?
Out of scope for just now, but we *could* consider an array exported from `metro.config.js` to mean arguments to `mergeConfig`, and libraries like `@react-native/metro-config` could export a (maybe async) function as a default export, so the future could look like this:

```ts
// metro.config.ts
import type {MetroConfig} from 'metro-config';

export default = [
  import('@react-native/metro-config'),
  (defaults) => ({
    resolver: {
      sourceExts: [...defaults.resolver.sourceExts, 'custom']
    }
  }),
  import('awesome-lib/metro-config'),
] satisfies MetroConfig[];
```

Differential Revision: D82221532

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -726,9 +726,13 @@ The default value is `['hg.update']`.
 
 Using the `metro-config` package it is possible to merge multiple configurations together.
 
-| Method                                  | Description                                                            |
-| --------------------------------------- | ---------------------------------------------------------------------- |
-| `mergeConfig(...configs): MergedConfig` | Returns the merged configuration of two or more configuration objects. |
+| Method                                  | Description                                                                         |
+| --------------------------------------- | ----------------------------------------------------------------------------------- |
+| `mergeConfig(...configs): MergedConfig` | Returns the merged configuration of two or more configuration objects or functions. |
+
+`configs` may be any combination of configuration objects or functions (from Metro 0.83.2). Functions are called with the merged config of all configs to the left, which may be useful for complex merges with the previous config.
+
+If any arguments are async functions, `mergeConfig` will return a `Promise`, otherwise it will return the merged config synchronously.
 
 :::note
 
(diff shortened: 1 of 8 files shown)


robhogan:
CC @tjzel - would this address the issues you had neatly configuring `react-native-worklets`?

tjzel:
@robhogan Thanks for the ping. My problems were already fixed with the PR to the CLI but this is probably even better 👍

PR #1594, opened 2025-10-08
feat: future modules

tjzel:
## Summary

Library `react-native-worklets` generates dependency files during Babel transformation for a full encapsulation of some functions as modules. However, Metro doesn't receive file system updates from Watchman in time, due to its synchronous nature. This results in Metro crashing due to a failed lookup from internal file system.

This PR adds a way of informing Metro ahead of time that a module is a `futureModule` meaning that it shouldn't look for this file in its internal file system. Registration is done by the interested library during Babel transpilation with `metadata.futureModules` property in the transformation state.

Future modules have their full path set so that they can be resolved with the default `resolve` function.

The SHA for future modules is calculated using `performance.now()` to guarantee a stale hash and a forceful read from the physical file system to obtain the contents of the file.

Changelog: [Feature] Feature Modules handling - modules that aren't yet in Metro filesystem but exist on disk.

## Test plan

1. `yarn add --dev react-native-worklets`
2. Apply patch for Worklets Babel Plugin to use `futureModules` feature:
```diff
diff --git a/node_modules/react-native-worklets/plugin/index.js b/node_modules/react-native-worklets/plugin/index.js
index 6419dc4..3c26fb8 100644
--- a/node_modules/react-native-worklets/plugin/index.js
+++ b/node_modules/react-native-worklets/plugin/index.js
@@ -560,6 +560,13 @@ var require_generate = __commonJS({
         (0, fs_1.mkdirSync)(filesDirPath, {});
       }
       const dedicatedFilePath = (0, path_1.resolve)(filesDirPath, `${workletHash}.js`);
+      const requirePath = `react-native-worklets/__generatedWorklets/${workletHash}.js`;
+      if (!state.file.metadata.metro.futureModules) {
+        state.file.metadata.metro.futureModules = /* @__PURE__ */ new Map();
+      }
+      state.file.metadata.metro.futureModules.set(requirePath, {
+        fullPath: dedicatedFilePath
+      });
       (0, fs_1.writeFileSync)(dedicatedFilePath, transformedProg);
     }
   }
@@ -1727,6 +1734,7 @@ module.exports = function WorkletsBabelPlugin() {
     pre() {
       runWithTaggedExceptions(() => {
         (0, globals_1.initializeState)(this);
+        this.opts.bundleMode = true;
       });
     },
     visitor: {
```
3. Add `'react-native-worklets/plugin'` to `babel.config.js` plugin list.
4. Create a following file in the root of the repository:
```js
function foo() {
  'worklet';
  return null;
}

foo();
```
5. `rm -fr node_modules/react-native-worklets/__generatedWorklets` to remove generated files for a clean build (only when you already had done step 6. in previous test)
6. `yarn start build file.js --out bundle.js --minify=false --reset-cache`, see that without these changes the build fails with errors related to missing files

| Before | After |
|--------|--------|
| Screenshot 2025-10-08 at 14 41 51 | Screenshot 2025-10-08 at 14 43 21 |


robhogan:
> generates dependency files during Babel transformation

Sorry @tjzel, but this just isn't something we can support. A Babel plugin should not be performing file IO. If it does, we can't hope to maintain cache integrity.

I think the solution here is not "future modules" but virtual modules, which would be purely a transform function of resolver output. I know we spoke about this before and sorry I haven't had chance to dig into whether that theoretical solution works for worklets, but it's something I'd want to consider first. I'm afraid this PR's proposal breaks the architectural model of Metro, where transforms are side-effect free functions of one individual file's content (plus config), and resolvers are (synchronous, ideally cheap) functions of the codebase.

If the resolver provided (for example) the whole contents of the origin module to the transformer, could the transformer then generate the dependencies needed for worklets without further file reads?

tjzel:
@robhogan Thanks for the feedback. The IO operations aren't necessary for this to work.

> If the resolver provided (for example) the whole contents of the origin module to the transformer, could the transformer then generate the dependencies needed for worklets without further file reads?

I'm not sure what do you mean exactly. I can't know what virtual modules will be created ahead of transformation, so the information about which modules are virtual has to come from the transformer. I can make the transformer output the virtual modules and then pass upstream like in this PR but I don't know if it's okay with you.

If the resolver had access to the transform function I could pass just the virtual module name and the origin module to the resolver and generate the module's code to be kept in-memory.

There are generally three cases that I need to cover:
1. A module creates one or more virtual modules and depends on all of them.
2. A module creates one or more virtual modules but doesn't depend on all of them.
3. A virtual module depends on one or more virtual modules but doesn't create them.

PR #1601, opened 2025-10-13
ref: Use slice().reverse() instead of toReversed() for backwards compatibility

antonis:
## Summary

Minor refactoring to use `slice().reverse()` instead of `toReversed()` for backwards compatibility. 

[toReversed is an ES2023 feature](https://www.w3schools.com/js/js_2023.asp?#mark_array_toreversed) not available in older JS environments, causing runtime errors in React Native projects with older Hermes versions.

This results in errors like the following 
```
TypeError: configs.toReversed is not a function
```

Changelog:

### DIFF
--- a/packages/metro-config/src/loadConfig.js
+++ b/packages/metro-config/src/loadConfig.js
@@ -221,15 +221,15 @@ function mergeConfig<
   ? T
   : Promise<T> {
   let currentConfig: T = typeof base === 'function' ? base() : base;
-  // Reverse for easy popping
-  const reversedConfigs = configs.toReversed();
+  // Reverse for easy popping - use slice().reverse() for compatibility
+  const reversedConfigs = configs.slice().reverse();
   let next;
   while ((next = reversedConfigs.pop())) {
     const nextConfig: InputConfigT | Promise<InputConfigT> =
       typeof next === 'function' ? next(currentConfig) : next;
     if (nextConfig instanceof Promise) {
       // $FlowFixMe[incompatible-type] Not clear why Flow doesn't like this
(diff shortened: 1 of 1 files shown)


robhogan:
Hi @antonis - thanks for the PR.

This code should only run on Node.js, not on the client (Hermes) - and I thought `.toReversed()` was supported by our minimum Node.js version. Can I ask what version you're seeing an issue with?

node.green says this is ok in 20.11: https://node.green/#ES2023-features-Change-Array-by-copy-Array-prototype-toReversed--

Metro's minimum is 20.19: https://github.com/facebook/metro/blob/0f7649e25c3bc7e34b1e13b01e990622fdf3a7c0/packages/metro/package.json#L80

antonis:
Hello @robhogan,
thank you for iterating on this 🙇 

> Can I ask what version you're seeing an issue with?

I've noticed this with RN 0.81.0 on one of our E2E tests that [use node 18](https://github.com/getsentry/sentry-react-native/blob/main/.github/workflows/e2e-v2.yml#L235)

robhogan:
Ah I see - Metro and RN 0.81 has a minimum Node.js version of 20.19 so failing on Node 18 isn't a bug.

I'd recommend enforcing `engines` constraints with Yarn to make that clearer - maybe we can modify the template so this is the default (Yarn v1 does out of the box, but I think 2+ needs a plugin).

I'm going to close this, but thanks for raising it - we should make it more obvious for folks running on unsupported versions.

taylorkline:
For me, the issue occurred because I still had an `.xcode.env.local` that pointed to an old version of Node.

PR #1605, opened 2025-10-30
feat: virtual modules

tjzel:
## Summary

Changelog: [Feature] Virtual Modules handling - modules that are in-memory only.

PR based on rejected
- #1594

This PR adds an early implementation of virtual modules handling in Metro, modules which exist only in memory and not on the disk.

My motivation for implementing virtual modules is that `react-native-worklets` would like to emit extra, encapsulated modules, based on some code, during the Babel transformation of the user's source code.

### Basic idea

A Babel plugin during transpilation of (physical or virtual) module can register virtual modules via APIs provided by Metro. Registration consists of providing the following data:
- An identifier of the module. This could be a path (that either exists or not) that uniquely identifies the virtual module.
- The source code of the module. The source code has to undergo the Metro transformation in the future, it's not ready to be bundled.
- The source maps of the module. I have yet to learn more on this topic - preferably the source maps would be inlined with the source code, but I don't know if it's possible with how Metro handles sourceURLs etc.

The registering module can either depend on the generated virtual module or not.
A module cannot depend on a virtual module it didn't generate - it might not be possible to synchronize the registrations in time. In the future this could be loosened.

Ideally the virtual module would land in the Metro's virtual FS and treated as a regular file for the purpose of getting its contents, calculating SHA etc. However I don't know yet how feasible that is since there's a great deal of asynchronicity between Metro workers and its core components, like the FS.

### Implementation

Because of the great amount of decentralization that is present in the Metro bundling system adding this feature seems to somehow add a lot of redundancy. Here are the simplified key points of the implementation:

1. A Babel plugin registers virtual modules in [packages/metro-transform-worker/src/index.js](https://github.com/facebook/metro/pull/1605/files#diff-a33a60c1147a6912face4916c1ab34e0c923f636f4670e931695767749288a7f) with `transformResult.metadata?.metro?.virtualModules` property.
2. The map of virtual modules gets serialized in [‎packages/metro/src/DeltaBundler/Worker.flow.js‎](https://github.com/facebook/metro/pull/1605/files#diff-6a9838d13ff5c199e16369ad162aa1a11b2492886cdc59e39c0b36f6e41fa25f) and passed to the DeltaBundler.
3. Delta bundler merges newly registered virtual modules with currently known ones and marks modules as virtual accordingly [‎packages/metro/src/DeltaBundler.js‎](https://github.com/facebook/metro/pull/1605/files#diff-b88f009cbf725e85da9ba5f10080ff50e294394cbeb23b846fd800917b96425c).
4. (currently a stub implementation) Node haste has special handling of virtual modules when creating a SHA [‎packages/metro/src/node-haste/DependencyGraph.js‎](https://github.com/facebook/metro/pull/1605/files#diff-d9479007d2e919675cefedcdcba0078188c51a35b2c20069d1ec0ba186151ad2)
5. (currently a stub implementation) Contents of the module are obtained from the registered source code [‎packages/metro/src/node-haste/DependencyGraph.js‎](https://github.com/facebook/metro/pull/1605/files#diff-d9479007d2e919675cefedcdcba0078188c51a35b2c20069d1ec0ba186151ad2)

## Test plan

TBA

I got it working properly in [Reanimated monorepo](https://github.com/software-mansion/react-native-reanimated).


jbroma:
Hey @tjzel,

do you have an example on how this is used within Reanimated to register virtual modules? In other words, how does the Metro public API looks like for VMs. 

For context, In Module Federation implementation for Metro we have hacked our own virtual modules implementation but it's very brittle long-term and would love to move to this stable solution 🙌

tjzel:
Hey @jbroma, sorry I missed your message.

The API to use it isn't pretty unfortunately since it's handled during Babel transpilation. You can see it being used on my draft branch:

https://github.com/software-mansion/react-native-reanimated/blob/%40tjzel/worklets/metro-pr-patches/packages/react-native-worklets/plugin/src/generate.ts#L95C1-L115C6

Long story short, we generate the full code of the virtual module and register it in the transpilation process metadata.

On the other hand, with current Metro APIs I believe it's quite a challenge to create a convenient API for such a feature.

JaMinsane:
Hi! Thanks for the work here.
Since `react-native-worklets` 0.10.0 now marks Bundle Mode as stable and recommended, I wanted to ask whether this draft PR is still expected to move forward.
From the Worklets setup docs, enabling Bundle Mode still requires temporary patches for `metro` and `metro-runtime`, so I’m trying to understand whether this PR is still the intended path to remove those workarounds, or if there is another Metro-side approach planned.
Thanks in advance for any context 🙌

tjzel:
@JaMinsane I hope so. At the moment there's a bit of a maintainer shortage for Metro and it's difficult to get new features in. I'm working with the Expo team to include similar changes on their end, so it'd work there without the patches.

Issue #1491, opened 2025-04-21
Incorrect TypeScript Types


nickhudkins:
Many of the TypeScript definitions specify `void` as a type parameter for generic types where the equivalent Flow type definition uses an *empty* type parameter:

### TypeScript Type Definition (🤨)
https://github.com/facebook/metro/blob/76413561abb3757285e0cb0305f1f9f616fa2b6c/packages/metro/types/IncrementalBundler.d.ts#L25

### Flow Type Definition
https://github.com/facebook/metro/blob/76413561abb3757285e0cb0305f1f9f616fa2b6c/packages/metro/src/IncrementalBundler.js#L36

I would posit that the TypeScript definition is... mad sus, and I suspect we might just be running into this because the types are hand-written. Am I missing something and believing this is wrong? I'll open a PR to correct assuming this is indeed an incorrect typing.

robhogan:
I believe the TS types were originally generated by [`flow-api-translator`](https://www.npmjs.com/package/flow-api-translator) plus some tweaking, and have been hand maintained since. They could well be wrong in places, but this might also be a translator bug.

CC @huntie

huntie:
@nickhudkins Yup they're currently hand maintained, so would appreciate a targeted PR with test plan :)

@robhogan We should look at type translation for Metro again — there's been 15+ fixes to flow-api-translator this year alone.

nickhudkins:
Looks like `flow-api-translator` might be missing a test case for empty type parameters. I'll start there.

nickhudkins:
@robhogan heads up: https://github.com/facebook/hermes/pull/1676/checks. Looks like there might be a bit of setup for me to be able to run the jest tests that is not documented, in the shortest term would you mind allowing the workflow to run over there?

robhogan:
@huntie yep - but I mainly CCed to alert to the fact that if this is a bug in `flow-api-translator` it probably means the RN types are also broken.

@nickhudkins thanks for digging into this!

nickhudkins:
Running into quite a number of issues getting the `js` tests running:

This comment has me concerned if I will be able to get this running. Any further docs would be much appreciated, otherwise will continue to dig as to how to get this up and running:

https://github.com/facebook/hermes/blob/11c154bc365447648fa56fdd67a1c142034b85a6/tools/hermes-parser/js/scripts/build.sh#L24

huntie:
[...]

PR #1608, opened 2025-11-07
fix(types): incorrect <void> use in typescript

nickhudkins:
## Summary
The use of `void` as a type parameter is semantically different in TS than Flow. Specifically, providing `void` in Flow will result in the use of the default type parameter value, while TS will treat it as an explicit use of `void` for the generic. 

Original Issue: https://github.com/facebook/metro/issues/1491

Changelog:
[Fix] `Bundler` types (`transformFile`)
[Fix] `IncrementalBundler` types (`OutputGraph`, `getDeltaBundler`, `getDependencies`, `buildGraph`, `initializeGraph` and `updateGraph`)
[Fix] `Worker` types (`transform`)
[Fix] `Metro.buildGraph` return type

## Test plan
Oof. This may be a larger discussion as keeping the TS and Flow types in-sync and correct may be a moving target as the project evolves. For now, the verification is "when consuming this code in a TS codebase, types are the same as if I was consuming the flow annotated code"

### DIFF
--- a/packages/metro/types/Bundler.d.ts
+++ b/packages/metro/types/Bundler.d.ts
@@ -33,7 +33,7 @@ export default class Bundler {
     transformOptions: TransformOptions,
     /** Optionally provide the file contents, this can be used to provide virtual contents for a file. */
     fileBuffer?: Buffer,
-  ): Promise<TransformResultWithSource<void>>;
+  ): Promise<TransformResultWithSource>;
 
   ready(): Promise<void>;
 }
--- a/packages/metro/types/DeltaBundler/Worker.d.ts
+++ b/packages/metro/types/DeltaBundler/Worker.d.ts
@@ -34,7 +34,7 @@ export interface TransformerConfig {
 }
 
 interface Data {
-  readonly result: TransformResult<void>;
+  readonly result: TransformResult;
   readonly sha1: string;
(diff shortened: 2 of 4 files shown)


robhogan:
Thanks @nickhudkins!

Several Metro packages do now have automatic TS types, but there are a few patterns unsupported by the translator and/or TS itself so we've been refactoring and touching up package-by-package to get translations working (any PRs to get us closer would be appreciated as well!). `metro-config` for example is automatic and validated in CI, but `metro` isn't.

https://github.com/facebook/metro/blob/main/scripts/generateTypeScriptDefinitions.js#L29

nickhudkins:
I actually saw this and began down the path of doing type generation for metro, let me see if I can take an incremental approach here!

nickhudkins:
Alrighty... `metro` types are gonna take some more significant work, especially just trying to reconcile flow types with typescript, especially those that have [spreads at the end](https://github.com/facebook/hermes/blob/b43a049628f6a81cd828afe2db93b05e49237542/tools/hermes-parser/js/flow-api-translator/src/flowDefToTSDef.js#L3398-L3401)

robhogan:
Closing this now that we have (almost) fully auto-generated TS types with validation - thanks for your help with that!

PR #1610, opened 2025-11-07
chore(types): auto-generate types for metro-transform-worker

nickhudkins:
## Summary

As I began working to hand-fix some types, the formidable @robhogan [pointed me](https://github.com/facebook/metro/pull/1608#issuecomment-3503159420) in the direction of `generateTypeScriptDefinitions.js`, which has been used here to generate types for `metro-transform-worker`

Changelog: [Internal]: Auto generate types for `metro-transform-worker`

## Test plan
Automatically tested in CI, also peformed a good-ol eyeball check.

### DIFF
--- a/packages/metro-transform-worker/types/index.d.ts
+++ b/packages/metro-transform-worker/types/index.d.ts
@@ -8,7 +8,6 @@
  * @oncall react_native
  */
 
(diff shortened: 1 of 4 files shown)


vzaidman:
changed the changelog to "internal" before landing this.

PR #1611, opened 2025-11-08
chore(types): auto-generate types for metro-file-map

nickhudkins:
## Summary

As I began working to hand-fix some types, the formidable @robhogan [pointed me](https://github.com/facebook/metro/pull/1608#issuecomment-3503159420) in the direction of `generateTypeScriptDefinitions.js`, which has been used here to generate types for `metro-file-map`

Changelog:
[Fix]: Auto generate types for `metro-file-map`

## Test plan
Automatically tested in CI, also peformed a good-ol eyeball check.

### DIFF
--- a/packages/metro-file-map/src/index.js
+++ b/packages/metro-file-map/src/index.js
@@ -236,7 +236,7 @@ const WATCHMAN_REQUIRED_CAPABILITIES = [
  */
 export default class FileMap extends EventEmitter {
   _buildPromise: ?Promise<BuildResult>;
-  canUseWatchmanPromise: Promise<boolean>;
+  _canUseWatchmanPromise: Promise<boolean>;
   _changeID: number;
   /*::
   _changeInterval: ?IntervalID;
--- a/packages/metro-file-map/src/lib/TreeFS.js
+++ b/packages/metro-file-map/src/lib/TreeFS.js
@@ -1010,7 +1010,7 @@ export default class TreeFS implements MutableFileSystem {
     return null;
   }
 
-  *metadataIterator(opts: MetadataIteratorOptions): Iterable<{
+  *metadataIterator(opts: MetadataIteratorOptions): Iterator<{
     baseName: string,
     canonicalPath: string,
     metadata: FileMetadata,


nickhudkins:
There are a few failures here, some are easily fixable, others will take some finagling. 

1.) `IntervalID` and `TimeoutID`, we can simply exclude them with `/*:: */` to avoid problems for now, I think the fields wouldn't ever be public anyway.
2.) Incomplete third-party types. For example `fb-watchman` has types defined in `flow-typed`. `flow-api-translator` does NOT play nicely with this. 

**Option 1:** Use `@types/fb-watchman`
 - 😔 these types are less complete than the types in `flow-typed`, and leave us with remaining errors.
 - We could contribute these types back upstream and wait.
 
**Option 2:** Some Naive Tooling!
- `flow-api-translator` will happily convert MOST of those types, but the contents need to be unwrapped from: `declare module '{{ moduleId }}' { ...definitions... }`. 
- The `declare module.exports:` syntax also needs translation.

> [!NOTE] 
> I have a local script that performs these pre-processing steps to be able to successfully generate TS *at least* for `fb-watchman`. It is all naive string matching. Maybe this is fine.

**Option 3:** Make flow-api-translator "Just Work"
This would be neat, but I am unsure if it is worth the investment

packages/metro-file-map/src/crawlers/watchman/index.js
```diff
@@ -92,7 +92,7 @@ export default async function watchmanCrawl({
         command,
       });
     };
-    let intervalOrTimeoutId: TimeoutID | IntervalID = setTimeout(() => {
+    let intervalOrTimeoutId: number = setTimeout(() => {
```
nickhudkins: revert this. use /*:: */ if need be.

packages/metro-file-map/src/plugins/MockPlugin.js
```diff
@@ -28,7 +28,7 @@ import path from 'path';
 export const CACHE_VERSION = 2;
 
 export default class MockPlugin implements FileMapPlugin<RawMockMap>, IMockMap {
-  +name = 'mocks';
+  +name: 'mocks' = 'mocks';
```
nickhudkins: same, `flow-api-translator` did not infer the type well here.

scripts/generateTypeScriptDefinitions.js
```diff
@@ -107,6 +119,31 @@ export async function generateTsDefsForJsGlobs(
         return;
       }
       try {
+        if (isModuleDeclarationOnly(source)) {
```
nickhudkins: this section does not lint, or prettify, or anything. I have a local refactor that turns this into a small "pipeline" of sorts with pre & post actions, since the difference between transforming source and module definitions fit cleanly in before & after while `translateFlowDefToTSDef` does the majority of the work.

packages/metro-file-map/src/__tests__/worker-test.js
```diff
@@ -297,6 +297,7 @@ describe('jest-worker interface', () => {
 
   beforeEach(() => {
     jest.resetModules();
+    // Re-require the worker module to reset its internal state.
```
nickhudkins: what? get rid of you

packages/metro-file-map/src/flow-types.js
```diff
@@ -472,4 +472,6 @@ export type WorkerMetadata = $ReadOnly<{
   content?: ?Buffer,
 }>;
 
-export type WorkerSetupArgs = $ReadOnly<{}>;
+export interface WorkerSetupArgs {
+  __future__?: false;
```
nickhudkins: this is another one of those "doesn't translate to TS well" (empty object types), at the moment tossed in a field, but realistically there is probably a better type all together?

packages/metro-file-map/src/lib/dependencyExtractor.js
```diff
@@ -14,26 +14,23 @@
 'use strict';
 
 const NOT_A_DOT = '(?<!\\.\\s*)';
-const CAPTURE_STRING_LITERAL = (pos /*: number */) =>
+const CAPTURE_STRING_LITERAL = (pos: number) =>
```
nickhudkins: I believe this is a file that *needs* to be CommonJS, though interestingly making this change did not cause test failures beyond the shape of the exports changing from using `export default` instead of `module.exports`

PR #1620, opened 2025-11-20
fix(metro-config): Convert absolute paths to file URL before calling await import

kitten:
## Summary

Errors may be swallowed in `require`, which then cascades the error to `await import`. However, this may then trigger an error on Windows since absolute Windows paths cannot be passed to `await import` and need to be file URLs instead (`C:` or `D:` look like protocols to `import` otherwise)

We could also expose the error from `require` instead if the `code`s match a set of allowed (or not match a set of disallowed) error codes, but it's hard to get a comprehensive list for those without testing across all Node LTS versions.

Changelog: [Fix] Prevent absolute path error on `await import()` for Windows absolute paths

## Test plan

This is admittedly a speculative fix, and I still need to test this, but I've only observed this in Expo CI for now.

### DIFF
--- a/packages/metro-config/src/loadConfig.js
+++ b/packages/metro-config/src/loadConfig.js
@@ -18,6 +18,8 @@ import {validate} from 'jest-validate';
 import * as MetroCache from 'metro-cache';
 import {homedir} from 'os';
 import * as path from 'path';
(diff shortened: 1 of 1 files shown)


Swiftwork:
In what release and when are we expecting this to be publicly available?

PR #1625, opened 2025-11-25
chore(types): provide ability to define .d.ts as source

nickhudkins:
Provides an escape hatch by allowing users to define .d.ts as sibling files to source, as type generation overwrites files in `types`, and removes orphaned files. This can also be used in the case where flow-api-translator produces incorrect typings, or fails entitely

## Summary

Discussions with @robhogan and other PRs that I hope to land could benefit from being able to bail on type-generation without manually defined .d.ts files being removed as part of generation.

Changelog:
[Internal] Allow .d.ts files in `src` to allow for manual definitions

## Test plan

https://github.com/user-attachments/assets/9d76b6cb-687f-4ebc-bfd9-40eff27bdd90

### DIFF
--- a/scripts/generateTypeScriptDefinitions.js
+++ b/scripts/generateTypeScriptDefinitions.js
@@ -25,6 +25,7 @@ import * as prettier from 'prettier';
 const WORKSPACE_ROOT = path.resolve(__dirname, '..');
 
 const TYPES_DIR = 'types';
+const SRC_DIR = 'src';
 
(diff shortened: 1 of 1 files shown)


nickhudkins:
One thing we could choose to do here is to turn on generation across the entire repo, and move the manually typings into `src` for the failure cases. Not sure if that's an approach we want to take, but it would at least prevent manual typings from drifting between now and whenever fully generated types can be supported without any overrides. 

Thoughts?

Issue #1629, opened 2025-12-10
Invalid content type, expected application/json, parseJsonBody.js, symbolicate endpoint


mlassauzay-pass:
Hi ! 

I just bumped RN from 0.78 to 0.80 and I'm getting this error message triggered in my code when I change screens.
```
Error: Invalid content type, expected application/json
    at .../node_modules/metro/src/lib/parseJsonBody.js:11:16
   ...
```
A request is made to the symbolicate endpoint, which attempts to parse the request via parseJsonBody.  
parseJsonBody has been modified between 0.78 and 0.80 and only accepts requests with `req.headers["content-type"]` which contains `application/json`, but in my case the request it receives contains "text/plain", so it displays this error message. The request content does indeed appear to be JSON, and if I comment out the header validation, the request seems to be processed correctly and I no longer get an error message.

The request trace points to a call to `navigate` in my code, but I have no further information and navigation works correctly. I tested this with react-navigation 6 and 7. 

I'm wondering if anyone else is familiar with this problem, which is causing a lot of error message on the my console ?
I haven't been able to determine if it's a problem with one of my libraries or if parseJsonBody could simply be more permissive.

robhogan:
Hi @mlassauzay-pass, thanks for the report. If the header constraint is new that might be an unintentional breaking change we could relax (although as you say the content must be JSON in any case).

It’d be interesting to know where the request to `/symbolicate` is coming from, since this isn’t a problem we’ve heard about iirc. Is this vanilla/community RN? Do you have anything installed that might be doing its own error handling (sentry or similar?)

robhogan:
Looks like the change was in https://github.com/facebook/metro/pull/1475 and I did spot the breaking change in (internal) review but it slipped through anyway.

If you want to put in a PR to update the call to `parseJsonBody` on `/symbolicate` to use `strict: false`, I’d be happy to accept that.

mlassauzay-pass:
Thanks for your answer @robhogan. 

Unfortunately I can't get any more information about the request send to symbolicate/. I know the trace sent refer to a `navigate` in my code, but I have no idea exactly what triggers the request and why.

I'm not sure that answers the question but the project doesn't use expo, we use directly `react-native": "0.80.2"` in package.json and `@react-native-community/cli` is in devDependencies. 
We use Sentry but it is disabled in `__DEV__` so I don't think comes from there. 

[...]

PR #1630, opened 2025-12-13
fix potential "invalid content type" error on /symbolicate #1629

MorvanL:
**It's my first PR, I have completed the CLA**

## Summary
Fix: https://github.com/facebook/metro/issues/1629

In my case with RN0.80, /symbolicate receives request with `req.headers["content-type"]` containing `text/plain` instead of `application/json` but the content is indeed a JSON. 
This causes this error message in the console : 
```
Error: Invalid content type, expected application/json
    at .../node_modules/metro/src/lib/parseJsonBody.js:11:16
   ...
```
Calling jsonParseBody with `strict: false` disable the header validation and the request is processed correctly and I no longer get an error message.
So we can propably consider making this header check more flexible and reintroduce it later as a breaking change if necessary (the original change was not introduced as a breaking change)

Changelog: [Fix] potential "invalid content type" error on /symbolicate #1629

### DIFF
--- a/packages/metro/src/Server.js
+++ b/packages/metro/src/Server.js
@@ -1382,7 +1382,7 @@ export default class Server {
         const body = await req.rawBody;
         parsedBody = JSON.parse(body);
       } else {
-        parsedBody = (await parseJsonBody(req)) as {
+        parsedBody = (await parseJsonBody(req, {strict: false})) as {
           stack: $ReadOnlyArray<StackFrameInput>,
           extraData: {[string]: mixed},
         };


vzaidman:
@robhogan should we perhaps improve the error thrown if the content is not JSON in 
https://github.com/facebook/metro/blob/8c22d54627a0bf6adf1c5ee6e635e2424fbed851/packages/metro/src/lib/parseJsonBody.js#L54 ?

huntie:
@MorvanL I'm still curious where the original request comes from, as we intentionally updated this ahead of time in the relevant CLI call sites. Requiring `application/json` is appropriate behaviour (in fact the `strict` option in our parsing util was only there for compatibility across CLI releases) — ideally we update the requestor instead.

Basically, I'm motivated by not adding complexity/multiple success paths here when technically the incoming request is out of spec.

If we do decide to merge this (cc @vzaidman), can we:
- Change this from a "Fix" to a "Feature" (we're _widening support_).
- Update this within the `parseJsonBody` util itself.

I'd still like to strongly push for fixing this problem at source however.

vzaidman:
@huntie for call sites coming from external tools outside of Metro, this is a breaking change so it's sensible to treat it as a fix that reverts a breaking change. But "feature" works as well.

robhogan:
@huntie See the attached issue for more context - this change was my recommendation because the issue caused by an unintended breaking change.

It would be interesting to know what the client is, but IMO it’s correct to fix the BC anyway.

huntie:
@robhogan Got it. If you're happy, let's go ahead :)

PR #1638, opened 2026-01-09
feat: add support for including the babel partial config to cache key

EvanBacon:
## Summary

Update the Metro Babel transformer to generate cache keys that include the contents of user Babel configuration files, ensuring cache invalidation when Babel configs change. This is similar to what Jest does and it seems to work really well for them.

Changelog: [Fix] Include user-defined babel config in transformer cache key to ensure correctness


robhogan:
Also, re previous discussion about perf impact - IIRC we both thought this was called for each worker - it's actually only called once by the main thread, so no concerns.

PR #1649, opened 2026-02-09
fix failing on latest babel types unsupported

vzaidman:
Summary:
A typo in the skip script was causing babel validations to not be skipped when running with latest babel versions.
Also, improved the comment above it.

Changelog: Internal

Differential Revision: D92703751


robhogan:
Merged in https://github.com/react/metro/commit/10354e6232e4e6b13378d05950fb74ecbc2720e4

PR #1652, opened 2026-02-17
Include npm install command in Getting Started guide

fatmapertiwi35l-ai:
Add npm installation command for metro and metro-core

## Summary

Changelog:


robhogan (changes requested):
It's already there further down. This PR adds it into the header, which is wrong.

Issue #1561, opened 2025-08-19
Tree Shaking Support


ArrayZoneYour:
I’ve ported most of the tree-shaking implementation from esbuild to a Metro (Babel) plugin and have been running it in production for about a year. Since #632 is closed, I’d like to know Metro’s plans for tree shaking.  I’m happy to contribute or upstream my work if that would be useful.

https://github.com/xhsfe/thor-x/blob/main/packages/metro-plugin-tree-shaking/DESIGN.md

robhogan:
Hey @ArrayZoneYour  - we're definitely interested in upstreaming any tree shaking implementation that's compatible with Metro's architecture (ie, does as much as of the heavy work as possible within cacheable transforms). 

At first look this is super impressive and you've obviously worked out Metro's internals in some depth! There are couple of it looks like we'd need to find solutions for - e.g. the limitation you mention "When tree-shaking is enabled, transform result is not only generated by input file." sounds like a potential problem for keying the transform cache.

I'll find some time this week to look through this more closely and understand the design better and get back to you. If you're not already on the React discord, drop me an email on my github handle at meta.com and I'll send you an invite, might be useful to chat there.

CC @EvanBacon for design thoughts/comparison with Expo tree shaking as well.

ArrayZoneYour:
@robhogan Thanks for your reply. I’ve sent the email and look forward to making further progress on tree shaking 😄.

ArrayZoneYour:
> "When tree-shaking is enabled, transform result is not only generated by input file." sounds like a potential problem for keying the transform cache.

Yes, I create metro server twice currently, The cache can be reuse in prev builds is ModuleGraph & FileCache. Transform cache should be isolated and handle carefully (should involve liveParts info if remote cache enabled ?)

 I use environment variable to identify whether the transform is tree shaking enable (like `process.env.PRE_BUILD`) then add it to partialKey stableHash args currently.

KingAmo:
any update on this?

Issue #227, opened 2018-08-23
Tree shaking while bundling


adnaan1703:
Is it possible anyhow to implement **tree shaking** during bundling.

Suppose I have a file say `greetings.js`:

```javascript
export const sayHi = (name) => {return `Jarvis says hi to ${name}`};
export const sayBye = (name) => {return `Jarvis says bye to ${name}`};
```

Now if in my project I am only using `sayHi` then I want `sayBye` to be removed from the final bundle which is currently not the case. 
Is it possible anyhow to implement this (even if I had to do changes in the config file for the metro bundler).

rafeca:
Metro currently does not support tree shaking based on module import/export. @mjesun is currently adding first-class support for Metro to understand them and this will eventually allow us to do tree shaking.

Stayed tuned! 😃

ruyadelaosan:
any outcome on tree-shaking ? Looking forward to that.

karanjthakkar:
@rafeca Now that we have [experimental support for import/export](https://github.com/facebook/metro/releases/tag/v0.45.1) inside metro, is tree shaking now available out of the box or is there work going on to support it? Anything that we can help with to make this possible?

rafeca:
@karanjthakkar : we're planning to implement some sort of tree shaking (or better said, minification of code based on the es6 module information) during H1 2019.

Based on our internal tests, the most impactful minification in terms of bundle size that can be done is to minify the import/export names, so for example:

```js
// bar.js
export const ANY_VARIABLE_NAME = 'value';
```

```js
// foo.js
import {ANY_VARIABLE_NAME} from './bar';
console.log(ANY_VARIABLE_NAME);
```

would get minified to:

```js
// bar.js
exports.a = 'value';
```

```js
// foo.js
console.log(r('./bar').a);
```

Afterwards, we're going to be able to do dead code elimination based on the exported values that are not being used anywhere

camdagr8:
Any movement or resolution on this?

aleclarson:
Is this still slated for H1 2019?

sunnylqm:
<img src="https://user-images.githubusercontent.com/615282/57971722-3dddae80-79c4-11e9-8a6c-0fec70d5e2b1.png" width="320">

Jeevan-Kishore:
Is it supported yet?

Mitsichury:
Hi ! Any news ?

vrgimael:
Seems that's rather late

neiker:
[...]

Issue #132, opened 2018-02-02
Add Tree Shaking on Production build


yordis:
I think the title is self-descriptive.

derrickbeining:
I second this request

phsantiago:
What we need to add tree shaking? Can't we use uglify js tree shaking before run metro bundler?

iketw:
@phsantiago could you expand on how you would achieve tree shaking with uglify js?

PR #1664, opened 2026-03-05
feat: experimental tree shaking support

bacarybruno:
## Summary

> [!NOTE]
> Sharing this draft early to get initial feedback and iterate.

Early benchmarking is promising:
- Baseline: 12,983,961 bytes 115s
- Tree-shake: 11,939,372 bytes, 107s
- Delta: -1,044,589 bytes (-8.05%), time -8s
Validation so far is based on repeated bundle runs (tree-shake off vs on).

Minimal metro config:
```ts
const useTreeShake = process.env.METRO_TREE_SHAKE === '1'

const config = {
  ...(useTreeShake
    ? {transformerPath: require.resolve('metro-transform-worker')}
    : {}),
  transformer: {
    unstable_treeShake: useTreeShake,
  },
}
```

Todo:
- [x] Add `unstable_treeShake` flag
- [x] `sideEffects` support
- [x] Bench before / after
- [ ] Manual tests
- [ ] Run broader before/after benchmarks (multiple apps/configs)
- [ ] Ensure no regression when `unstable_treeShake` is disabled / not set
- [ ] Documentation

## Test plan

Related issues:
- https://github.com/facebook/metro/issues/1561
- https://github.com/facebook/metro/issues/132
- https://github.com/facebook/metro/issues/227

The implementation plan can change depending on https://github.com/react-native-community/discussions-and-proposals/pull/949

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -38,4 +38,4 @@ untyped-import
 untyped-type-import
 
 [version]
-^0.303.0
+^0.304.0
--- a/flow-typed/environment/node.js
+++ b/flow-typed/environment/node.js
@@ -21,7 +21,7 @@ interface ErrnoError extends Error {
   syscall?: string;
 }
 
-type Node$Conditional<T: boolean, IfTrue, IfFalse> = T extends true
+type Node$Conditional<T extends boolean, IfTrue, IfFalse> = T extends true
   ? IfTrue
   : T extends false
     ? IfFalse
@@ -282,7 +282,7 @@ type child_process$execFileOpts = Readonly<{
   signal?: AbortSignal,
 }>;
 
-type child_process$execFileCallback<T: string | Buffer> =
+type child_process$execFileCallback<T extends string | Buffer> =
   child_process$execCallback<T>;
 
 type child_process$execFileSyncOpts = Readonly<{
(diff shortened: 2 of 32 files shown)


bacarybruno:
Open question: where should tree shaking live?

Expo has its own experimental tree shaking as a [serializer plugin](https://github.com/expo/expo/blob/e0d97ef10488918e1fe096ad10f350a6d06c138d/packages/%40expo/metro-config/src/serializer/treeShakeSerializerPlugin.ts#L177). It also has constraints that make adopting Metro-core tree shaking non-trivial: custom `transformerPath`, bypasses `Server.build()` entirely via [`_bundleDirectAsync`](https://github.com/expo/expo/blob/e0d97ef10488918e1fe096ad10f350a6d06c138d/packages/%40expo/cli/src/start/server/metro/MetroBundlerDevServer.ts#L1925), and relies on its own serializer pipeline.

Should tree shaking live in Metro core as the source of truth with Expo eventually delegating to it, or should Metro close the door and fully delegate to Expo's implem?

kitten:
@bacarybruno: re

> Should tree shaking live in Metro core as the source of truth with Expo eventually delegating to it, or should Metro close the door and fully delegate to Expo's implem?

We don't currently intend to designate the current, experimental tree-shaking implementation in Expo's core to be stable. It's a proof of concept, and transferring the AST from the transformer to the serialiser, while being the path of least resistance, isn't what we intend to do.

If Metro ships an implementation upstream, we're happy to adopt it and/or tweak as needed to fit it to the existing expectations we need it to comply to.

The main assumption we had was that it'd be nice if a single transformer worker pass could instead operate on source ranges. However, this is potentially a larger change in how our serializers work, but it's one idea we've seen that looks promising. The long and short of it is that we assume we want to keep the serializer lean even with tree-shaking, prevent uncacheable work, and stop transferring ASTs for tree-shaking to work.

PR #1675, opened 2026-04-07
fix: Add ESM unwrapping to all CommonJS require calls

kitten:
## Summary

This was added to `loadConfig.js` but is missing in other places. While it's arguably not useful in many cases in a "standard" Metro setup with Metro's own packages, it'd be nice to have this behave consistently even with transpiled ESM output (or with require-ESM in Node).

Specifically this affects:
- metro-file-map plugin workers that are loaded with `require`
- metro-file-map's main worker loading
- metro's asset plugins
- metro-transform-worker's `getMinifier` implementation
- metro-transform-worker's `babelTransformerPath` loading

This is probably specifically interesting for `babelTransformerPath` which is often overridden by third-party setups.

Changelog: [Fix] Unwrap ES modules when requiring metro-file-map workers, asset plugins, the minifier, or the babel transformer

## Test plan

- n/a: This will likely be best left untested, but is a consistent implementation that matches ESM->CJS semantics, also used in `loadConfig.js`

### DIFF
--- a/packages/metro-file-map/src/plugins/DependencyPlugin.js
+++ b/packages/metro-file-map/src/plugins/DependencyPlugin.js
@@ -76,7 +76,9 @@ export default class DependencyPlugin
     if (this.#dependencyExtractor != null) {
       // Dynamic require to get extractor's cache key
       // $FlowFixMe[unsupported-syntax] - dynamic require
-      const extractor = require(this.#dependencyExtractor);
+      const mod = require(this.#dependencyExtractor);
+      const extractor =
+        mod.__esModule === true && 'default' in mod ? mod.default : mod;
       return JSON.stringify({
         extractorKey: extractor.getCacheKey?.() ?? null,
         extractorPath: this.#dependencyExtractor,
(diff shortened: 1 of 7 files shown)


packages/metro-file-map/src/plugins/DependencyPlugin.js
```diff
@@ -76,7 +76,9 @@ export default class DependencyPlugin
     if (this.#dependencyExtractor != null) {
       // Dynamic require to get extractor's cache key
       // $FlowFixMe[unsupported-syntax] - dynamic require
-      const extractor = require(this.#dependencyExtractor);
+      const mod = require(this.#dependencyExtractor);
+      const extractor =
+        mod.__esModule === true && 'default' in mod ? mod.default : mod;
```
kitten, line 81: Spotted on a re-review that `getCacheKey` below means that we should be doing:

```
const getCacheKey =
  mod?.getCacheKey ??
  (mod.__esModule === true && 'default' in mod ? mod.default : mod).getCacheKey;
```

PR #1680, opened 2026-04-08
refactor: Tweak metro-resolver's Node module resolution hot-path

kitten:
## Summary

The current implementation in `resolve.js` seems slightly optimised for readability over performance, but is a hot-path that is highly impactful for bundling performance. This is slightly amplified depending on the depth of folders.

This branch aims to:
- eliminate redundant `fileSystem.lookup` and instead resolve immediately and return the earliest result
- avoid redundant array allocations and instead iterate target paths directly
- delay building a full array of target `nodeModulesPaths` when a `FailedToResolveNameError` is constructed

While this interleaves the actual resolution, avoiding redundant work and allocations makes up for this.

A very primitive (LLM-authored) benchmark can be found here, which shows an up to 10% benefit or neutral results for resolution: https://github.com/kitten/metro/commit/a36d55825cb0d492d9ea5b7ac5f42b0b8e7b0a77

Changelog: [Internal] Refactor performance sensitive metro-resolver Node module resolution hot path

## Test plan

- CI tests should pass unchanged

### DIFF
--- a/packages/metro-resolver/src/resolve.js
+++ b/packages/metro-resolver/src/resolve.js
@@ -166,26 +166,6 @@ export default function resolve(
 
   const {disableHierarchicalLookup} = context;
 
-  const buildNodeModulePaths = () => {
-    const nodeModulesPaths: string[] = [];
-
-    if (!disableHierarchicalLookup) {
-      let next = path.dirname(originModulePath);
-      let candidate;
-      do {
-        candidate = next;
-        const nodeModulesPath = candidate.endsWith(path.sep)
-          ? candidate + 'node_modules'
-          : candidate + path.sep + 'node_modules';
-        nodeModulesPaths.push(nodeModulesPath);
-        next = path.dirname(candidate);
-      } while (candidate !== next);
-    }
-
-    nodeModulesPaths.push(...context.nodeModulesPaths);
-    return nodeModulesPaths;
-  };
-
   const resolveModuleFromTargetPath = (
     targetPath: string,
   ): Resolution | null => {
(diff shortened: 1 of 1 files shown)


robhogan:
Nice 👌

robhogan:
> optimised for readability over performance

That's very charitable 😅

packages/metro-resolver/src/resolve.js
```diff
@@ -166,70 +166,141 @@ export default function resolve(
...
+        ? nodeModulePath + path.sep + parsedSpecifier.firstPart
+        : nodeModulePath;
+    const lookupResult = context.fileSystemLookup(mustBeDirectory);
+    if (!lookupResult.exists || lookupResult.type !== 'd') {
+      return null;
+    }
+    return resolveModuleFromTargetPath(
+      context,
+      platform,
+      nodeModulePath + path.sep + realModuleName,
```
robhogan, line 190: Not quite, I think, because `realModuleName` is typically posix-separated for a bare specifier and `nodeModulePath` will use system separators. I've tweaked that on the import (`.replaceAll` on Windows only is still faster than a `path.join`)

kitten: makes sense 👍 didn't check whether it was prenormalised

robhogan: Yeah, "normalised" Metro file paths are always system separated, whereas normalised resolution specifiers are posix separated. Tbh the former is a decision I'd revisit if we could - just too many breaking changes and footguns in the way.

(Presumably also the reason Windows itself still uses `\` despite the pain it causes everyone else 😅)

kitten: yea, it's hard to keep track off, but maybe opaque types could help here. Although, on the file-map level it's consistently system separators, so not too hard to keep track off. The boundaries are just harder to keep track off for the resolver. Minor things

Anyway, let me know if you'd like this branch to be updated

robhogan: No need to update, I've made some minor changes on the internal import and it's landing now.

PR #1686, opened 2026-04-11
feat(metro-file-map): Lazily stat files and populate symlinks for Node crawled file trees

kitten:
## Summary

> [!NOTE]
> This is part of the same pick/patch-set from [an experiment on the Expo repo](https://github.com/expo/expo/pull/44567),
> like #1676, #1677, and #1687, for `metro-file-map` (conflicts are expected, since changes were pulled out into separate PRs for clarity)

After #1677 the highest impact on initial crawling performance with the Node crawler are the individual `lstat` calls and `readlink` calls. The former happens for every file that's discovered and can hence get quite large for monorepos (or assets that aren't excluded with blocklists). The latter, readlinks, are expensive if we're in an isolated installation (pnpm & bun), which can contain a much larger amount of symlinks than was likely anticipated by the current code paths.

This is the last (and most minimal iteration) of a few attempts to reduce the cost.

The idea is to skip the `lstat` call for new files that aren't in the previous FS snapshot (entry doesn't exist). To make this safe, these files are also excluded from `getDifference`, i.e. when both the previous and new entries have `0 | null` mtime values. In both paths `0 | null` indicates that the file is skipped for `lstat`.

The `mtime` is later added via an `fs.promises.lstat` call in `getOrComputeSha1` i.e. when the file is actually accessed. If the file isn't accessed the mtime value remains at null.

This cascades to require/allow us to lazily evaluate symlinks for mtime values of null. Their values aren't read until accessed in `#resolveSymlinkTargetToNormalPath` in `TreeFS`. This does introduce an update on read, but lazily populates the symlink value. This trades off the async `readlink` call on all symlinks for a sync `readlink` call on a subset of symlinks.

We've observed this to be an overall win, but this could be influenced further by this commit: https://github.com/expo/expo/commit/af67435c2ebe9c8635e0e1e38c694bd06a6f7847 (which could also be upstreamed but changes the `getContent` signature. This change can be avoided by awaiting first, but seems overall like a better change than awaiting `getContent` unconditionally before plugins)

Changelog: [Internal] Lazily stat Node crawled files and lazily populate symlink targets

## Test plan

- Unit tests were added to demonstrate this
- This can further be tested E2E in [the linked Expo PR](https://github.com/expo/expo/pull/44567), although not in isolation (if necessary I can create a new, isolated branch with a pnpm patch for these changes)

### DIFF
--- a/packages/metro-file-map/src/__tests__/index-test.js
+++ b/packages/metro-file-map/src/__tests__/index-test.js
@@ -176,6 +176,19 @@ jest.mock('fs', () => ({
     error.code = 'ENOENT';
     throw error;
   }),
+  readlinkSync: jest.fn(path => {
+    const entry = mockFs[path];
+    if (!entry) {
+      const error = new Error(`Cannot read path '${path}'.`);
+      // $FlowFixMe[prop-missing] code
+      error.code = 'ENOENT';
+      throw error;
+    }
+    if (typeof entry.link !== 'string') {
+      throw new Error(`Not a symlink: '${path}'.`);
+    }
+    return entry.link;
+  }),
   writeFileSync: jest.fn((path, data, options) => {
     expect(options).toBe(require('v8').serialize ? undefined : 'utf8');
     mockFs[path] = data;
@@ -746,6 +759,50 @@ describe('FileMap', () => {
 
         expect(deepNormalize(await fileMap.read())).toEqual(cacheContent);
       },
+    );
+  });
+
+  test('defers symlink resolution for entries with null mtime', async () => {
+    const node = require('../crawlers/node').default;
+    const fsModule = require('fs');
+
+    // $FlowFixMe[prop-missing]
+    // $FlowFixMe[missing-local-annot]
+    node.mockImplementation(options => {
+      const changedFiles = createMap<FileMetadata>({
+        [path.join('fruits', 'Strawberry.js')]: [32, 42, 0, null, 0, null],
+        [path.join('fruits', 'LinkToStrawberry.js')]: [
+          null,
+          0,
+          0,
+          null,
+          1,
+          null,
+        ],
+      });
+      return Promise.resolve({changedFiles, removedFiles: new Set()});
+    });
+
+    const {fileSystem} = await buildNewFileMap({
+      enableSymlinks: true,
+      useWatchman: false,
+    });
+
+    expect(fsModule.promises.readlink).not.toHaveBeenCalledWith(
+      expect.stringContaining('LinkToStrawberry'),
+    );
+
+    expect(
+      fileSystem.lookup(
+        path.join('/', 'project', 'fruits', 'LinkToStrawberry.js'),
+      ),
+    ).toMatchObject({
+      exists: true,
+      realPath: path.join('/', 'project', 'fruits', 'Strawberry.js'),
+    });
+
+    expect(fsModule.readlinkSync).toHaveBeenCalledWith(
+      path.join('/', 'project', 'fruits', 'LinkToStrawberry.js'),
     );
   });
 
--- a/packages/metro-file-map/src/crawlers/__tests__/integration-test.js
+++ b/packages/metro-file-map/src/crawlers/__tests__/integration-test.js
@@ -80,44 +80,55 @@ function oneOf(this: $FlowFixMe, actual: unknown, ...expectOneOf: unknown[]) {
  * https://fburl.com/gdoc/y8dn025u */
 expect.extend({oneOf});
 
-const CASES = [
-  [
-    true,
-    new Map([
-      ['foo.js', [expect.any(Number), 245, 0, null, 0, null]],
-      [
-        join('directory', 'bar.js'),
-        [expect.any(Number), 245, 0, null, 0, null],
-      ],
-      [
-        'link-to-directory',
-        [expect.any(Number), 9, 0, null, expect.oneOf(1, 'directory'), null],
-      ],
-      [
-        'link-to-foo.js',
-        [expect.any(Number), 6, 0, null, expect.oneOf(1, 'foo.js'), null],
-      ],
-    ]),
-  ],
-  [
-    false,
-    new Map([
-      [
-        join('directory', 'bar.js'),
-        [expect.any(Number), 245, 0, null, 0, null],
-      ],
-      ['foo.js', [expect.any(Number), 245, 0, null, 0, null]],
-    ]),
-  ],
-];
+function getCases(skipsStat: boolean) {
+  const fileEntry = skipsStat
+    ? [null, 0, 0, null, 0, null]
+    : [expect.any(Number), 245, 0, null, 0, null];
+  return [
+    [
+      true,
+      new Map([
+        ['foo.js', fileEntry],
+        [join('directory', 'bar.js'), fileEntry],
+        [
+          'link-to-directory',
+          skipsStat
+            ? [null, 0, 0, null, 1, null]
+            : [
+                expect.any(Number),
+                9,
+                0,
+                null,
+                expect.oneOf(1, 'directory'),
+                null,
+              ],
+        ],
+        [
+          'link-to-foo.js',
+          skipsStat
+            ? [null, 0, 0, null, 1, null]
+            : [expect.any(Number), 6, 0, null, expect.oneOf(1, 'foo.js'), null],
+        ],
+      ]),
+    ],
+    [
+      false,
+      new Map([
+        [join('directory', 'bar.js'), fileEntry],
+        ['foo.js', fileEntry],
+      ]),
+    ],
+  ];
+}
 
 describe.each(Object.keys(CRAWLERS))(
   'Crawler integration tests (%s)',
   crawlerName => {
     const crawl = CRAWLERS[crawlerName];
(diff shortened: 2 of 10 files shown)


kitten:
This (LLM-generated) branch of a squash of this PR and #1687 combined demonstrates the rough performance when both changes are combined: https://github.com/kitten/metro/tree/%40kitten/bench/metro-file-map/crawl-and-lookup

The goal (which seems to be reached, at least in the synthetic tests on the branch) is to keep performance roughly the same when all files are accessed. Performance should improve when a large amount of crawled files are never accessed.

robhogan:
I've merged `main` into this, resolving the conflicts with #1687 and #1693 the way `@expo/metro-file-map` does it: symlinks with a known mtime are still read eagerly, and ones crawled without an `lstat` are read on first traversal, both stored as #1687's lexically resolved normal POSIX targets.

Benchmarked the new head (0bf6640) against `main` (a5d11f7), on an RN 0.87 app installed with pnpm (3,031 symlinks, 42k files). Both built from source, 40 rounds, each running both builds in separate processes in random order. Values are medians ± 95% bootstrap CI, and the change is the median per-round ratio. M5 Pro, Node 22.13.1, so macOS only.

The first ten rows are `metro-file-map` alone: each lookup pass is ~10k lookups through symlinks (the link itself, `<link>/package.json`, a file via the link, and `hierarchicalLookup` for `package.json`), and warm start is deserialising a cached snapshot plus one lookup pass. The replay rows are end to end: all 12,990 `resolveDependency` calls from an iOS dev build of the same app, replayed through each build's own `DependencyGraph`, cold (empty file map cache) and warm (the cache the cold run wrote). The first pass goes through the resolution cache as a build does, and the repeat pass calls the resolver directly. Resolutions were identical across all 160 replays.

| Metric | main | PR | Change | 95% CI of change |
|---|---|---|---|---|
| Crawl + build | 163.7 ± 1.6 ms | 83.0 ± 1.3 ms | **−49.0%** | −49.6% to −48.4% |
| Warm start | 19.5 ± 0.1 ms | 31.9 ± 0.3 ms | **+64.7%** | +62.8% to +65.8% |
| First lookup pass | 21.1 ± 0.2 ms | 37.1 ± 0.2 ms | **+75.4%** | +72.9% to +76.8% |
| Later lookup passes | 13.7 ± 0.1 ms | 13.9 ± 0.1 ms | **+1.2%** | +1.0% to +1.3% |
| Lookup: link itself | 988 ± 11 ns | 999 ± 6 ns | **+1.0%** | +0.7% to +1.5% |
| Lookup: `<link>/package.json` | 1,220 ± 12 ns | 1,229 ± 8 ns | **+0.5%** | +0.1% to +0.9% |
| Lookup: deep file via link | 1,327 ± 9 ns | 1,335 ± 9 ns | **+0.7%** | +0.3% to +1.1% |
| `hierarchicalLookup` | 1,862 ± 13 ns | 1,900 ± 12 ns | **+1.9%** | +1.4% to +2.2% |
| Snapshot size | 1,523 KB | 1,019 KB | **−33.1%** | same every run |
| Heap after first pass | 24.16 MB | 23.46 MB | **−2.9%** | same every run |
| Replay, cold: startup | 222.8 ± 2.7 ms | 129.8 ± 1.3 ms | **−41.8%** | −42.9% to −41.1% |
| Replay, cold: first pass | 209.3 ± 1.0 ms | 211.2 ± 1.2 ms | **+0.8%** | +0.3% to +1.4% |
| Replay, cold: startup + first pass | 432.7 ± 2.2 ms | 340.8 ± 2.6 ms | **−21.2%** | −21.7% to −20.7% |
| Replay, cold: repeat pass | 177.1 ± 0.8 ms | 176.8 ± 0.9 ms | −0.1% | −0.4% to +0.4% |
| Replay, warm: startup | 216.0 ± 3.0 ms | 139.9 ± 1.4 ms | **−35.1%** | −35.9% to −34.5% |
| Replay, warm: first pass | 203.7 ± 1.4 ms | 204.3 ± 1.7 ms | **+0.5%** | +0.2% to +1.1% |
| Replay, warm: startup + first pass | 419.7 ± 3.8 ms | 344.5 ± 2.2 ms | **−17.5%** | −18.3% to −17.2% |
| Replay, warm: repeat pass | 171.6 ± 0.7 ms | 171.9 ± 0.8 ms | +0.0% | −0.2% to +0.3% |

Bold where the CI excludes zero.

robhogan:
AI:

---

I split this into its parts to see where the numbers above come from. Same app, benchmarks and stats as above, five builds from the merged head, 30 rounds with all five in random order per round. All 300 replays resolved identically.

- `main` - a5d11f7
- `hoist` - the extension check moved before `lstat` (part of this PR's crawler change), still `lstat`ing every match, symlinks read at crawl time
- `lazystat` - this PR's crawler, symlinks read at crawl time
- `lazylink` - `main`'s crawler, symlinks read on first traversal
- `tip` - this PR

Changes are the median per-round delta against `main`, with 95% bootstrap CI. Bold where the CI excludes zero.

| | main | hoist | lazystat | lazylink | tip |
|---|---|---|---|---|---|
| `lstat` per start | 53,081 | 36,545 | 0 | 53,081 | 0 |
| `readlink` per start | 3,031 async | 3,031 async | 3,031 async | 132 sync | 132 sync |
| Crawl + build | 163.3 ms | **−16.1 ms** [−10.5%, −7.2%] | **−69.5 ms** [−43.6%, −41.5%] | **−16.1 ms** [−10.1%, −7.9%] | **−81.2 ms** [−50.0%, −48.8%] |
| Replay, cold: startup + first pass | 430.0 ms | **−10.8 ms** [−3.9%, −1.9%] | **−74.0 ms** [−17.7%, −16.7%] | **−6.8 ms** [−2.2%, −0.9%] | **−89.5 ms** [−21.4%, −20.2%] |
| Replay, warm: startup + first pass | 415.0 ms | **+13.1 ms** [+2.4%, +3.7%] | **−72.5 ms** [−18.5%, −17.0%] | −2.5 ms [−1.7%, +0.4%] | **−72.1 ms** [−18.3%, −16.8%] |
| Replay, cold: first pass | 209.7 ms | −2.3 ms [−2.4%, +0.5%] | −0.8 ms [−1.0%, +0.5%] | **+2.0 ms** [+0.1%, +1.5%] | +1.6 ms [−0.1%, +1.2%] |
| Warm start | 19.6 ms | **+0.1 ms** [+0.2%, +1.1%] | +0.1 ms [−0.5%, +1.2%] | **+12.9 ms** [+65.1%, +67.7%] | **+12.4 ms** [+61.5%, +66.3%] |
| First lookup pass | 21.0 ms | **+0.4 ms** [+0.7%, +3.3%] | **+0.5 ms** [+1.2%, +3.4%] | **+15.8 ms** [+74.1%, +78.5%] | **+16.3 ms** [+75.3%, +80.8%] |
| Later lookup passes | 13.8 ms | **+1.0%** [+0.5%, +1.6%] | **+1.3%** [+0.7%, +1.6%] | −0.2% [−0.5%, +0.4%] | **+1.2%** [+0.7%, +1.6%] |
| Snapshot size | 1,523 KB | 0 | −17.2% | −15.9% | −33.1% |

- **Lazy stat is almost all of it.** Of the −90 ms cold end to end, lazy stat is −74 ms (of which the hoist is −11 ms) and lazy symlinks −7 ms. Warm, it's all lazy stat (−73 ms) - lazy symlinks are within noise.
- In `metro-file-map` alone the build saving splits roughly into hoist −16 ms, the rest of lazy stat −53 ms, and lazy symlinks −16 ms.
- **Lazy symlinks trade 3,031 async reads at crawl for 132 sync reads during resolution** - only the links on this build's resolution paths. That's +2 ms on the cold first pass, ~15µs a read.
- Lazy symlinks are also the whole of the `metro-file-map` warm start and first pass regressions (+13-16 ms). The snapshot is written before any link is read, so every new process reads them again, and those benchmarks traverse every link. End to end it's within noise.
- The +1-2% on repeat lookups comes with the crawler change (`hoist`, `lazystat`, `tip`), not lazy symlinks, even though lookup code is the same in all four. I haven't pinned down why - possibly tree layout.
- The snapshot saving splits about evenly between null mtimes/sizes and unstored symlink targets.
- `hoist` is 13 ms slower warm than `main` despite 16.5k fewer `lstat`s. That's the crawler's per-file `getMtimeByNormalPath`, a full lookup from the root for every crawled file once the previous tree is populated. Removing just that call from `hoist` saves 29.0 ms on warm startup (−12.5%, CI −14.9% to −10.5%; 16 rounds), and this PR pays the same cost on every warm start.

robhogan:
High level thoughts having looked closely at this and rebased on two related, now merged ones:

 - Lazy statting is worth doing, but since statting isn't relevant to resolution or the core function of TreeFS, I think we can lift stats into a lazy plugin, alongside SHA1, probably after https://github.com/react/metro/pull/1950. At a minimum, it needs to be through an abstraction compatible with custom file systems and should use the same `processFile` path as SHA1. 
 - Lazy symlinks are of less clear value - the time we save during start up is paid (and then some) during the first build, which is arguably worse (a bundle request is more often on a user's critical path). It also unavoidably brings synchronous I/O into lookups.
 
 Parking this for now, but hopefully we can get the benefits of lazy statting soon.

PR #1687, opened 2026-04-11
refactor(metro-file-map): Pre-resolve symlink targets and store normal posix paths

kitten:
## Summary

> [!NOTE]
> This is part of the same pick/patch-set from [an experiment on the Expo repo](https://github.com/expo/expo/pull/44567),
> like #1676, #1677, and #1686, for `metro-file-map` (conflicts are expected, since changes were pulled out into separate PRs for clarity)

The symlinks have two oddities to them compared to other file data operations and entries:
- a separate `WeakMap` cache is kept for resolved symlink targets that have been converted
- the values stored for symlink targets may be absolute paths and are unnormalised system paths 

The `WeakMap` (in theory) is inefficient especially for projects with a larger amount of symlinks. It's also not necessary since the values can be trivially pre-resolved to a normal path. We can additionally store this value as a posix normal path, to make sure that it's identical between systems, which may help testing.

Edit: A primitive (LLM-generated) benchmark that shows that performance remains within margins with this change can be found at https://github.com/kitten/metro/commit/eee7a821ae3c17fca994ca035b8f41f2cb15faa1

Changelog: [Internal] Pre-resolve symlinks and store normalized symlink targets in file map

## Test plan

- Unit tests were added to demonstrate the changes
- This can further be tested E2E in [the linked Expo PR](https://github.com/expo/expo/pull/44567), although not in isolation (if necessary I can create a new, isolated branch with a pnpm patch for these changes)

### DIFF
--- a/packages/metro-file-map/src/lib/TreeFS.js
+++ b/packages/metro-file-map/src/lib/TreeFS.js
@@ -21,6 +21,7 @@ import type {
 } from '../flow-types';
 
 import H from '../constants';
+import normalizePathSeparatorsToSystem from './normalizePathSeparatorsToSystem';
 import {RootPathUtils} from './RootPathUtils';
 import invariant from 'invariant';
 import path from 'path';
@@ -1212,11 +1213,12 @@ export default class TreeFS implements MutableFileSystem {
     symlinkNode: FileMetadata,
     canonicalPathOfSymlink: Path,
   ): NormalizedSymlinkTarget {
-    const symlinkTarget = symlinkNode[H.SYMLINK];
+    let symlinkTarget = symlinkNode[H.SYMLINK];
     invariant(
       typeof symlinkTarget === 'string',
       'Expected symlink target to be populated.',
     );
+    symlinkTarget = normalizePathSeparatorsToSystem(symlinkTarget);
     const result = {
       ancestorOfRootIdx: this.#pathUtils.getAncestorOfRootIdx(symlinkTarget),
       normalPath: symlinkTarget,
--- a/packages/metro-file-map/src/lib/TreeFS.js
+++ b/packages/metro-file-map/src/lib/TreeFS.js
@@ -738,13 +738,13 @@ export default class TreeFS implements MutableFileSystem {
         // with the remaining path results in collapsing segments, e.g:
         // '../..' + 'parentofroot/root/foo.js' = 'foo.js', then we must add
         // parentofroot and root as ancestors.
-        ancestorOfRootIdx =
-          this.#pathUtils.getAncestorOfRootIdx(normalSymlinkTarget);
         if (
           collectAncestors &&
           !isLastSegment &&
           // No-op optimisation to bail out the common case of nothing to do.
-          (ancestorOfRootIdx === 0 || joinedResult.collapsedSegments > 0)
+          ((ancestorOfRootIdx =
+            this.#pathUtils.getAncestorOfRootIdx(normalSymlinkTarget)) === 0 ||
+            joinedResult.collapsedSegments > 0)
         ) {
           let node: MixedNode = this.#rootNode;
           let collapsedPath = '';
--- a/packages/metro-file-map/src/lib/TreeFS.js
+++ b/packages/metro-file-map/src/lib/TreeFS.js
@@ -38,12 +38,6 @@ function isRegularFile(node: FileNode): boolean {
   return node[H.SYMLINK] === 0;
 }
 
-type NormalizedSymlinkTarget = {
-  ancestorOfRootIdx: ?number,
-  normalPath: string,
-  startOfBasenameIdx: number,
-};
-
 type DeserializedSnapshotInput = {
   rootDir: string,
   fileSystemData: DirectoryNode,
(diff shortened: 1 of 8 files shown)


packages/metro-file-map/src/lib/TreeFS.js
```diff
@@ -1214,33 +1213,17 @@ export default class TreeFS implements MutableFileSystem {
     symlinkNode: FileMetadata,
     canonicalPathOfSymlink: Path,
   ): NormalizedSymlinkTarget {
-    const cachedResult = this.#cachedNormalSymlinkTargets.get(symlinkNode);
-    if (cachedResult != null) {
-      return cachedResult;
-    }
```
robhogan, line 1220: On the face of it, this looks like it could be a perf regression, because we're introducing a new allocation and some string manipulation every time the same symlink is traversed, vs once per symlink.

I'd be interested to see a benchmark on a symlink-heavy project. Weakmap lookups should generally be very fast.

packages/metro-file-map/src/lib/TreeFS.js
```diff
@@ -1214,33 +1213,17 @@ export default class TreeFS implements MutableFileSystem {
...
-    );
+    symlinkTarget = normalizePathSeparatorsToSystem(symlinkTarget);
     const result = {
-      ancestorOfRootIdx:
-        this.#pathUtils.getAncestorOfRootIdx(normalSymlinkTarget),
-      normalPath: normalSymlinkTarget,
-      startOfBasenameIdx: normalSymlinkTarget.lastIndexOf(path.sep) + 1,
+      ancestorOfRootIdx: this.#pathUtils.getAncestorOfRootIdx(symlinkTarget),
+      normalPath: symlinkTarget,
+      startOfBasenameIdx: symlinkTarget.lastIndexOf(path.sep) + 1,
```
robhogan, line 1225: If we're not caching, we may be better off computing some of this lazily (and eg, just return the `normalPath` here).

packages/metro-file-map/src/lib/TreeFS.js
kitten: this was admittedly tied to the on-demand file system change, which necessitated changes here anyway. It then made sense to pre-normalise values which ended up being a net neutral change (performance-wise) across all combined changes (!)

I can give benchmarking this in isolation a go next week 👍

Together with the lazy symlink resolution I would expect this to be mainly a neutral change since the old code paths assumed a low amount of symlinks. But I don't have an intuition of what this would look like in a project with many symlinks where the cost isn't negligible. My intuition overall with this change was that the cost of `WeakMap` lookups can add up faster than computing the path once, provided symlink resolution is lazier than it is now

kitten: This seems to show that it's mostly neutral: https://github.com/kitten/metro/commit/eee7a821ae3c17fca994ca035b8f41f2cb15faa1

Intuitively, the same amount of work happens (at least for macOS), since the path is prenormalised and stored in the file tree data structure, which means that the work (which was negligible to begin with) is now happening earlier and makes the `WeakMap` obsolete.

I didn't test this on Windows, where presumably there's a small additional cost due to the path normalisation (win -> posix -> win), which keeps the file map data structure portable. We could simply skip this and keep this denormalised, but the parity is nice imo

packages/metro-file-map/src/lib/TreeFS.js
kitten: Removed the intermediate data structure in https://github.com/facebook/metro/pull/1687/changes/0a6435120887217ecc560ff35f7a5272cee6a9b8

robhogan:
Benchmarked the current head (e6e45d3) against its merge base on `main` (3173de1), on an RN 0.87 app installed with pnpm (3,030 symlinks, 42k files). Both built from source, 40 rounds, each running both builds in separate processes in random order. Values are medians ± 95% bootstrap CI, and the change is the median per-round ratio. M5 Pro, Node 22.13.1, so macOS only.

Each lookup pass is 10,480 lookups through symlinks (the link itself, `<link>/package.json`, a file via the link, and `hierarchicalLookup` for `package.json`). Warm start is deserialising a cached snapshot plus one lookup pass.

| Metric | main | PR | Change | 95% CI of change |
|---|---|---|---|---|
| Crawl + build | 157.3 ± 2.3 ms | 161.4 ± 1.5 ms | **+2.6%** | +1.5% to +3.8% |
| Warm start | 22.4 ± 0.2 ms | 19.0 ± 0.2 ms | **−15.6%** | −16.1% to −14.3% |
| First lookup pass | 24.6 ± 0.2 ms | 20.9 ± 0.2 ms | **−15.2%** | −16.1% to −14.4% |
| Later lookup passes | 13.4 ± 0.1 ms | 13.4 ± 0.0 ms | +0.3% | −0.0% to +0.6% |
| Lookup: link itself | 962 ± 4 ns | 967 ± 5 ns | +0.6% | +0.0% to +1.1% |
| Lookup: `<link>/package.json` | 1,185 ± 6 ns | 1,194 ± 5 ns | +0.5% | −0.0% to +1.3% |
| Lookup: deep file via link | 1,287 ± 9 ns | 1,295 ± 7 ns | +0.2% | −0.3% to +0.5% |
| `hierarchicalLookup` | 1,819 ± 8 ns | 1,815 ± 6 ns | −0.4% | −0.7% to +0.2% |
| Snapshot size | 1,485 KB | 1,522 KB | **+2.6%** | same every run |
| Heap after first pass | 25.42 MB | 24.14 MB | **−5.0%** | same every run |

Bold where the CI excludes zero.

robhogan (approved):

PR #1690, opened 2026-04-16
Replace chalk with Node core util.styleText

stianjensen:
## Summary

All supported Node versions in `engines` ship `util.styleText` with the array-format API, removing the need for the chalk dependency.


robhogan (approved):
Thanks!

PR #1692, opened 2026-04-20
...

retyui:
...

### DIFF
--- a/packages/metro-transform-plugins/src/utils/createInlinePlatformChecks.js
+++ b/packages/metro-transform-plugins/src/utils/createInlinePlatformChecks.js
@@ -22,13 +22,20 @@ type PlatformChecks = {
     node: MemberExpression,
     scope: Scope,
     isWrappedModule: boolean,
+    filename?: string,
   ) => boolean,
   isPlatformSelectNode: (
     node: CallExpression,
     scope: Scope,
     isWrappedModule: boolean,
+    filename?: string,
   ) => boolean,
 };
+
+const REACT_NATIVE_MODULES_REGEX = /[\/\\]node_modules[\/\\]react-native[\/\\]/;
+
+const isReactNativeFile = (filename?: string): boolean =>
+  filename != null && REACT_NATIVE_MODULES_REGEX.test(filename);
 
 export default function createInlinePlatformChecks(
   t: Types,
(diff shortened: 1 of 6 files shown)


retyui:
@robhogan could you please review this PR ?

robhogan:
Stop spamming me, @retyui, please.

retyui:
Oops, sorry Rob for the notifications! I'll ease up. Hope you're having a good day otherwise

packages/metro-transform-plugins/src/utils/createInlinePlatformChecks.js
```diff
@@ -148,6 +148,32 @@ export default function createInlinePlatformChecks(
...
+    // import Platform from '../../Utilities/Platform';
+    // const test = Platform.OS === 'ios' ? 1 : 2;
+    //
+    // 2. After `@react-native/babel-preset`
+    // var _Platform = _interopRequireDefault(require("../../Utilities/Platform"));
+    // var test = _Platform.default.OS === 'ios' ? 1 : 2;
+    // ```
+    if (!identifier && isMemberExpression(node)) {
+      const objIdentifier = patterns.find((pattern: {name: string}) =>
+        isIdentifier(node.object, {name: `_${pattern.name}`}),
```
javache, line 192: We shouldn't assume the `_` prefix. Instead, can we trace whether the variable came from an `_interopRequireDefault` expression?

javache (changes requested):

robhogan:
I think, fundamentally, matching on a relative import without the context that we're inside the `react-native` package is too prone to false positives.

Since this only affects `react-native`'s own source (and forks), IMO it'd be better to fix there by rewriting those imports from relative paths to self-referencing package imports. I added self-referencing import support to Metro last week, not yet released, partly to facilitate that.

rubennorte:
> IMO it'd be better to fix there by rewriting those imports from relative paths to self-referencing package imports

This is very unidiomatic tbh and we actually have lint rules to prevent this.

Can we not know the path of the module being transformed to detect we're inside react-native?

packages/metro-transform-plugins/src/utils/createInlinePlatformChecks.js
retyui: I have to add prefix as `isImportOrGlobal` is a common util that uses provided patterns `[{name: 'Platform'}]` to find Platform identifiers that were renamed 

```jsx
var test = _Platform.default.OS === 'ios' ? 'ios' : 'not-ios';
                // ^^^ isImportOrGlobal(..., [{name: 'Platform'}]) // won't works 
```

stacktrace: 
```tsx
<MemberExpression|CallExpression>(path: NodePath, state: State)
  isPlatformNode(node, scope, !!opts.isWrapped)
    isPlatformOS(node, scope, isWrappedModule)
      isImportOrGlobal(node.object, scope, [{name: 'Platform'}], isWrappedModule);
```

robhogan:
> > IMO it'd be better to fix there by rewriting those imports from relative paths to self-referencing package imports
> 
> 
> 
> This is very unidiomatic tbh and we actually have lint rules to prevent this.

Yeah I don't love it either 😅. It's not idiomatic, and it'd be a bad idea in general to import via a barrel file - this is a bit of a special case as Metro transforms the import away but it is still ugly. To solve for consumers of RN other than Meta it *could* be a pre-publish rewrite rather than something that lives in source.

> Can we not know the path of the module being transformed to detect we're inside react-native?

Potentially, yeah - the `projectRoot`-relative path is a cache-safe transform input. It's not so easy to cover using RN directly from source in that case though.

Another ugly option is to guess that we're inside RN based on the copyright header or `@oncall react_native`. 

Or, invent an explicit compiler hint and annotate the import.

packages/metro-transform-plugins/src/utils/createInlinePlatformChecks.js
```diff
@@ -148,6 +166,39 @@ export default function createInlinePlatformChecks(
...
+    // import Platform from '../../Utilities/Platform';
+    // const test = Platform.OS === 'ios' ? 1 : 2;
+    //
+    // 2. After `@react-native/babel-preset`
+    // var _Platform = _interopRequireDefault(require("../../Utilities/Platform"));
+    // var test = _Platform.default.OS === 'ios' ? 1 : 2;
+    // ```
+    if (
+      filename != null &&
+      filename.includes('/node_modules/react-native/') &&
```
robhogan: IIRC (worth verifying) this would need to use Windows path separators on Windows.

packages/metro-transform-plugins/src/utils/createInlinePlatformChecks.js
javache: You can't just look at the variable name, you need to trace where it was bound, what import it resolved, and whether that matches the known path for the react-native Platform module.

Don't really need to know whether you're inside node_modules/react-native or not, you just need to evaluate the relative path.

retyui: you right, but current implementation doesn't do it too, it's just search for `Platform.OS` or `Platform.select()` expressions  and that a `Platform` identifier has a top level binding 

```tsx
var MyPlatform = require('react-native').Platform;
MyPlatform.select({ios: 1, android: 2});
// ^^^ if you change the variable name, plugin won'r work
```

javache: if we make changes here, we should do better than the current implementation, as it may have unintended side-effects in how the core code is compiled. Let's make sure we trace the binding and only transform those.

javache (changes requested):

PR #1694, opened 2026-04-22
fix(metro): Fix regression to allow scale assets to be resolved again for single asset requests

### LINKED ISSUE #1667: Error back-icon@3x.png could not be found
Cross-reported to react-navigation
https://github.com/react-navigation/react-navigation/issues/13023

### Current behavior

I'm getting the following error from Metro v0.83.5 while running a development react-native build locally on my Pixel 9a, and the back arrow icon normally shown in the upper-left is not appearing:
```
Error: 'node_modules/@react-navigation/elements/lib/module/assets/back-icon@3x.png' could not be found, because it is not within the projectRoot or watchFolders, or it is blocked via the resolver.blockList config
    at getAsset (HeaderBackMissing\node_modules\metro\src\Assets.js:226:13)
    at Server._processSingleAssetRequest (HeaderBackMissing\node_modules\metro\src\Server.js:443:47)
    at async Server._processRequest (HeaderBackMissing\node_modules\metro\src\Server.js:550:7)
```

It works ok from a production release build for both Android and iOS, so I suspect this is related to Metro development only.

This started around 25-Feb-2026 after updating from react-native v0.84.0 to v.0.84.1 along with all of the react-navigation dependencies.

Our app uses @react-navigation/elements@2.9.10, however we specifically extract the `<Header ...>` portion of the `<NativeStackView>` component in order to customize the header.
https://github.com/react-navigation/react-navigation/blob/%40react-navigation/elements%402.9.10/packages/native-stack/src/views/NativeStackView.tsx#L116

I've included a reproduction Snack, however the issue is only visible if you use the included App.tsx and NativeStackHeader.tsx locally (possibly on this specific Android device), here's a screen capture:
https://snack.expo.dev/@jalbertsyncrotech/headerbackmissing?platform=android
Image

### Expected behavior

The back icon should be shown in the upper left of a development build when using the @react-navigation/elements NativeStackView Header on a Pixel 9a.

### Reproduction

https://snack.expo.dev/@jalbertsyncrotech/headerbackmissing?platform=android

### Platform

- [x] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS

### Packages

- [ ] @react-navigation/bottom-tabs
- [ ] @react-navigation/drawer
- [ ] @react-navigation/material-top-tabs
- [ ] @react-navigation/stack
- [x] @react-navigation/native-stack
- [ ] react-native-drawer-layout
- [ ] react-native-tab-view

### Environment

- [x] I've removed the packages that I don't use

| package | version |
| -------------------------------------- | ------- |
| @react-navigation/elements | 2.9.10 |
| @react-navigation/native | 7.1.33 |
| @react-navigation/native-stack | 7.14.4 |
| react-native-screens | 4.16.0 |
| react-native-safe-area-context | 5.6.0 |
| react-native | 0.84.1 |
| node | 22.22.0 |
| yarn | 1.22.22 |

kitten:
## Summary

Resolves #1667

The prior change checks the existence of the file path too soon, and ignores that scales may be resolved later in this function.

> [!NOTE]
> Unrelated to this change, the last change feels a bit ad-hoc and incomplete. We should replace the `fs.promises.readdir` call with the file map. That's out of scope for this PR though.

Changelog: [Fix] Fix regression to allow single asset request to resolve scaled assets

## Test plan

- Unit tests added

### DIFF
--- a/packages/metro/src/Assets.js
+++ b/packages/metro/src/Assets.js
@@ -299,29 +299,38 @@ export async function getAsset(
   }
 
   // NOTE: If fileExistsInFileMap is not provided, we fall back to pathBelongsToRoots for backward compatibility, as getAsset is part of the public API.
-  if (fileExistsInFileMap != null) {
-    if (!fileExistsInFileMap(absolutePath)) {
-      throw new Error(
-        `'${relativePath}' could not be found, because it is not within the projectRoot or watchFolders, or it is blocked via the resolver.blockList config`,
-      );
-    }
-  } else {
-    if (!pathBelongsToRoots(absolutePath, [projectRoot, ...watchFolders])) {
-      throw new Error(
-        `'${relativePath}' could not be found, because it cannot be found in the project root or any watch folder`,
-      );
-    }
+  if (
+    fileExistsInFileMap == null &&
+    !pathBelongsToRoots(absolutePath, [projectRoot, ...watchFolders])
+  ) {
+    throw new Error(
+      `'${relativePath}' could not be found, because it cannot be found in the project root or any watch folder`,
+    );
   }
 
   const record = await getAbsoluteAssetRecord(absolutePath, platform ?? null);
 
   for (let i = 0; i < record.scales.length; i++) {
     if (record.scales[i] >= assetData.resolution) {
+      if (
+        fileExistsInFileMap != null &&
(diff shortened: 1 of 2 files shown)


robhogan:
Thanks!

PR #1696, opened 2026-04-23
Fix: Resolving normalPath when it belongs to a separate drive location (as absolute path)

MxKevinBeqo:
## Summary

On Windows, path.relative() returns an absolute path when source and target are on different drives (e.g. Z:\ -> C:\...). In that case the stored "normal path" is already absolute, so return it as-is instead of prepending rootDir which would produce Z:\C:\... paths.

Changelog: [Fix] Fix resolution to absolute paths when the normalPath is an absolute path in another drive.

## Test plan

A new test case is added. When running the test case without the added fix, it will fail, showing the incorrect path resolution. If we run the test case after the fix is applied, the test case passes.

### DIFF
--- a/packages/metro-file-map/src/lib/RootPathUtils.js
+++ b/packages/metro-file-map/src/lib/RootPathUtils.js
@@ -150,6 +150,9 @@ export class RootPathUtils {
     if (right.length === 0) {
       return left;
     }
+    if (path.isAbsolute(right)) {
+      return right;
+    }
     // left may already end in a path separator only if it is a filesystem root,
     // '/' or 'X:\'.
     if (i === this.#rootDepth) {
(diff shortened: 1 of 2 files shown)


robhogan:
Hi @MxKevinBeqo - have you observed a bug due to this?

> the stored "normal path" is already absolute

This should not be the case. Metro represents normal paths on Windows as if drives were top level directories, ie if `projectRoot` is `C:\Foo`, `D:\Bar` is represented as `..\..\D:\Bar`.

`metro-file-map` has a single root node.

MxKevinBeqo:
Hey @robhogan , thank you for the quick response.

I did encounter an issue when the project I was building was in drive C and my node_modules were in another drive (e.g. D).
- I tried both cases even with configuring `projectRoot` explicitly, even without.
- I have added the proper path of the `node_modules` in the `watchFolders`.

I was then getting an error where some package files could not be found on a path `C:\D:\Projects...`. When I get some free time I will reproduce it though and post the actual error log because unfortunately my logs are cleared and will have to redo some configuration to reproduce it.

----

By the way, since you mentioned it, is my scenario possible using metro-s default resolver and only by adding the dependencies in the `watchFolder` ? The scenario being `I was building was in drive C and my node_modules were in another drive (e.g. D)`:

kitten:
@MxKevinBeqo: Hiya :wave: I've filed #1711 with some targeted fixes for this. If you have a project that reproduces this issue, it'd be really helpful if you could give the changes a quick runthrough, if it's not too much work. Cheers! ❤️

PR #1711, opened 2026-05-11
fix(metro-file-map): Fix Windows cross-device path resolution

kitten:
## Summary

Related PR for `expo/expo` fork: https://github.com/expo/expo/pull/45648
Supersedes #1696

Cross-device path normals are intended to be relative paths from the rootDir. If the rootDir is at `C:\project` then a cross-device normal path at `D:\temp` is represented as `..\..\D:\temp`. This assumption is broken in several places, like `normalToAbsolute` printing, leading to `C:\D:\` paths that are invalid.

Changelog: [Fix] Fix Windows path handling for cross-drive paths

## Test plan

- Unit tests were adjusted to demonstrate all failing cases
- **Note:** Invalid unit tests have been removed for paths that rise above the filesystem root (they're assumed to be impossible, and their output doesn't have to be deterministic/correct)

### DIFF
--- a/packages/metro-file-map/src/lib/RootPathUtils.js
+++ b/packages/metro-file-map/src/lib/RootPathUtils.js
@@ -152,10 +152,9 @@ export class RootPathUtils {
     const right = pos === 0 ? normalPath : normalPath.slice(pos);
     if (right.length === 0) {
       return left;
-    } else if (pos > this.#rootDepth * UP_FRAGMENT_SEP_LENGTH) {
-      // When we walk above the filesystem root, emit the remaining path as
-      // is. This is important on Windows, since we canonicalize cross-device
-      // paths as relative paths from rootDir.
+    } else if (IS_WIN32 && pos > this.#rootDepth * UP_FRAGMENT_SEP_LENGTH) {
+      // On Windows, up-fragments at the root encode target on anoher drive
+      // (e.g. '..\..\D:\file'), but on POSIX '..' re-enters the root
       return right;
     }
     // left may already end in a path separator only if it is a filesystem root,
--- a/packages/metro-file-map/src/lib/RootPathUtils.js
+++ b/packages/metro-file-map/src/lib/RootPathUtils.js
@@ -153,8 +153,10 @@ export class RootPathUtils {
     if (right.length === 0) {
       return left;
     } else if (IS_WIN32 && pos > this.#rootDepth * UP_FRAGMENT_SEP_LENGTH) {
-      // On Windows, up-fragments at the root encode target on anoher drive
-      // (e.g. '..\..\D:\file'), but on POSIX '..' re-enters the root
+      // On a real file system, navigating to `..` at the top level (posix `/`
+      // or Windows drive) is a no-op, but we can't respect that on Windows
+      // because Metro uses e.g. `..\..\D:\foo` to represent cross-drive
+      // relative paths.
       return right;
     }
     // left may already end in a path separator only if it is a filesystem root,
--- a/packages/metro-file-map/src/lib/RootPathUtils.js
+++ b/packages/metro-file-map/src/lib/RootPathUtils.js
@@ -149,6 +149,11 @@ export class RootPathUtils {
     const right = pos === 0 ? normalPath : normalPath.slice(pos);
     if (right.length === 0) {
       return left;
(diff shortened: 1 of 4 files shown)


robhogan:
Thanks again for looking at this!

> Invalid unit tests have been removed for paths that rise above the filesystem root (they're assumed to be impossible, and their output doesn't have to be deterministic/correct)

This is very edge-casey but the behaviour of `..` at the root *is* actually specified (at least on posix), those test cases weren't invalid or accidental.

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap04.html#tag_04_13

> The special filename dot shall refer to the directory specified by its predecessor. The special filename dot-dot shall refer to the parent directory of its predecessor directory. **As a special case, in the root directory, dot-dot may refer to the root directory itself.**

In particular, if you create a symlink with the target `../../../../file-at-root.txt` at `/`, it will deference to `/file-at-root.txt` on posix, so it should with Metro as well.

robhogan:
(Also, nit: device != drive. These are cross-drive paths)

kitten:
interesting! that's an easy fix, but requires us to use `startsWithDriveLetter` (as per https://github.com/expo/expo/pull/45648) in one place, I believe, so it's a bit of a trade-off. I didn't measure performance in that case, but I think it might not be otherwise possible to differentiate between the Windows/Posix root case otherwise. Although, I could be mistaken and there could be an easier way, if we ignore this case for Windows.

packages/metro-file-map/src/lib/TreeFS.js
```diff
@@ -1230,10 +1240,18 @@ export default class TreeFS implements MutableFileSystem {
...
       absoluteSymlinkTarget,
     );
+    // path.relative leaves cross-device targets absolute on Windows (e.g.
+    // 'D:\\file' from a 'C:\\' root). Canonicalise those via RootPathUtils so
+    // they round-trip as a '..'-chain from rootDir.
+    if (path.isAbsolute(normalSymlinkTarget)) {
+      normalSymlinkTarget = this.#pathUtils.absoluteToNormal(
+        absoluteSymlinkTarget,
+      );
+    }
```
robhogan: Any reason we need two different paths here? Why not...

```suggestion
    const normalSymlinkTarget = this.#pathUtils.absoluteToNormal(absoluteSymlinkTarget);
```

packages/metro-file-map/src/lib/RootPathUtils.js
```diff
@@ -149,6 +152,10 @@ export class RootPathUtils {
     const right = pos === 0 ? normalPath : normalPath.slice(pos);
     if (right.length === 0) {
       return left;
+    } else if (IS_WIN32 && pos > this.#rootDepth * UP_FRAGMENT_SEP_LENGTH) {
+      // On Windows, up-fragments at the root encode target on anoher drive
+      // (e.g. '..\..\D:\file'), but on POSIX '..' re-enters the root
+      return right;
     }
```
robhogan, line 161: I think this is not really a fundamental difference between OSes and more a limitation due to Metro's weird (sorry) cross-drive path encoding.

It we *could* maintain "`..` goes nowhere at the root" we would, but we can't.

```suggestion
    } else if (IS_WIN32 && pos > this.#rootDepth * UP_FRAGMENT_SEP_LENGTH) {
      // On a real file system, navigating to `..` at the top level (posix `/`
      // or Windows drive) is a no-op, but we can't respect that on Windows
      // because Metro uses e.g. `..\..\D:\foo` to represent cross-drive
      // relative paths.
      return right;
    }
```

packages/metro-file-map/src/lib/TreeFS.js
kitten: No reason. I think I got confused while converting this from the Expo-fork's version, where we don't go through `path.resolve` or `path.relative` anymore. This looks correct to me

kitten: Actually, this seems to be failing tests. I think in Expo it doesn't because we manually normalise `target[len - 1] === path.sep` with a `slice(0, -1)` for symlink targets, which `path.relative` here does for us

kitten: I think realistically, an `endsWith(path.sep)` check here is simplest, if we don't want to replicate the `path.resolve` replacement here but drop `path.relative`. There's probably other options, but they may be more intrusive

robhogan (approved):

PR #1713, opened 2026-05-13
chore: declare contents: read on nightly-tests workflow

arpitjain099:
The nightly tests workflow runs the test suite on a cron schedule. Pure CI: checkout + yarn install + jest. No git push, no API write. Other workflows in this repo already declare permissions.

### DIFF
--- a/.github/workflows/nightly-tests.yml
+++ b/.github/workflows/nightly-tests.yml
@@ -40,6 +40,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: [test]
     if: ${{ always() && needs.test.result == 'failure' }}
+    permissions:
+      contents: read
+      issues: write
     steps:
       - uses: actions/github-script@v8
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-tests.yml
```diff
@@ -12,6 +12,9 @@ on:
     # Daily at at 5:00 UTC
     - cron:  '0 5 * * *'
 
+permissions:
+  contents: read
```
CalixTang, line 16: we may also need write permissions for `issues`. 
The `comment-on-pr-for-failures` job in this workflow needs to be able to write, so locking down to `contents: read` isn't permissive enough.

CalixTang (changes requested):

arpitjain099:
Good catch @CalixTang, thanks. Pushed a per-job permissions block on `comment-on-pr-for-failures` (`contents: read` + `issues: write`) so the rest of the workflow stays at the workflow-level `contents: read` and only the comment-posting job gets the wider scope. Let me know if you'd prefer the write on the workflow level instead.

PR #1734, opened 2026-06-14
Collect dependencies from createRequire-bound require calls

jeferson-sb:
## Summary
 ESM packages use `const require = createRequire(import.meta.url)` to emulate CJS require. Previously, `collectDependencies` skipped these calls because `require` had a local binding. This change detects when the binding comes from a factory call

Upstream issue from Expo: [46129](https://github.com/expo/expo/issues/46129)

## Test plan

Regular unit tests and the project from the issue: https://github.com/floklein/expo-server-sdk-packagejson-bug

### DIFF
--- a/packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
+++ b/packages/metro/src/ModuleGraph/worker/__tests__/collectDependencies-test.js
@@ -1209,6 +1209,125 @@ test('collects dependencies from module.createRequire factory pattern', () => {
   ]);
 });
 
+test('collects dependencies when createRequire result is bound to non-require name', () => {
+  const ast = astFromCode(`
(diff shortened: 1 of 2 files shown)


robhogan:
Hi @jeferson-sb 

This is Node JS specific which Metro doesn't directly support, but I see no issue recognising these dependencies for framework use *as long as the heuristic is sufficiently tight*.

So this should recognise *only* functions created by calling a factory where that factory is bound to the `createRequire` named export of a module called `node:module`.

Ideally, it recognises those even if they're not called `require`.

If we can tighten that up a bit (with tests against false positives) I'd be happy to accept this.

Thanks for working on it!

jeferson-sb:
thanks @robhogan Can you take another look?

PR #1757, opened 2026-06-30
Fix RAM bundle passing the deps function (not its result) to getTransformOptions

durvesh1992:
## Summary

`getRamBundleInfo` hands custom `getTransformOptions` a callback to look up a module's transitive dependencies. But it builds that callback as:

```js
async (x: string) => Array.from(getDependencies),
```

`getDependencies` is the *function* `(filePath) => getTransitiveDependencies(filePath, graph)`. This calls `Array.from` on the function itself and ignores `x` — it never invokes `getDependencies`. Since the closure has arity 1, `Array.from` treats it as array-like via `.length` and returns `[undefined]`. So any `getTransformOptions` that uses this callback to compute `preloadedModules`/`ramGroups` receives `[undefined]` for every file it asks about.

## Fix

```diff
-    async (x: string) => Array.from(getDependencies),
+    async (x: string) => Array.from(getDependencies(x)),
```

Changelog: [Fix] RAM bundle builds now pass real transitive dependencies to a custom `getTransformOptions`

## Test plan

```
jest packages/metro/src/DeltaBundler/Serializers/__tests__/getRamBundleInfo-test.js
```
Added a regression test that supplies a `getTransformOptions` which calls the callback for `/root/foo.js` and asserts it returns foo's transitive deps (`bar`, `baz`, `qux`). Fails before (`[undefined]`), passes after. Existing tests + snapshots unchanged (4 passed, 2 snapshots).


robhogan (approved):
Thanks!

PR #1866, opened 2026-08-19
ESM live bindings 3/n: Accept mode-1 helper shape in dependency analysis

vzaidman:
Summary:
Two related relaxations in the dependency-analysis pipeline that
together let it accept a mode-selecting second argument on Metro's
inlineable helper calls, and observe the underlying usage through the
namespace-shaped wrapper the mode-1 form returns.

## 1. `collectDependencies.getModuleNameFromCallArgs` arity

Previously rejected any inlineable-helper call with != 1 argument.
Now accepts 1 or 2 args. The dependency name is always the first arg;
the optional second arg is metadata for the runtime helper (a mode
selector - see the sibling change in
`metro-runtime/src/polyfills/require.js`). Behaviour unchanged for
existing 1-arg callers. 3+ args still rejected.

##    mode-1 helper results

The mode-1 form of `_$$_IMPORT_DEFAULT(id, 1)` returns a
namespace-shaped wrapper `{default: <value>}`. Under
`unstable_liveBindings`, downstream code accesses this via
`_$$_IMPORT_DEFAULT(id, 1).default`. The `.default` accessor is a
shape convention, not a semantic operation - analysers built on
`visitDependencyUses` (for example `extractSoundResources`, which
extracts sound-file names from `sx(name)` call sites) need to observe
the OUTER usage, not the intermediate accessor.

`walkReferences` now walks through `.default` accessors on mode-1
results, propagating mode-1 provenance through constant `var _foo =
helper(id, 1)` bindings. This means:

- `_$$_IMPORT_DEFAULT(id, 1).default(arg)` - the visitor sees the
  outer call and dispatches `visitCall`.
- `var _foo = _$$_IMPORT_DEFAULT(id, 1); _foo.default(arg)` - same,
  via the hoisted binding.
- `require('X').default` (arbitrary CJS namespace access, NOT mode-1)
  - unchanged, `.default` is preserved as the observable operation.

The mode-1 detection is a static structural check: the helper call
carries a numeric literal `1` as its second argument.

## Why one diff, before the plugin change

These two fixes are prerequisites for D115038400 (the plugin change
that starts emitting the mode-1 shape under `unstable_liveBindings`).
Together they let the dep-analysis pipeline handle the shape end to
end - relaxing this once, in isolation from the plugin change, means
each subsequent diff stands alone and builds cleanly.

Reviewed By: huntie

Differential Revision: D115860462

### DIFF
--- a/packages/metro/src/ModuleGraph/worker/visitDependencyUses.js
+++ b/packages/metro/src/ModuleGraph/worker/visitDependencyUses.js
@@ -95,6 +95,9 @@ export default function visitDependencyUses(
     importAllParamBinding,
     depMapParamBinding,
   } = bindModuleIRElements(file);
+  // `importDefaultParamBinding` is a getter that redoes a scope lookup on each
+  // access, and this is read once per reference below - so resolve it once.
+  const modeArgHelperName = importDefaultParamBinding.identifier.name;
   for (const path of requireParamBinding.referencePaths.concat(
     importDefaultParamBinding.referencePaths,
   )) {
@@ -107,7 +110,10 @@ export default function visitDependencyUses(
     if (dependencyFilter != null && !dependencyFilter(dep)) {
       continue;
     }
-    for (const {path: referencePath} of walkReferences(req.exprPath)) {
+    for (const {path: referencePath} of walkReferences(
+      req.exprPath,
+      modeArgHelperName,
+    )) {
       if (
         referencePath.parentPath &&
         referencePath.parentPath.node.type === 'CallExpression' &&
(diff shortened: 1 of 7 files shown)


packages/metro/src/ModuleGraph/worker/visitDependencyUses.js
```diff
@@ -95,6 +95,9 @@ export default function visitDependencyUses(
     importAllParamBinding,
     depMapParamBinding,
   } = bindModuleIRElements(file);
+  // `importDefaultParamBinding` is a getter that redoes a scope lookup on each
+  // access, and this is read once per reference below - so resolve it once.
```
vzaidman, line 99: modeArgHelperName is derived from importDefaultParamBinding.identifier.name unconditionally at the top of the function. If bindModuleIRElements can return a nullish binding for this element (as it does for depMapParamBinding, typed ?Binding), this eager dereference would throw earlier than the previous code path did. Worth confirming the binding is always present.

@robhogan

PR #1886, opened 2026-08-28
Back out "Add createStaticCrawler"

robhogan:
Original commit changeset: 3d0f468f29da

This is an AI slop API I think I accidentally left published when I left Meta, and was landed alongside some other work.

There's no need for a helper like this to exist in core and even if there were, this isn't the right shape.

 - "Static file" isn't really a thing - the crawler API is meant to allow for virtual file systems, they're not necessarily static.
 - Needlessly takes its input in a different shape to the output and converts (essentially most of the function) - expensive on large file sets.
 - `Reflect.set` to avoid a Flow error.. .*what*.

This has never been released, so we can still remove it as non-breaking.

Changelog: Internal

Test plan:
CI


cortinico (approved):
Review automatically exported from Phabricator review in Meta.

PR #1892, opened 2026-08-31
Use indexed watch folder paths for external assets

### LINKED ISSUE #19: Assets not picked up with multi project roots
Bug.

If you have a project setup with multiple roots and there are overlapping relative directories, assets are only picked up from directory in the first root.

Here is a failing testcase:
```
      const server = new AssetServer({
        projectRoots: ['/root', '/root2'],
        assetExts: ['png'],
      });

      fs.__setMockFilesystem({
        'root': {
          imgs: {
            'a.png': 'a image',
          },
        },
        'root2': {
          imgs: {
            'b.png': 'b image',
          },
        },
      });

      return Promise.all([
        server.get('imgs/a.png').then(data =>
          expect(data).toBe('a image')
        ),
        server.get('imgs/b.png').then(data =>
          expect(data).toBe('b image')
        ),
      ]);
```

The test fails on not being able to find `imgs/b.png`, because internally `_findRoot` always returns the first matching path (`imgs`) and no `b.png` is present there.

Checkout the fork at https://github.com/AlbertBrand/metro-bundler

I added the above testcase to `AssetServer-test.js` that shows the issue. 

I would expect the asset server to look in each project root for the asset. Or, even more correct: the AssetServer `get` call should take in account the root of the asset. If you then would refer to an asset from a specific root, it can directly resolve the correct one.

### LINKED ISSUE #290: Asset paths (httpServerLocation) invalid when asset is in watchFolders

### LINKED ISSUE #1614: Static images fail to load in monorepo on iOS
Hi Metro team 👋

I’m opening this new issue because I’m still experiencing **the same bug as in #1599**, but the original reporter is not active anymore and comments are closed.

## Summary
In a monorepo (Yarn workspaces), iOS fails to load static images imported via `require(...)` when the image file is located outside the app directory.

The behavior is **identical** to #1599:
- The image silently fails to display on iOS

### LINKED ISSUE #1615: Image assets fail to render in Yarn workspace monorepo with React Native 0.81.* and NX
### Description
Image assets fail to render in Yarn workspace monorepo with React Native 0.81.* and NX. 
If move assets into application's directory then metro resolves them correctly. But if assets are located in outside application's directory in the monorepo then bundler can not resolve them.

**FAILS TO RESOLVE:**
```
repo-root/
├─ apps/
│  └─ someApp/                    
│     ├─ src/
│      │  └─ App.tsx
│
└─ packages/
   └─ assets/                    ← asset lives OUTSIDE the app
     ├─ images/
      │  └─ logo.svg 
```

**SUCCESSFULLY RESOLVES:**
```
repo-root/
│
├─ apps/
│  └─ someApp/    
                
│     ├─ src/
│         ├─ assets/
│          │ └─ logo.svg      ← asset lives INSIDE the app
│      │  └─ App.tsx
│
└─ packages/
```
             
### Environment

- React Native: 0.81.5
- React: 19.1.0
- Yarn: 4.x (workspaces) and 1.22.22
- Node: 22.x
- OS: macOS 15 (M1 Pro)
- Bundler: Metro (default)

### Relevant dependencies

```
{
  "@react-native/metro-config": "0.81.5",
  "@react-native/babel-preset": "0.81.5",
  "@react-native/typescript-config": "0.81.5",
  "react-native-svg-transformer": "1.5.1",
  "react-native": "0.81.5",
  "react": "19.1.0",
  "nx": "21.6.3",
  "@nx/react-native": "21.6.3"
}
```

### App usage

```
// packages/mobile/App.tsx
import React from 'react'
import { Image } from 'react-native'

export const App = () => (
  <Image source={require('@someWorkspace/assets/logo.png')} style={{ width: 64, height: 64 }} />
)
```

When run

`nx start <project>`

### Metro config:

```
const {withNxMetro} = require('@nx/react-native');
const {getDefaultConfig, mergeConfig} = require('@react-native/metro-config');
const {
  getMetroAndroidAssetsResolutionFix,
} = require('react-native-monorepo-tools');
const {withDatadogMetroConfig} = require('@datadog/mobile-react-native/metro');

const assetsFix = getMetroAndroidAssetsResolutionFix();

const defaultConfig = getDefaultConfig(__dirname);
const {assetExts, sourceExts} = defaultConfig.resolver;

/**
 * Metro configuration
 * https://reactnative.dev/docs/metro

 *
 * @type {import('metro-config').MetroConfig}
 */
const customConfig = {
  transformer: {
    babelTransformerPath: require.resolve('react-native-svg-transformer'),
    publicPath: assetsFix.publicPath,
  },
  resolver: {
    assetExts: assetExts.filter(ext => ext !== 'svg'),
    sourceExts: [...sourceExts, 'cjs', 'mjs', 'svg'],
  },
  server: {
    enhanceMiddleware: assetsFix.applyMiddleware,
  },
};

module.exports = withNxMetro(
  withDatadogMetroConfig(mergeConfig(defaultConfig, customConfig)),
);
```

### NX project config:

```
{
  "name": "projectName",
  "$schema": "../../node_modules/nx/schemas/project-schema.json",
  "sourceRoot": "apps/someApp/src",
  "projectType": "application",
  "tags": ["app"],
  "targets": {
    "start": {
      "executor": "@nx/react-native:start",
      "dependsOn": ["sync-deps"],
      "options": {
        "port": 8081,
        "resetCache": true
      }
    }
}
...Other config
}
```

The app builds and runs, but the image renders as an empty space.

### Expected behavior
Metro/React Native should correctly resolve and bundle static image assets exported from shared workspace packages — the same way it handles JS modules.

### Actual behavior
When importing images from another workspace (e.g., @someWorkspace/assets), the app runs without any build or runtime error, but the image simply does not appear.

### Screenshots / recordings

Image

### Related issues:

- https://github.com/facebook/metro/issues/1599
- https://github.com/facebook/metro/issues/1614

### Additional info:

Possible fixe from above issues didn't work out.

robhogan:
## Summary

Assets outside `projectRoot` currently derive their `httpServerLocation` from a path relative to `projectRoot`. This can produce `..` paths which escape `/assets`, and does not identify which configured watch folder owns the asset.

This uses Metro's existing `[metro-watchFolders]/N/` namespace for external asset URLs - the same namespace `_sourceRequestRoutingMap` already serves source requests from, and which `_resolveWatchFolderPrefix` already decodes:

- asset transforms and `getAssets` encode the path relative to the containing watch folder
- the encoded path is in the transform cache key, since it reaches the transformer and is baked into the module output
- the `/assets` endpoint resolves the watch folder prefix before selecting platform and density variants
- assets within `projectRoot` retain their existing URLs

Keeping these requests under `/assets` preserves Metro's platform-specific and density-aware asset resolution.

The URL path reaches the default transformer as a separate positional argument rather than as a field on `JsTransformOptions`, because those options are spread wholesale into `BabelTransformerArgs` and this isn't something custom Babel transformers should be given.

Fixes: https://github.com/react/metro/issues/19
Fixes: https://github.com/react/metro/issues/290
Fixes: https://github.com/react/metro/issues/1614
Fixes: https://github.com/react/metro/issues/1615

Changelog:
```
 - **[Fix]**: Use indexed watch folder paths for assets outside `projectRoot`
```

### Expo

This doesn't change anything for Expo, but it should let ~~them~~ us delete a workaround.

Expo already avoids the `..` problem, by never letting the path into the URL path in the first place: `@expo/cli` sets `transformer.publicPath` to `/assets/?unstable_path=.` in development, so an external asset is addressed as `/assets/?unstable_path=./../shared/media` and the relative segment survives as an opaque, URL-encoded query parameter. For exports they use `?export_path=` plus the same `../` to `_` rewrite React Native applies. Both are handled in their own forks of `asset-transformer` and `getAssets`, which is also why this PR won't reach Expo projects on a Metro bump alone - their transformer calls `metro-transform-worker`'s `transform` with five arguments, so it never receives `assetUrlPath`.

`[metro-watchFolders]/N/` is the thing `?unstable_path=` is standing in for: an unambiguous name for a file outside `projectRoot` that survives URL normalisation, in the namespace the source-request routing map already uses. If Expo adopts it, the query parameter, the encode/decode step, and the `../` to `_` rewrite in `getAssets` can go, and external assets go back through the ordinary `/assets` path with platform and density selection intact.

## Test plan

Created a minimal project which imports an asset from a sibling watch folder:

```text
metro-watch-folder-e2e/
├── project/
│   ├── AssetRegistry.js
│   └── index.js
└── shared/
    └── media/
        ├── tone.mp3
        ├── tone@2x.mp3
        ├── tone@1x.ios.mp3
        └── tone@2x.ios.mp3
```

`index.js` imports the sibling asset:

```js
const asset = require('../shared/media/tone.mp3');
```

Started Metro from this checkout with `shared` configured as a watch folder:

```console
$ yarn start serve \
    --config /Users/robhogan/cowork/metro-watch-folder-e2e/metro.config.js \
    --host 127.0.0.1 \
    --port 8099 \
    --reset-cache

Metro ready
```

Requested an iOS bundle and inspected the generated asset metadata:

```console
$ curl --fail --silent --show-error \
    --output /private/tmp/metro-watch-folder-bundle.js \
    'http://127.0.0.1:8099/index.bundle?platform=ios&dev=true&minify=false'

$ rg -A 5 '"httpServerLocation"' /private/tmp/metro-watch-folder-bundle.js
    "httpServerLocation": "/assets/[metro-watchFolders]/1/media",
    "scales": [1, 2],
    "hash": "6db41d4f75e3e7e47d792c57abc17cb9",
    "name": "tone",
    "type": "mp3"
  });
```

Metro normalises `projectRoot` as watch folder 0, so the configured sibling directory is watch folder 1.

Requested the emitted asset path at 2x density for iOS:

```console
$ curl --silent --show-error --globoff --include \
    'http://127.0.0.1:8099/assets/[metro-watchFolders]/1/media/tone@2x.mp3?platform=ios&hash=6db41d4f75e3e7e47d792c57abc17cb9'

HTTP/1.1 200 OK
Content-Type: audio/mpeg
Content-Length: 7

ios-2x
```

The response body matches `tone@2x.ios.mp3`, confirming that the indexed URL resolves through the sibling watch folder while retaining platform and density selection.

New unit tests cover the two halves of that URL scheme meeting in the middle: `getAssetUrlPath` composed with `_resolveWatchFolderPrefix` round-trips back to the original absolute path for in-project, nested and watch-folder assets. It fails if the mapping is perturbed - an off-by-one in the watch folder index breaks it.

### Release builds

`httpServerLocation` is also read by React Native when it copies assets for a release build, so this changes where an external asset lands on disk and what it is called. Checked with a fixture holding one in-project asset and one in a sibling watch folder, running `Metro.runBuild` with `assets: true` from this checkout and passing the resulting `AssetData` through React Native 0.87.1's own `getAssetDestPathIOS` / `getAssetDestPathAndroid` / `filterPlatformAssetScales` - the three functions `saveAssets` uses to implement `--assets-dest`:

```text
before (main)
  ios      /assets/../shared/media  ->  assets/_shared/media/logo.png
  ios      /assets                  ->  assets/local.png
  android  /assets/../shared/media  ->  drawable-mdpi/_shared_media_logo.png
  android  /assets                  ->  drawable-mdpi/local.png

after
  ios      /assets/[metro-watchFolders]/1/media  ->  assets/[metro-watchFolders]/1/media/logo.png
  ios      /assets                               ->  assets/local.png
  android  /assets/[metro-watchFolders]/1/media  ->  drawable-mdpi/metrowatchfolders_1_media_logo.png
  android  /assets                               ->  drawable-mdpi/local.png
```

1x rows only; the `@2x` and `drawable-xhdpi` rows differ from these by suffix and folder alone.

In-project assets are untouched on both platforms. For external assets:

- The destination path and the Android resource identifier both change. `AssetSourceResolver` derives the runtime lookup from the same `httpServerLocation` that the CLI wrote the file from, so both halves move together and neither platform needs a change.
- The Android resource identifier now encodes the watch folder index, so reordering `watchFolders` renames the generated resource. Previously it encoded the relative path, so moving the directory renamed it instead. It is derived either way, and was never stable across a change in project layout.
- `..` no longer reaches either consumer. `scaledAssetURLNearBundle` and `getAssetDestPathIOS` each rewrite `../` to `_` independently to keep external assets inside the assets directory, which collapses distinct directories onto the same name; `assetServerURL` does not rewrite it at all, which is the escaping this PR fixes.

Also ran:

```console
$ yarn jest packages/metro/src packages/metro-transform-worker/src
$ yarn flow check
$ yarn typecheck-ts
$ yarn verify-api-snapshots
$ yarn lint
```

### DIFF
--- a/packages/metro-transform-worker/API.md
+++ b/packages/metro-transform-worker/API.md
@@ -74,9 +74,18 @@ export type MinifierOptions = {
 
 export type MinifierResult = {code: string; map?: BasicSourceMap | undefined};
 
-export const transform: (config: JsTransformerConfig, projectRoot: string, projectRelativePath: string, data: Buffer, options: JsTransformOptions) => Promise<TransformResponse>;
+export const transform: (
+config: JsTransformerConfig,
+projectRoot: string,
+projectRelativePath: string,
+data: Buffer,
+options: JsTransformOptions,
+extras?: TransformExtras,
+) => Promise<TransformResponse>;
 
 export type transform = typeof transform;
+
+export type TransformExtras = Readonly<{assetUrlPath?: string | undefined}>;
 
 export type Type = 'script' | 'module' | 'asset';
 
--- a/packages/metro-transform-worker/src/index.js
+++ b/packages/metro-transform-worker/src/index.js
@@ -133,6 +133,15 @@ export type JsTransformOptions = Readonly<{
   unstable_transformProfile: TransformProfile,
 }>;
 
+/**
+ * Inputs computed by Metro and passed alongside the public transform options.
+ * Kept out of JsTransformOptions because those are spread into
+ * BabelTransformerArgs, and these are for the default transformer only.
+ */
+export type TransformExtras = Readonly<{
+  assetUrlPath?: string,
+}>;
+
 opaque type AbsolutePath = string;
 opaque type ProjectRelativePath = string;
 
@@ -164,6 +173,7 @@ type JSONFile = {
 
 type TransformationContext = Readonly<{
   config: JsTransformerConfig,
+  extras: TransformExtras,
   projectRoot: AbsolutePath,
   options: JsTransformOptions,
 }>;
@@ -538,6 +548,7 @@ async function transformAsset(
     getBabelTransformArgs(file, context),
     assetRegistryPath,
     assetPlugins,
+    context.extras.assetUrlPath,
   );
 
   const jsFile = {
@@ -678,8 +689,10 @@ export const transform = async (
   projectRelativePath: string,
   data: Buffer,
   options: JsTransformOptions,
+  extras?: TransformExtras = {},
 ): Promise<TransformResponse> => {
   const context: TransformationContext = {
+    extras,
     config,
     options,
     projectRoot,
--- a/packages/metro-transform-worker/src/__tests__/index-test.js
+++ b/packages/metro-transform-worker/src/__tests__/index-test.js
@@ -135,6 +135,27 @@ test('transforms a simple module', async () => {
   expect(result.dependencies).toEqual([]);
 });
 
+test('uses the indexed watch folder path for asset URLs', async () => {
+  fs.mkdirSync('/root/external', {recursive: true});
+  fs.writeFileSync('/root/external/test.mp4', 'asset data');
+
+  const result = await Transformer.transform(
+    baseConfig,
+    '/root',
+    'external/test.mp4',
+    Buffer.from('asset data'),
+    {
+      ...baseTransformOptions,
+      type: 'asset',
+    },
+    {assetUrlPath: '[metro-watchFolders]/1/test.mp4'},
+  );
+
+  expect(result.output[0].data.code).toContain(
+    '"httpServerLocation": "/assets/[metro-watchFolders]/1"',
+  );
+});
+
 test('transforms a module with dependencies', async () => {
   const contents = [
     '"use strict";',
--- a/packages/metro-transform-worker/src/utils/assetTransformer.js
+++ b/packages/metro-transform-worker/src/utils/assetTransformer.js
@@ -20,6 +20,7 @@ export async function transform(
   {filename, options, src}: BabelTransformerArgs,
   assetRegistryPath: string,
   assetDataPlugins: ReadonlyArray<string>,
+  assetUrlPath?: string,
 ): Promise<{ast: File, ...}> {
   options = options || {
     platform: '',
@@ -32,7 +33,7 @@ export async function transform(
 
   const data = await getAssetData(
     absolutePath,
-    filename,
+    assetUrlPath ?? filename,
     assetDataPlugins,
     options.platform,
     options.publicPath,
--- a/packages/metro/src/Assets.js
+++ b/packages/metro/src/Assets.js
@@ -243,6 +243,50 @@ export async function getAssetData(
   return await applyAssetDataPlugins(assetDataPlugins, assetData);
 }
 
+/**
+ * Returns the path used to identify an asset in its development server URL.
+ * Assets outside projectRoot use an indexed watch folder prefix so that the
+ * URL unambiguously identifies their configured root.
+ *
+ * Roots are resolved as Server does when it decodes these paths, so that the
+ * index here and the index it reads back refer to the same directory.
+ */
+export function getAssetUrlPath(
+  assetPath: string,
+  projectRoot: string,
+  watchFolders: ReadonlyArray<string>,
+): string {
+  const projectRelativePath = path.relative(
+    path.resolve(projectRoot),
+    assetPath,
+  );
+  if (isPathInsideRoot(projectRelativePath)) {
+    return normalizePathSeparatorsToPosix(projectRelativePath);
+  }
+
+  for (let i = 0; i < watchFolders.length; i++) {
+    const watchFolderRelativePath = path.relative(
+      path.resolve(watchFolders[i]),
+      assetPath,
+    );
+    if (isPathInsideRoot(watchFolderRelativePath)) {
+      return normalizePathSeparatorsToPosix(
+        path.join('[metro-watchFolders]', String(i), watchFolderRelativePath),
+      );
+    }
+  }
+
+  return normalizePathSeparatorsToPosix(projectRelativePath);
+}
+
+function isPathInsideRoot(relativePath: string): boolean {
+  return (
+    relativePath !== '..' &&
+    !relativePath.startsWith('..' + path.sep) &&
+    !path.isAbsolute(relativePath)
+  );
+}
+
 async function applyAssetDataPlugins(
   assetDataPlugins: ReadonlyArray<string>,
   assetData: AssetData,
--- a/packages/metro/src/DeltaBundler/Serializers/__tests__/getAssets-test.js
+++ b/packages/metro/src/DeltaBundler/Serializers/__tests__/getAssets-test.js
@@ -10,7 +10,7 @@
 
 jest.mock('../../../Assets');
 
-import {getAssetData} from '../../../Assets';
+import {getAssetData, getAssetUrlPath} from '../../../Assets';
 import getAssets from '../getAssets';
 
 beforeEach(() => {
@@ -18,6 +18,9 @@ beforeEach(() => {
     path,
     localPath,
   }));
+  getAssetUrlPath.mockImplementation(
+    jest.requireActual('../../../Assets').getAssetUrlPath,
+  );
 });
 
 test('should return the bundle assets', async () => {
@@ -82,16 +85,32 @@ test('should return the bundle assets', async () => {
         ],
       },
     ],
+    [
+      '/external/6.png',
+      {
+        path: '/external/6.png',
+        output: [
+          {
+            type: 'js/module/asset',
+            data: {code: '//', lineCount: 1, map: [], functionMap: null},
+          },
+        ],
+      },
+    ],
   ]);
 
   expect(
     await getAssets(dependencies, {
       projectRoot: '/tmp',
-      watchFolders: ['/tmp'],
+      watchFolders: ['/tmp', '/external'],
       processModuleFilter: () => true,
     }),
   ).toEqual([
     {path: '/tmp/3.png', localPath: '3.png'},
     {path: '/tmp/5.mov', localPath: '5.mov'},
+    {
+      path: '/external/6.png',
+      localPath: '[metro-watchFolders]/1/6.png',
+    },
   ]);
 });
--- a/packages/metro/src/DeltaBundler/Serializers/getAssets.js
+++ b/packages/metro/src/DeltaBundler/Serializers/getAssets.js
@@ -12,7 +12,7 @@
 import type {AssetData} from '../../Assets';
 import type {Module, ReadOnlyDependencies} from '../types';
 
-import {getAssetData} from '../../Assets';
+import {getAssetData, getAssetUrlPath} from '../../Assets';
 import {getJsOutput, isJsModule} from './helpers/js';
 import path from 'node:path';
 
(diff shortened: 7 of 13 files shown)


vzaidman:
What if an asset is outside of a watched folder?

packages/metro-transform-worker/API.md
```diff
@@ -74,10 +74,19 @@ export type MinifierOptions = {
 
 export type MinifierResult = {code: string; map?: BasicSourceMap | undefined};
 
-export const transform: (config: JsTransformerConfig, projectRoot: string, projectRelativePath: string, data: Buffer, options: JsTransformOptions) => Promise<TransformResponse>;
+export const transform: (
```
vzaidman: formatting issue? @robhogan

robhogan:
> What if an asset is outside of a watched folder?

By the preexisting design, it's not allowed to be. Assets have to be resolved by the resolver just like JS modules, and the resolver can't "see" anything that isn't watched.

In fact, assets that were previously visible despite being in the file map `blockList` were deemed to be a security hole: https://github.com/react/metro/pull/1643. So it's a pretty hard rule in Metro's design that assets must be within the projectRoot + watchFolders.

A more extreme version of the security hole back in 2016 - the assets endpoint allowed reading any file on disk: https://github.com/facebook/react-native/issues/6398

packages/metro-transform-worker/API.md
robhogan: I don't think so - the params just got reformatted to multiline because `extras` was added

packages/metro-transform-worker/src/index.js
```diff
@@ -678,8 +689,10 @@ export const transform = async (
   projectRelativePath: string,
   data: Buffer,
   options: JsTransformOptions,
+  extras?: TransformExtras = {},
 ): Promise<TransformResponse> => {
   const context: TransformationContext = {
+    extras,
```
robhogan: @motiz88 / @vzaidman  - I'm not sure how I feel about `extras` when we already have `config` and `options` - they do serve different purposes but it looks messy. Open to ideas, or I'm happy to go back to this.

(This PR was more fresh-from-Claude than I'm proud of, but I do recall threading this option through `JSTransformOptions` and removing it where it didn't belong was in some ways worse.)

packages/metro-transform-worker/API.md
vzaidman: it is. the idents are broken
Image
Image

robhogan: Oh I see. Weird - those API.md files are auto-generated, but I'll have a look..

robhogan: Opened #1936 - it's pre-existing and independent but worth fixing

packages/metro-transform-worker/src/index.js
vzaidman: should we rename them to make clear what purposes they serve a bit better?

vzaidman: extras is too general. maybe `context` is better.

robhogan: `context` is already the name of the thing containing `extras`.

I've had a bit of a re-think of this. _Ideally_, we'd reduce the transformer's 6 arguments to 3:
 - `config` (fixed transformer config, as we already have)
 -  `options` (things varied on the request - platform, dev/prod, etc, as we already have), 
 - `file` (facts about this file, immutable per-file under the current config - things like relative path, this URL and body data)

These are legitimately three separate buckets, two of which already exist and are fine. `file` is spread across our other positional args.

I don't want to introduce a breaking change in this PR, but I think that helps me settle on the idea that until we group the `file` facts together into one bag, we can just add them as positional args. Realistically, I don't see us needing to add any more before the next major release.

So I've updated the PR to flatten the argument down to a string and remove the "extras" bag. Hope that makes sense - if so @motiz88, it'll need re-importing.

robhogan:
Incidentally, @motiz88, this adds to the case for your #1706 and #1707. It's compatible, as far as I can tell - I can pick that stack up after unless you're planning to get back to it?

robhogan:
@motiz88 / @vzaidman - anything I can do to help with this one? There are a few things I'd like to start sequencing behind it. Thanks 🙏

PR #1931, opened 2026-09-14
Separate formatting from lint in GitHub CI

cortinico:
Summary:
Give formatting its own GitHub CI step and preserve the same coverage in the aggregate `yarn test` command. This makes formatter failures independently visible before Metro becomes GitHub-first while leaving `yarn lint-fix` as the convenient combined fix command.

___

Differential Revision: D119662169


robhogan:
Landed as 4967a2f72374fdfd8fd17410d46d6c5bce3e352e

